PR #2, opened 2014-11-20
Override FileSystem.getScheme()

rmetzger:
My application is accessing the FileSystem.getScheme() method, which leads to the following exception:

``` java
throw new UnsupportedOperationException("Not implemented by the " + getClass().getSimpleName() + " FileSystem implementation");
```

I have done this change using the GitHub Web UI because I was too lazy to check out the code. I think my code is building but I haven't tested it ;)


medb:
This has been fixed already.

PR #6, opened 2015-06-11
Update google-apis, add .gitignore

CBrophy:
Updated to 1.20 google-api, rev35 and rev213 gcs and biq query dependencies to fix certificate errors when using the GCS connector from a GCE instance: www.googleapis.com/##.##.##.## does not match *.googleapis.com...etc

Added .gitignore to stop accidental garbage commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+target/
+bin/
+*.iml
+.idea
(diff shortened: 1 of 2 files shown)


dennishuo:
Thanks for the suggestion! This repo is not quite tooled for accepting direct pull requests at this point, but we went ahead and upgraded the storage/bigquery dependencies to the latest versions in https://github.com/GoogleCloudPlatform/bigdata-interop/commit/8c4177b461482fc975bf49f9b957c0272064ecab and added a .gitignore aligned with some other related Google projects here: https://github.com/GoogleCloudPlatform/bigdata-interop/commit/178a54f8c3aa6db8e798f237728cb03f0add6e6e

PR #18, opened 2015-10-27
Set SNAPSHOT versions

AngusDavis:


dennishuo:
Obsolete as of new version releases.

PR #24, opened 2016-01-18
Fix infinite wait if there is no 0-record file

### LINKED ISSUE #23: DynamicFileListRecordReader error
I'm experiencing issue with proper reader termination.
DynamicFileListRecordReader expects 0-record file to be present at the bucket path, but looks like it's not automatically created, so the reader goes into infinite loop.
 According to this
https://cloud.google.com/bigquery/exporting-data-from-bigquery 
this 0-record file is created only when using wildcard urls. But according to hadoop log single-url with no wildcards is used and that wait is unnecessary

My config for haddop job is

``` java
        BigQueryConfiguration.configureBigQueryInput(conf, "publicdata:samples.shakespeare");
        conf.set("fs.gs.impl", "com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem");
        // tempoary path where to download data from BiqQuery tables
        conf.set(BigQueryConfiguration.TEMP_GCS_PATH_KEY, "gs://mybucket/mypath");
        conf.set(BigQueryConfiguration.PROJECT_ID_KEY, "myprojid");
        conf.set(GoogleHadoopFileSystemBase.GCS_PROJECT_ID_KEY, "myprojid");
```

So maybe remove that wait or make DynamicFileListRecordReader more flexible?

xardazz:
Fix infinite wait if there is no 0-record file and all records are already read.
This fixes  #23

### DIFF
--- a/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/DynamicFileListRecordReader.java
+++ b/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/DynamicFileListRecordReader.java
@@ -282,7 +282,7 @@ public class DynamicFileListRecordReader<K, V>
    *     files we plan to read.
    */
   private boolean shouldExpectMoreFiles() {
-    if (endFileNumber == -1 || knownFileSet.size() <= endFileNumber) {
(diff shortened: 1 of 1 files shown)


xardazz:
I signed it!

PR #25, opened 2016-01-18
Fix infinite wait if there is no 0-record file

### LINKED ISSUE #23: DynamicFileListRecordReader error

xardazz:
Fix infinite wait if there is no 0-record file and all records are already read.
Fixes #23

### DIFF
--- a/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/DynamicFileListRecordReader.java
+++ b/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/DynamicFileListRecordReader.java
@@ -282,7 +282,7 @@ public class DynamicFileListRecordReader<K, V>
    *     files we plan to read.
    */
   private boolean shouldExpectMoreFiles() {
-    if (endFileNumber == -1 || knownFileSet.size() <= endFileNumber) {
+    if ((endFileNumber == -1 && recordsRead < estimatedNumRecords) || knownFileSet.size() <= endFileNumber) {
       return true;
     }
     return false;


AngusDavis:
HI @xardazz, thanks for the PR.
How often are you experiencing the 0-record file not showing up? According to the BigQuerydocs (https://cloud.google.com/bigquery/exporting-data-from-bigquery) on multiple wildcard export paths, the 0-record file should always be written and AFAIK the 0-record file is the only way we can be certain that we've read all data.

xardazz:
Hi, @AngusDavis ! Actually I'm experiencing this issue with every table from public data independent from number of files. Yeah, I know that according to link it should always be written, but I've never seen it. (record reader just  looped infinitely waiting for that 0 -record, though all records were read (recordsRead = estimatedRecordNumber))

xardazz:
@AngusDavis Some more info
This issue is somehow related to this
com.google.cloud.hadoop.io.bigquery.ShardedExportToCloudStorage.java
```java

  // If we try 'sharded export' with only one shard, then BigQuery doesn't recognize it as a
  // sharded export and thus doesn't generate an end-marker file. Even if BigQuery will only
  // export one file, as long as we specify two shards, BigQuery will create both end-marker files
  // while only populating actual data into one of the two shards.
  public static final int MIN_SHARDS_FOR_SHARDED_EXPORT = 2;

```

computeNumShards can actually return value less than MIN_SHARDS_FOR_SHARDED_EXPORT (if 1 map task is specified for example). But even if return MIN_SHARDS_FOR_SHARDED_EXPORT, it still fails, because 0-record is created only in first shard path, but second shard remains empty.

pmkc:
Closing since  #23 has been resolved.

PR #26, opened 2016-02-05
Fix infinite loop

xardazz:
There is a problem with infinite loop in record reader in two cases:
1. BigQuery returned 0 records, but didn't create 0-record file
2. First task attempt failed. In that case on the second task attempt record reader will wait indefinitely, but no files will appear.


medb:
@xardazz may you grant edit access to this PR for repo owners, so I can rebase and merge it?

medb:
Superseded by https://github.com/GoogleCloudPlatform/bigdata-interop/pull/233

PR #30, opened 2016-03-18
Return object checksums from GCS metadata

rohansingh:


steffengr:
@dennishuo Could you have a look at this?

AngusDavis:
@rohansingh, @steffengr - I'm a bit concerned about using metadata for this (and I think I made a few mistakes with types in the metadata interface that I'd like to be able to clean up). Would adding a class 

    class ValidationAttributes {
       Optional<String> md5Hash;
       Optional<String> crc32c;
       // ctor and getters omitted for example purposes.
    }

and then exposing that via a new GoogleCloudStorageItemInfo#getValidationAttributes() make sense? 

Also, how are these intended to be used? The comment 8f748a0 leads me believe that this is the envisioned process:

1. Create a new MR that computes GCS compatible checksums for all files and preserve them somewhere
2. Perform distcp with checksums turned off
3. Check GCS checksums and compute a list of 'bad objects' to recopy
4. Re-run distcp on bad objects + any new source files 

Is that correct?

rohansingh:
@AngusDavis Regarding adding a `ValidationAttributes` class and exposing it, yeah, that seems like a great idea. I did feel a bit dirty just stuffing it into the metadata.

You are correct about the intended use case. Does that sound like a reasonable approach?

AngusDavis:
@rohansingh - Sounds reasonable to me. Right now, this project isn't set up to handle inbound PRs very well. I will get a PR out (or internal change + export to github) to wire in a ValidationAttributes to GoogleCloudStorageItemInfo. 
I have a tracking bug for a new connectors release that keeps slipping, but adding this in should get it done a bit more quickly.

rohansingh:
@AngusDavis Great. So just to confirm, this PR is obsolete and this will be taken care of by your PR/export?

AngusDavis:
@rohansingh - correct. Note that instead of using Optional<String>, I'll be using byte[] for each attribute and that byte array will be the result of decoding the base64 string from GCS.

rohansingh:
Awesome. In that case I'll just close this.

AngusDavis:
@rohansingh - This became VerificationAttributes in commit 867e0d751e53b24fe3a50d574cf282e8f92cb198 / connector release 1.4.5.

PR #35, opened 2016-08-11
Support specifying partition in table name

xardazz:
Currently connector doesn't allow table names containing $ sign (that specifies partition for time-partitioned tables). So this PR fixes it.


tswast:
Mind adding a test to [BigQueryOutputFormatTest](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/bigquery/src/test/java/com/google/cloud/hadoop/io/bigquery/BigQueryOutputFormatTest.java) with an example input name and output name?

xardazz:
@tswast Ok, added

PR #40, opened 2016-09-29
Perform lazy seek before in GoogleCloudStorageReadChannel.position

rajeshbalamohan:
Main issue: When reading ORC data (hive columnar format which extensively uses positional reads), it ended up throwing the following exception with gcs-connector. 

`Caused by: java.io.IOException: Somehow read -1 bytes trying to skip 617 more bytes to seek to position 2634, size: 6950
        at com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadChannel.position(GoogleCloudStorageReadChannel.java:535)
        at com.google.cloud.hadoop.fs.gcs.GoogleHadoopFSInputStream.seek(GoogleHadoopFSInputStream.java:301)
        at org.apache.hadoop.fs.FSInputStream.read(FSInputStream.java:74)
        at com.google.cloud.hadoop.fs.gcs.GoogleHadoopFSInputStream.read(GoogleHadoopFSInputStream.java:266)
        at org.apache.hadoop.fs.FSInputStream.readFully(FSInputStream.java:121)
        at org.apache.hadoop.fs.FSDataInputStream.readFully(FSDataInputStream.java:111)`

Simple demo is available in https://github.com/rajeshbalamohan/gcs-connector-issue-demo.

It would be good to use FS contract tests of Hadoop (https://issues.apache.org/jira/browse/HADOOP-9361, https://github.com/apache/hadoop/tree/trunk/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract)

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -527,6 +527,8 @@ public class GoogleCloudStorageReadChannel
       if (skipBuffer == null) {
         skipBuffer = new byte[SKIP_BUFFER_SIZE];
       }
+      // perform lazy seek before reading
+      performLazySeek();
       while (seekDistance > 0) {
         int bytesRead = readChannel.read(
             ByteBuffer.wrap(skipBuffer, 0, (int) Math.min(skipBuffer.length, seekDistance)));


AngusDavis:
Thanks for the PR! I'm currently trying to find a nice way to accomplish this without necessarily requiring a reopen (assuming the seek will drop some / most of the buffer, it doesn't make a lot of sense to initiate a read operation just to throw it away).

We have a giant TODO to add FS contract tests.

steveloughran:
Adding the Hadoop FS contract tests would have found this; everyone gets seek() wrong at least once. Regarding optimising seek performance, you should look at [S3a phase II](https://issues.apache.org/jira/browse/HADOOP-11694) and it's fadvise=random option, alongside other tunings of expensive metadata operations. But before you play with all that, you need to

1. Pull in the hadoop contract tests and start running them
1. Fix the failures
1. Instrument the code with metric collection on close vs abort, forward seek vs backward, etc, etc, so you can assess the cost of changes in downstream code.
1. For s3a, this instrumentation [is wired up for regression testing](https://steveloughran.blogspot.co.uk/2016/04/distributed-testing-making-use-of.html)
1. Tune that performance; add more metrics based testing to isolate the performance testing from any wall clock timings.

I'm happy to talk about how to do all of this; you should be adding those contract tests before any other lines of code.

dennishuo:
Thanks everyone for the input! Closed by https://github.com/GoogleCloudPlatform/bigdata-interop/commit/e69042883e197b7d741950a2659986f5762cbff0

evilsoapbox:
Hi Steve,

Thanks for the feedback - we appreciate it. At this point we have a somewhat large to-do list but everything you mention is on our radar. Last we checked (admittedly it was awhile ago) some connectors were also not fully passing all tests. We will investigate more on this front.

That said, we're committed to making things better for the GCS connector. We'll also happily take you up on your offer - should we ping you via GitHub or some other means?

Thanks!

- Dataproc/data OSS team

steveloughran:
Well, I do appreciate the honest answer, even if it is not the one I'd prefer to have seen. In pariticular It doesn't make me confident that your module is 100% ready for use, even if it first shipped in 2014. 

Those contract tests went in precisely because the openstack swift:// connector hit problems after it went out the door in 2013;, problems which surfaced because of quirks we hadn't expected. Indeed, if you look at [HADOOP-9361](https://issues.apache.org/jira/browse/HADOOP-9361) you can see that pretty much every FS connector had a surprise or two, and we are still getting round to the fun ones —globStatus()— But we do at least try, especially before doing major bits of work on the ASF maintained object stores.

bq.  Last we checked (admittedly it was awhile ago) some connectors were also not fully passing all tests. We will investigate more on this front.

That's not quite the case; more that alongside the we not having defined the semantics of things like [delete("/")](https://issues.apache.org/jira/browse/HADOOP-12977) in a way that makes sense to users, we often hit the problem that the semantics of an object store are subtly different from HDFS. To address that, we let each FS provide info on how they behave, hence tests can run with a different set of expectations for [file://](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/test/resources/contract/localfs.xml) versus [s3a://](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/test/resources/contract/s3a.xml). 

I think you should get on the hadoop common-dev list and say "how do we test our connector?". Put in an afternoon or two to wire up the tests, hookup jenkins/your CI tooling, and then you have a new set of compliance and regression tests to play with. Then you can start to worry about Hive and Spark testing. And do that before you try to add any more functionality to this module, features or performance optimisation —because there's no point optimising something that doesn't work.

I'm reachable via github and as stevel @ apache.org .

dennishuo:
Oh, we're as excited as anyone else to see the movement towards a trustworthy single consolidated source of truth :) Great to see continued progress standardizing on the new contract tests.

This project predated the contract tests, (looks like initial github commit *just* missed the [main addition of contract tests](https://github.com/apache/hadoop/commit/e09ea0c06ee1caa5a9ebae0a8f0273dfe04d05e5) and then the bulk of the initial work on the connector occurred within google-internal source repositories before getting the paperwork together to set up this github repo) and so we were never happy with having to rely on our own setup of behavioral comparison tests between HDFS/GCS/LocalFS, etc. For posterity, this had a similar layout as the contract tests in the form of an [abstract base class](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java) and [pluggable method-specific "behavior" interfaces](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HdfsBehavior.java):

```
public class HdfsBehavior
    implements DeletionBehavior, MkdirsBehavior, RenameBehavior {

  // -----------------------------------------------------------------
  // Deletion behavior.
  // -----------------------------------------------------------------

  /**
   * Returns the MethodOutcome of trying non-recursive deletes for non-empty directories.
   */
  @Override
  public MethodOutcome nonEmptyDeleteOutcome() {
    return new MethodOutcome(MethodOutcome.Type.THROWS_EXCEPTION, IOException.class);
  }

  /**
   * Returns the MethodOutcome of trying to delete a non-existent filename.
   */
  @Override
  public MethodOutcome nonExistentDeleteOutcome() {
    return new MethodOutcome(MethodOutcome.Type.RETURNS_FALSE);
  }
  ...
```

So it's especially good to see how quirks like `seek(-1)`--which indeed tripped us up and [remains a comment](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java#L472) in our version behavior comparison tests--is now documented and reconciled in the [contract seek test](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/contract/AbstractContractSeekTest.java#L165). The sooner we can delete [all of our code in our own equivalent of contract tests](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java) knowing that the upstream contract tests fully cover what we need to test, the better. And maybe we're lucky and that day is today! We'll just have to take some time to verify proper functional coverage.

On a related note, how are you approaching contract-test documentation and coverage of other popular closed-source FileSystem implementations which may or may not have sibling implementations within Hadoop?

Also, do you happen to have any handy JIRAs or discussion groups/documents detailing the overall relationship between the real contract tests and the other [FileSystemContractBaseTest](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/test/java/org/apache/hadoop/fs/FileSystemContractBaseTest.java) and its related test base classes?

steveloughran:
> So it's especially good to see how quirks like seek(-1)--which indeed tripped us up and remains a comment in our version behavior comparison tests--is now documented and reconciled in the contract seek test. 

It's a question of when does seek() fail: in the call or in a follow-on {{read()}}. Posix lets you go past the EOF, because more data may have arrived. Nothing should let you do a seek(-1), that's the kind of thing its "obvious" we have to fix. Harder is when, say, HDFS doesn't do rename() the way Posix does. Do we tell the Posix team that they are wrong? Do we fix HDFS? Either way we are trapped.

> On a related note, how are you approaching contract-test documentation and coverage of other popular closed-source FileSystem implementations which may or may not have sibling implementations within Hadoop?

Real filesystems or things supporting an object store API? For filesystems, it's left to the supplier; redhat test GlusterFS, for example. Object Stores? Probably the same. One thing to consider is that if there is some ambiguity/quirk/bug that should have a new test, that's something that can go in, especially if it comes with a patch for the Hadoop filesystems which would fail the test. However, you will be expected to argue the case for patching HDFS, that is, who gets to define the correct behaviour of any operation. Generally, the HDFS team consider it their domain. 

{{FileSystemContractBaseTest}} predates the others; we leave it alone because we know external people have subclassed it, possibly adding their own tests. Which is also why it's stuck as a JUnit 3 test suite which ignores the ability to skip tests selectively. All new tests go into the {{org.apache.hadoop.fs.contract}} package, extending existing test cases if they are just exploring more corner cases everything should support (we do locally test on all the filesystems and object stores before adding such tests); adding whole new test suite classes for features. Where they are weak is lack of concurrency testing; those tests are very much "does the FS API do roughly what we expect in for a single caller"? It's foundational stuff —but it is the first step to testing up the stack

dennishuo:
Makes sense. One specific other object-store-based FileSystem that came to mind is EMR FS since it has the unique position of S3 having ~3 "different" implementations in the wild, and I was curious if anyone knew offhand if that matches S3A's contract config out of the box.

Also, have you considered (or possibly already finished) creating in-memory reference implementations of the lower-layers of things like S3 or Azure blob storage? For expediency we have a full swappable lower-layer fake [InMemoryGoogleCloudStorage](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java) which so far has done a surprisingly good job of surfacing nearly exactly the same contract-test errors as the integration test version (the only difference so far was it [throwing an exception](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java) when reading a being-written file vs just "not found" like the real GCS service, but now it's fixed; everything else showed the same "wrong IOException vs FileAlreadyExistsException", etc type of errors as the real service.

It'd never be a full replacement for integration tests of course, but having such a thing to enable always-on unittests which don't require service credentials and which run in seconds would really remove barriers of entry to more developer contributions.

dennishuo:
@steveloughran Quick question, while I'm in the process of adding the contract tests, re: other connectors not fully passing tests yet - for the [S3A explicit override of testRenameDirIntoExistingDir](https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/contract/s3a/ITestS3AContractRename.java#L60), is the plan going to be to bring it in line with the existing contract test, or is it more likely you'll add something like "supports-rename-into-non-empty-directory" as a configurable contract option? Or is the heavyweight override of not supporting rename into non-empty directories considered the long-term contract for S3A?

Out of the box [our rename is closer to HDFS](https://github.com/GoogleCloudPlatform/bigdata-interop/commit/a4cf5ce25d9435cf5f9fc5222fe1e4c37a0c2595) than LocalFileSystem (by design, since we want to facilitate HDFS migrations/hybridizations more so than impose POSIX semantics), and GCS supports rename-into-non-empty-directory the same way HDFS does. But for ease of use of course we might want to offer some configurable mode for doing everything POSIX-style instead; it'll be good to know whether the S3A behavior is something we should account for in terms of configurability.

steveloughran:
> EMR FS since it has the unique position of S3 having ~3 "different" implementations in the wild, and I was curious if anyone knew offhand if that matches S3A's contract config out of the box.

That's Amazon's problem. Do note that Azure can be used as a drop-in replacement for HDFS even for running HBase, which is the ultimate stress test. Look at its code to see what's done there, taking advantage of the fact that WASB supports leases for atomic/transacted operations. And it's rename is so, so fast. And Microsoft work with us at Hortonworks to test everything: they get to file JIRAs against my name.

>  have you considered (or possibly already finished) creating in-memory reference implementations of the lower-layers of things like S3 or Azure blob storage?

No,, because they would represent our interpretations of the API, rather than the implementation —and hide network effects as well as all performance issues we are trying to find. I if you look at the [S3a phase II)(https://issues.apache.org/jira/browse/HADOOP-11694) and [Phase III](https://issues.apache.org/jira/browse/HADOOP-13204) work, you can see some of the fun things that have surfaced. 

1. Auth problems between S3 Frankfurt (v4 API) and S3 US-east/west; we're still improving error messages there.
1. tests failing in the presence of an HTTP proxy
1. Visible consistency problems
1. Performance tuning: when to use forward seek vs HTTP abort/re-open. Wouldn't be visible on a local test system
1. Performance: whether to GET in blocks or use GET pos-EOF for reads
1. performance: is there a visible speedup from full object listing vs recursive through mock directories

[HADOOP-13560](https://issues.apache.org/jira/browse/HADOOP-13560) alone has to address
1. excess memory consumption if bandwidth to S3 is significantly less than data generation rate. (this predated HADOOP-13560; it's tests just quantified it). fix: multiple buffer options, including blocks to disk.
1. network failures on a PUT failing to be replayed because the on-file data was wrapped in a BufferedOutputStream, which couldn't go back to the older data. Fix: don't buffer.
1. transient network failures on the final PUT of the multipart metadata not being retried, so losing the entire upload. Fix: retry logic with backoff.

All the network failure problems aren't being looked for, they just surface the more you test things like multiGB uploads over long-haul links, test against different endpoints, test in different environments (e,g behind firewalls). With a local mock implementation we'd miss all those problems, and ship something we couldn't say "was production ready"

You have an advantage of the original codebase; it'd be like MiniHDFS and MiniYARN cluster, a good first step. But it's not a substitute for the real thing. Which, with object stores and tests designed not to persist data, is something anyone with a login and a small amount of money can run. The barrier to running object store tests is low. Which is good, as ASF Jenkins/Yetus can't run those tests on patch submission (security reasons). This is why there is a strict policy of people [explicitly declaring which endpoint they tested a patch against](https://wiki.apache.org/hadoop/HowToContribute#Submitting_patches_against_object_stores_such_as_Amazon_S3.2C_OpenStack_Swift_and_Microsoft_Azure): 

What we do like, however, is

* Public read-only test datasets.  Ideally: 20+ MB CSV, ORC, Parquet files, complex directory structures, the entire TCP-DS datasets. This lets us have read-only tests free at the point of use for anyone playing with Hadoop source. Really we need code to generate all this data so it can be consistent across public/private infras. For now, we have tests which hit AWS landsat's index files.
* Access to any endpoints capable of simulating failure conditions.
* Cloud storage providers to test release versions of Hadoop against the nightly builds of their infrastructure. Nothing like a new version breaking authentication.
* Cloud storage providers to test pre-release versions of Hadoop against their nightly & release builds. Stops us breaking things.
* Enough details about implementation to understand how to tune code. When to skip vs close/reopen, what makes a good block size, impact of VM type on bandwidth, rename bandwidth, failure conditions. Oh, and : what hashing algorithm is optimal for distributing blobs so as to avoid throttling. Very important that.
* Under NDA if necessary: plans for future features.

Those of us working full time on this stuff love donated storage and VM time, so we can do a lot of our testing from VMs in the infrastructure for the faster test runs; and for the full scale/perf tests of things like Hive & Spark. Those run up bills fast.

steveloughran:
> other connectors not fully passing tests yet - for the S3A explicit override of testRenameDirIntoExistingDir, is the plan going to be to bring it in line with the existing contract test,

[HADOOP-13402](https://issues.apache.org/jira/browse/HADOOP-13402). *S3A should allow renaming to a pre-existing destination directory to move the source path under that directory, similar to HDFS.*. We know of it, nobody has fixed it yet. Patch welcome. 

Rename is a sore point, ask anyone who does a real FS about what constitutes an atomic rename transaction across multiple parent dirs. We made a mistake there in not restricting it to the same parent directory. Unfortunately, it's the standard mechanism for implementing atomic commits with either speculative execution or a risk of network partition events.

[HADOOP-9565](https://issues.apache.org/jira/browse/HADOOP-9565) is a long standing JIRA saying "we really need an object store API", though it also needs to work with filesystems. I do want to get on it it with support purely for: PUT, DELETE, COPY, LIST prefix & suffix; stuff blobstores can do nicely, filesystems can mimic if not present. A key one is PUT; if we can do it atomically, then you get a commit that doesn't need rename.

I plan to specify that little project in TLA+, making the commit algorithm provable; the work in s3guard is clearly related.

steveloughran:
Ps, now that you are implement the tests (thank you!), you may want to look at [HADOOP-13502](https://issues.apache.org/jira/browse/HADOOP-13502); we're tuning some of the object store tests

PR #54, opened 2017-06-08
Retry upload request

ravwojdyla:


ravwojdyla:
Signed via Spotify.

dennishuo:
Thanks for the pull request; in general there are two types of retries: low-level HTTP-interceptor retry handlers and then end-to-end retry handlers involving `ResilientOperation`. The low-level stuff should already be plugged in via `ChainingHttpRequestInitializer`: https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/util/src/main/java/com/google/cloud/hadoop/util/ChainingHttpRequestInitializer.java#L73

There should be an [HttpBackOffUnsuccessfulResponseHandler](http://grepcode.com/file/repo1.maven.org/maven2/com.google.http-client/google-http-client/1.19.0/com/google/api/client/http/HttpBackOffUnsuccessfulResponseHandler.java?av=f#139) plugged in under the hood which causes low-level retries when retries are allowed inside [HttpRequest](http://grepcode.com/file/repo1.maven.org/maven2/com.google.http-client/google-http-client/1.19.0/com/google/api/client/http/HttpRequest.java?av=f#1006). Indeed, a very common case is for an oauth access token to expire a default 5-limit TTL in the middle of an upload operation, in which case we expect the low-level retry handler to kick in with the credential-refresh flow.

That said, it's always possible there's some pathological case here for the main upload operation which isn't being covered by the low-level request, but I'm not sure a high-level retry to call `execute` on the same upload-initializing request would work, and whether it might even be dangerous by reusing some lost upload state (e.g. closed channels).

Do you have evidence of uploads failing at that level (e.g. not being handled by low-level retries) that seemed to have evidence of being retriable?

And also what the behavior of re-calling `execute()` is when underlying byte channels aren't rewound somehow?

My naive expectation of upload requests would be that each `execute()` call would open a completely brand-new upload session, and that any server-buffered upload data uploaded in the previous session would be lost.

ravwojdyla:
Hi @dennishuo . I see - thanks for fast response and insights. For an example of the issue please take a look at these dataflow jobs:
 * 2017-06-08_00_21_58-6541194219975362799
 * 2017-06-05_08_40_33-5853187527764663530

and you should be able to see the exceptions like:

```
java.io.IOException: com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable
Service Unavailable
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.waitForCompletionAndThrowIfUploadFailed(AbstractGoogleAsyncWriteChannel.java:431)
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.close(AbstractGoogleAsyncWriteChannel.java:289)
	at org.apache.beam.sdk.io.FileBasedSink$FileBasedWriter.close(FileBasedSink.java:655)
	at org.apache.beam.sdk.io.FileBasedSink$FileBasedWriter.close(FileBasedSink.java:546)
	at org.apache.beam.sdk.io.Write$WriteBundles.finishBundle(Write.java:239)
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable
Service Unavailable
	at com.google.api.client.googleapis.json.GoogleJsonResponseException.from(GoogleJsonResponseException.java:146)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:113)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:40)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:432)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:469)
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel$UploadOperation.call(AbstractGoogleAsyncWriteChannel.java:357)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

In this case we are dealing with Dataflow job which reads from Bigtable and outputs to 6 different locations ~55K partitions (part files) each (determined by the bundle size). Our suspicion is that it might cause issues on the GCS (probably closing temporary objects). We have validated that lowering number of output-files/shards solves the problem - but at a cost of *expensive* shuffle (to enforce number of shards) [0]. I understand your concerns about the `upload`'s retry, it's something that we would need to test/validate. We have tested a version of a job with retries (logic from this PR). Dataflow job id: 2017-06-09_01_22_39-6953577886261981916. We see slightly different error:

```
java.io.IOException: com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable
Service Unavailable
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.waitForCompletionAndThrowIfUploadFailed(AbstractGoogleAsyncWriteChannel.java:444)
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.close(AbstractGoogleAsyncWriteChannel.java:296)
	at java.nio.channels.Channels$1.close(Channels.java:178)
	at com.google.cloud.dataflow.worker.repackaged.com.google.common.io.CountingOutputStream.close(CountingOutputStream.java:68)
	at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.close(IsmSink.java:281)
	at com.google.cloud.dataflow.worker.util.common.worker.WriteOperation.finish(WriteOperation.java:89)
	at com.google.cloud.dataflow.worker.util.common.worker.MapTaskExecutor.execute(MapTaskExecutor.java:80)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowWorker.executeWork(DataflowWorker.java:336)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowWorker.doWork(DataflowWorker.java:295)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowWorker.getAndPerformWork(DataflowWorker.java:242)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowBatchWorkerHarness$WorkerThread.doWork(DataflowBatchWorkerHarness.java:123)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowBatchWorkerHarness$WorkerThread.call(DataflowBatchWorkerHarness.java:103)
	at com.google.cloud.dataflow.worker.runners.worker.DataflowBatchWorkerHarness$WorkerThread.call(DataflowBatchWorkerHarness.java:90)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
	Suppressed: java.nio.channels.ClosedChannelException
		at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.throwIfNotOpen(AbstractGoogleAsyncWriteChannel.java:421)
		at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.write(AbstractGoogleAsyncWriteChannel.java:261)
		at java.nio.channels.Channels.writeFullyImpl(Channels.java:78)
		at java.nio.channels.Channels.writeFully(Channels.java:101)
		at java.nio.channels.Channels.access$000(Channels.java:61)
		at java.nio.channels.Channels$1.write(Channels.java:174)
		at java.io.OutputStream.write(OutputStream.java:75)
		at java.nio.channels.Channels$1.write(Channels.java:155)
		at com.google.cloud.dataflow.worker.repackaged.com.google.common.io.CountingOutputStream.write(CountingOutputStream.java:59)
		at java.io.DataOutputStream.writeByte(DataOutputStream.java:153)
		at com.google.cloud.dataflow.worker.repackaged.com.google.common.hash.BloomFilter.writeTo(BloomFilter.java:476)
		at com.google.cloud.dataflow.worker.util.ScalableBloomFilter$ScalableBloomFilterCoder.encode(ScalableBloomFilter.java:70)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.finish(IsmSink.java:268)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.close(IsmSink.java:280)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.abort(IsmSink.java:286)
		at com.google.cloud.dataflow.worker.util.common.worker.WriteOperation.abort(WriteOperation.java:98)
		at com.google.cloud.dataflow.worker.util.common.worker.MapTaskExecutor.execute(MapTaskExecutor.java:86)
		... 10 more
	Suppressed: java.nio.channels.ClosedChannelException
		at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.throwIfNotOpen(AbstractGoogleAsyncWriteChannel.java:421)
		at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel.write(AbstractGoogleAsyncWriteChannel.java:261)
		at java.nio.channels.Channels.writeFullyImpl(Channels.java:78)
		at java.nio.channels.Channels.writeFully(Channels.java:101)
		at java.nio.channels.Channels.access$000(Channels.java:61)
		at java.nio.channels.Channels$1.write(Channels.java:174)
		at java.io.OutputStream.write(OutputStream.java:75)
		at java.nio.channels.Channels$1.write(Channels.java:155)
		at com.google.cloud.dataflow.worker.repackaged.com.google.common.io.CountingOutputStream.write(CountingOutputStream.java:59)
		at java.io.DataOutputStream.writeByte(DataOutputStream.java:153)
		at com.google.cloud.dataflow.worker.repackaged.com.google.common.hash.BloomFilter.writeTo(BloomFilter.java:476)
		at com.google.cloud.dataflow.worker.util.ScalableBloomFilter$ScalableBloomFilterCoder.encode(ScalableBloomFilter.java:70)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.finish(IsmSink.java:268)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.close(IsmSink.java:280)
		at com.google.cloud.dataflow.worker.runners.worker.IsmSink$IsmSinkWriter.abort(IsmSink.java:286)
		at com.google.cloud.dataflow.worker.util.common.worker.WriteOperation.abort(WriteOperation.java:98)
		at com.google.cloud.dataflow.worker.util.common.worker.MapTaskExecutor.execute(MapTaskExecutor.java:86)
		... 10 more
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable
Service Unavailable
	at com.google.api.client.googleapis.json.GoogleJsonResponseException.from(GoogleJsonResponseException.java:146)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:113)
	at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:40)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:432)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:469)
	at com.google.cloud.hadoop.util.ResilientOperation$AbstractGoogleClientRequestExecutor.call(ResilientOperation.java:166)
	at com.google.cloud.hadoop.util.ResilientOperation.retry(ResilientOperation.java:66)
	at com.google.cloud.hadoop.util.AbstractGoogleAsyncWriteChannel$UploadOperation.call(AbstractGoogleAsyncWriteChannel.java:367)
	... 4 more
```

which seems to prove your concern, that said the job finished successfully, might this be the overhead of extra failure from `ResilientOperation` - at this point I can only hypothesize. I would like to see further:
 * if uploads are currently properly retried (specifically on 503)
 * if there is a need for better retry for uploads

I'm more then happy to adjust this PR, to whatever we believe is the best way to go further. Do you have an specific recommendations to look further?

[0] there might be a way for use to enforce larger bundles on the Bigtable reads thus force less shards without the cost of shuffle, we are currently evaluating this. That said, I believe this issue is orthogonal to that effort.

dennishuo:
Yeah, the fundamental problem here is that for these kinds of streaming uploads, either the chunked upload logic handles all retries, or you do real end-to-end retries either by having the entire set of data buffered somewhere or regenerating it all from scratch.

For more reference, the [AbstractGoogleClientRequest](http://grepcode.com/file/repo1.maven.org/maven2/com.google.api-client/google-api-client/1.20.0/com/google/api/client/googleapis/services/AbstractGoogleClientRequest.java/#428) is calling into the [MediaHttpUploader](http://grepcode.com/file/repo1.maven.org/maven2/com.google.api-client/google-api-client/1.20.0/com/google/api/client/googleapis/media/MediaHttpUploader.java/#407)'s while-loop. You can see how it's just calling individual chunks of [HttpRequest.execute()](http://grepcode.com/file/repo1.maven.org/maven2/com.google.api-client/google-api-client/1.20.0/com/google/api/client/googleapis/media/MediaHttpUploader.java#545), and inside there, as I linked before, the low-level retry logic should already be in place.

> if uploads are currently properly retried (specifically on 503)
> if there is a need for better retry for uploads

To address what you want to see, ideally the low-level client library would also log unsuccessful HTTP codes the way it [logs IOExceptions](http://grepcode.com/file/repo1.maven.org/maven2/com.google.http-client/google-http-client/1.20.0/com/google/api/client/http/HttpRequest.java#993)--that line is why sometimes you might see  a sstring of "WARNING: SocketTimeoutExceptions" in a log next to an otherwise successful job; those warnings are indeed harmless since they just meant the retry kicked in successfully. Below those lines there's equivalent `unsuccessfulResponseHandler` calls for things like 503 HTTP codes, but indeed there's no logging in those cases, unfortunately.

An option would be to instrument better logging to know when low-level retries are already happening. If we find evidence that it's already retrying say 20 times under the hood over tens of minutes, then it's questionable whether higher-level immediate retries would help anyways.

The pure-Hadoop version of things uses this setup: https://github.com/GoogleCloudPlatform/bigdata-interop/blob/097b5178977f56c1414e65c23881a8861eec4d09/util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java#L181 which indeed has a logging interceptor which would log `HttpStatus.SC_SERVICE_UNAVAILABLE` which is `503`. However, it looks like you're using the Dataflow SDK version here, so we're looking for this logging logic instead: http://grepcode.com/file/repo1.maven.org/maven2/com.google.cloud.dataflow/google-cloud-dataflow-java-sdk-all/0.4.150727/com/google/cloud/dataflow/sdk/util/RetryHttpRequestInitializer.java/#110

That appears to be set to DEBUG level, so if you set logging for `com.google.cloud.dataflow.sdk.util.RetryHttpRequestInitializer=DEBUG` you might be able to look for that message "Request failed with code {} will retry: {}".

Finally, to answer your questions:

> have you experienced/seen errors like this on GCS?

I don't usually see these 503 errors; as you mentioned, it's likely related to the hotspotting of lots of files in a similar location at the same time. Even if it's not these kinds of 503 errors though, there's always fundamentally going to be failure mores where a backend retry isn't possible; for example, a `410 gone` error will occur (rarely, normally only once out of many tens of thousands or less) which means that the GCS-managed dropzone for the data being uploaded before it's been committed to persistent replicas was lost, possibly due to hardware failures. This can only happen mid-upload, since on commit/completion the data is guaranteed to be persisted to many replicas in different locations.

In general though, yeah hotspotting issues with large numbers of files can result in higher error rates. That brings us to...

> what's the recommended way to restart failed uploads to GCS?

The current design intentionally doesn't rely on any client-side buffering; for example, you can generate/upload a 1TB file in GCS even if your entire machine only has 10GB of disk and 4GB of RAM. This is ideal for most situations, since wasting memory or disk just raises more problems. However, it also means if the upload session is irrecoverably lost, you have no choice but to regenerate the data from scratch.

In the context of most data processing jobs in both Hadoop and Dataflow, this shouldn't be a problem as long as such failures are rare; it's probably best in such cases for the high-level framework to do an actual clean top-level retry anyways, and then it doesn't matter whether it was a poisoned upload session or if the worker node had faulty hardware. In your case, the question would be why these types of complete upload failures occurs too often for Dataflow's orchestration to still allow jobs to be successful with complete worker retries.

An alternative is to implement complete disk or memory buffering in the GCS connector. Essentially we could provide an option where the write channel is a pointer to a local file handle, and then we have a separate thread upload from the local file into GCS. Then if the GCS upload fails, it can do a complete hard retry from the beginning of the local file again. The main problems to consider here would be making sure cleanup is reliable so you don't leak disk space, and also to then keep close accounting of available disk/memory space. Performance would also suffer significantly.

> do you have an specific recommendations to look into further?

1. See if you can get debug logs verifying whether low-level retries are happening. If they're not happening, something's wrong with the RetryHttpInitializer setup. If they are happening, then you could try to find the settings for cranking up the number of retries, or it might just indicate that the number of files you're using is problematic and you should get in contact with the GCS team to help fix this on GCS service's side.
2. We could consider a file/memory-buffering flavor of the GCS upload logic; if you're up for it you could take a stab at implementing it.

PR #55, opened 2017-06-22
Support as many parts of Application Default Credentials as safely possible

pmkc:
This allows using Cloud SDK credentials with  google.cloud.auth.service.account.json.keyfile (even though it's not a service account keyfile).

### DIFF
--- a/bigquery/bigquery-hadoop1.pom.xml
+++ b/bigquery/bigquery-hadoop1.pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigquery-connector-parent</artifactId>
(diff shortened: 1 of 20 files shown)


bigquery/bigquery-hadoop1.pom.xml
```diff
@@ -30,7 +30,7 @@
   <description>
     Hadoop 2 MapReduce input and output formats for use with BigQuery
   </description>
-  <version>0.10.3-SNAPSHOT-hadoop1</version>
+  <version>0.11.3-SNAPSHOT-hadoop1</version>
```
AngusDavis: 0.11.0

AngusDavis:
LGTM

liyinan926:
@pmkc I'm wondering which release this is going into and when it will go out?

PR #67, opened 2017-09-08
Fix version upgrade

pmkc:
I really broke this in #55


pmkc:
@AngusDavis is the expert (or at least author) of the build. I think we should fix this breakage now, and possibly file an issue for tracking future improvements to the build.

AngusDavis:
AFAIK, it gets tricky when a single version is not used throughout. The script used to set versions should probably be made to be open-source compatible and included in this repo.

PR #71, opened 2017-10-26
Use Java's Optional instead of Guava's

medb:


medb:
Yes, pom already references Java 8.

It will be nice to delete Hadoop 1 code/artifacts, looks like nobody really use them anyway:
https://www.google.com/search?q="bigdataoss"+"-hadoop1"+site%3Agithub.com+-site%3Agithub.com%2FGoogleCloudPlatform%2Fbigdata-interop

I think it will be better to create 1.6 non-Java 8 branch when/if we will need it (from the last tagged commit in 1.6.x release).

PR #85, opened 2018-01-14
Fix rename and delete behavior to throw in certain cases

chemikadze:
Original HDFS rename implementation [returns](https://github.com/apache/hadoop/blob/release-2.9.2-RC0/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/protocol/ClientProtocol.java#L496:L508):

> true if successful, or false if the old name does not exists or or if the new name already belongs to the namespace.

In other cases exception is thrown.

This patch adds re-throwing in cases of access and server errors.

Some implementations, such as Apache Hive, expect rename to throw:
- https://github.com/apache/hive/blob/branch-2.1/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L2717
- https://github.com/apache/hive/blob/branch-2.1/ql/src/java/org/apache/hadoop/hive/ql/metadata/Hive.java#L2738

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -883,8 +883,8 @@ public abstract class GoogleHadoopFileSystemBase extends FileSystem
    *
    * @param src Source path.
    * @param dst Destination path.
-   * @return true if successful, or false if the old name does not exist
-   * or if the new name already belongs to the namespace.
+   * @return true if successful, or false if the old name does not exist or if the new name already
+   *     belongs to the namespace.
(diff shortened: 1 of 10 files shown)


chemikadze:
I signed it!

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -906,13 +907,10 @@ public boolean rename(Path src, Path dst) throws IOException {
     try {
       getGcsFs().rename(srcPath, dstPath);
```
functicons: Explain what kind of errors.

medb: It's already mentioned in java doc

PR #87, opened 2018-01-23
Use new API-specific GCS batch endpoint

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/BatchHelper.java
```diff
@@ -70,6 +70,8 @@ private BatchHelper(HttpRequestInitializer requestInitializer, Storage gcs,
       long maxRequestsPerBatch) {
     this.pendingBatchEntries = new LinkedList<>();
     this.batch = gcs.batch(requestInitializer);
+    // Force use of API-specific batch endpoint
+    this.batch.getBatchUrl().appendRawPath("/" + Storage.DEFAULT_SERVICE_PATH);
```
AngusDavis: Is there anything that can / should be done to guard this from happening in 1.7?

medb: This change is intended only for 1.6.x branch, it will not be ported to master/1.7 branch, so there no need to guard it in any way.

medb: Added if condition, in case somebody will use 1.6 GCS connector with newer GCS API libs.

AngusDavis:
LGTM

PR #93, opened 2018-03-09
Fix dependencies in 1.6 branch

medb:
Added Guava's Preconditions class to each module's test sources, because Truth uses methods that are not available in older Guava versions.

Here are jars diffs before and after this PR:
 - BigQuery connector - https://www.diffchecker.com/sTvskPAA
 - GCS connector - https://www.diffchecker.com/1o6MNBJn

### DIFF
--- a/bigquery/CHANGES.txt
+++ b/bigquery/CHANGES.txt
@@ -2,7 +2,8 @@
 
   1. POM updates for GCS connector 1.6.4.
 
-  2. Update all dependencies to latest versions.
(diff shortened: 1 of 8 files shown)


medb:
I [updated Guava](https://github.com/GoogleCloudPlatform/bigdata-interop/commit/e44c3d3038ac69ad574c3daa7adb15b2768deeee) to fix compilation error, other than that - no issues.

With this PR I wanted to align master and 1.6 branch as close as possible.

medb:
@AngusDavis reverted non-test dependencies update, PTAL

PR #94, opened 2018-03-15
Fix Spark Pub/Sub connector shading to leave PubsubMessage unshaded

pmkc:
This shades google-cloud-pubsub and unshades the generated Java from protos.

That is because we expose PubsubMessage in the generated Java package [com.google.pubsub.v1](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L30).


AngusDavis:
LGTM

PR #95, opened 2018-03-16
1.6.x minimize shaded jars fix

medb:
Shaded jars difference with previous connectors release (1.6.3/0.10.4):
- BQ connector - https://www.diffchecker.com/e6hKs5HR
- GCS connector - https://www.diffchecker.com/oRIE4hoH


medb:
@AngusDavis Added shaded jars diffs to description, PTAL

PR #97, opened 2018-03-30
Fix broken build in 1.6.x branch

medb:
Build was broken because some methods are absent in old Google API libraries


AngusDavis:
LGTM

PR #101, opened 2018-05-28
Retry on HTTP status code 410 Gone

fittecs:
GCS sometimes returns 'Backend Error' with status code 410, causing whole Spark job to fail


fittecs:
I'm sorry I missed it.

PR #108, opened 2018-07-02
Defer closeChannel in case of GoogleCloudStorageReadChannel backward

sidseth:
seek.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -31,6 +31,7 @@ import com.google.cloud.hadoop.util.RetryBoundedBackOff;
 import com.google.cloud.hadoop.util.RetryDeterminer;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+
 import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.EOFException;
@@ -74,6 +75,7 @@ public class GoogleCloudStorageReadChannel
 
   // Size of buffer to allocate for skipping bytes in-place when performing in-place seeks.
   @VisibleForTesting
+  // TODO: Make this configurable.
   static final int SKIP_BUFFER_SIZE = 8192;
 
   // Used to separate elements of a Content-Range
@@ -97,6 +99,7 @@ public class GoogleCloudStorageReadChannel
 
   // Current read position in the channel.
   private long currentPosition = -1;
+  private long lastKnownChannelPosition = -1;
 
   // When a caller calls position(long) to set stream position, we record the target position
   // and defer the actual seek operation until the caller tries to read from the channel.
@@ -381,6 +384,7 @@ public class GoogleCloudStorageReadChannel
         }
         totalBytesRead += numBytesRead;
         currentPosition += numBytesRead;
+        lastKnownChannelPosition = currentPosition;
 
         if (retriesAttempted != 0) {
           LOG.info("Success after {} retries on reading '{}'",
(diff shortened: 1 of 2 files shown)


sidseth:
A corporate CLA should already be on file. Had submitted it a couple weeks ago. Is there any way to track the status?

medb:
Thank you for taking time to investigate improvements to `GoogleCloudStorageReadChannel`!

Our team has improvements to `GoogleCloudStorageReadChannel` on the roadmap and we work work on some of them already, namely "fadvise" functionality.

Regarding this change, if I understood correctly, it seems that it's not enough to not close `readChannel` during backward seek, because we do not have functionality that can "rewind" `readChannel` to read bytes from the beginning, i.e `readChannel` can read only forward.

medb:
Thank you for explanation.

I already did significant refactoring of the `GoogleCloudStorageReadChannel` class in scope of working on fadvise feature and incorporated your fix. I expect to push it to GitHub after review, hopefully today or early next week.

I didn't write new test for this specific use case though, so it would be nice to submit your test after I push my changes.

sidseth:
Sounds good. Will update the PR with just the test after your commit goes in.
Any thoughts on how the first read performance can be improved, or at least reuse connections when 100s or 1000s of files are being opened once?

sidseth:
Also, @medb  - is it possible for you to check on the CLA, or point me to someone who could help with getting it resolved.

medb:
Slow first read could be related to the redundant GetObjectMetadata requests, I have fixed this in my PR that includes your fix and adaptive range reads implementation: https://github.com/GoogleCloudPlatform/bigdata-interop/pull/110

After this change, read channel still makes 1 GetObjectMetadata request in constructor, so delay could be there, I guess we can have flag to disable it, but it needed to get object size and encoding.

In what scenario 1,000s files are read only once? How client knows at what position read these files?

I will try to check CLA thing.

medb:
@sidseth I have checked CLA and you need to have all of next:
1. Your company Corporate CLA is accepted (I have checked all accepted Corporate CLAs and I don't see Hortonworks there)
2. You need to be authorized contributor by your company.
3. You need to link your corporate email to your GitHub account.
4. You need to configure git to use your corporate email to author git commits.

May you check with your legal team if you are in compliance with items 1 and 2?

medb:
@sidseth may you share test that you have used to detect 90-250 ms delay when files opened only once?

sidseth:
> Slow first read could be related to the redundant GetObjectMetadata requests

Most of the time here was in the executeMedia call.

> In what scenario 1,000s files are read only once? How client knows at what position to read these files?

This is for split generation when running a Hive on Tez or Hive on MR job with ORC. Rather than one read per file, this is actually one read per stripe (which is typically about 64MB). As an example, if you look at data which is partitioned by day - and there's a query which accesses 5 years worth of data - split generation could end up accessing over 1800 files. This data is also accessed to potentially eliminate a bunch of work further down based on metadata in the stripes. Seems more of a connection re-establishment cost than anything else.

> @sidseth may you share test that you use to detect 90-250 ms delay when opening ORC files only once?

Let me put together a quick test for this. It doesn't necessarily need to be ORC file content. Will share this by tomorrow.

Thanks for the info on the CLA. Checking with relevant folks on getting this sorted out.

sidseth:
The access pattern is quite simple to reproduce.
Generate a bunch of files. Perform the following on each of them (ORC by default will launch 10 threads to read the stripe details from these files)

byte[] buffer = new byte[16384];
FileSystem fs = FileSystem.get(new URI(pathString), conf);
InputStream is = fs.open(new Path(pathString));
((FSDataInputStream) is).readFully(fileLength-16K, buffer, 0, 16384);

https://github.com/rajeshbalamohan/hadoop-aws-wrapper is useful in tracking various invocations in the readFully call. This may need some modifications. (Can post a PR to that repo tomorrow)

medb:
Thank you for explanation, I will take a look into this.

I have a new change that makes connector to send just 1 GCS request instead of 4 requests that it sends now when client opens a new stream and reads its footer.

I have tested it with Parquet. From my limited testing, ORC tries to pref-etch last 16 KiB of file when it reads footer, but in my case it was not enough and it end up issuing additional request to get reminder of the footer, so this change should improve reading of ORC files too.

medb:
This issue should have been fixed in https://github.com/GoogleCloudPlatform/bigdata-interop/commit/7f9c52c9664e987de41638a6c49e75b0c6ba4da3

Please, re-base your PR on master branch and I will merge it as soon as CLA is ready.

medb:
Merged: https://github.com/GoogleCloudPlatform/bigdata-interop/commit/39a2c4e0a14f49045c03b392e71688f1a2bd4dd1

PR #112, opened 2018-07-11
Report the UGI user in FileStatus instead of process owner.

sidseth:
Hadoop uses UserGroupInformation all over the place to keep track of the current user.

Cloud Storage can't really report any actual owner. However, instead of reporting the unix process owner in FileStatus - there's some small benefits of reporting the UGI user. 
1. There's a bunch of places where permission checks are done. For processes which run on behalf of other users (HiveServer2, Oozie) - these checks can end up failing. A workaround is to set reported permissions to 777.
2. The other HCFS implementations work with UGI - it's good to have consistent behaviour across these.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -73,6 +73,7 @@ import org.apache.hadoop.fs.GlobPattern;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
 import org.apache.hadoop.util.Progressable;
 import org.slf4j.Logger;
@@ -108,9 +109,6 @@ public abstract class GoogleHadoopFileSystemBase extends GoogleHadoopFileSystemB
 
   /** Default value of replication factor. */
   public static final short REPLICATION_FACTOR_DEFAULT = 3;
-
-  /** We report this value as a file's owner/group name. */
-  private static final String USER_NAME = System.getProperty("user.name");
 
   /** Splitter for list values stored in a single configuration value */
   private static final Splitter CONFIGURATION_SPLITTER = Splitter.on(',');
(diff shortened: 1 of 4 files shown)


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -117,4 +117,9 @@ public void testFileSystemIsRemovedFromCacheOnClose()
   public void testConfigurablePermissions()
       throws IOException {
   }
+
+  @Test @Override
```
medb: Please, replace with:
```
public void testFileStatusUser() throws IOException {}
```

medb:
Also, please add a change note to [CHANGES.txt](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/gcs/CHANGES.txt)

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1682,9 +1681,11 @@ public Path getHomeDirectory() {
   /**
    * Gets FileStatus corresponding to the given FileInfo value.
    */
-  private FileStatus getFileStatus(FileInfo fileInfo) {
+  private FileStatus getFileStatus(FileInfo fileInfo) throws IOException {
     // GCS does not provide modification time. It only provides creation time.
     // It works for objects because they are immutable once created.
+    UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
```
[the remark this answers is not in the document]
sidseth: I've changed the patch to reduce the number of invocations for list APIs. I'm not sure how big a performance penalty this is. For HDFS - the access(..) method invokes this each time. Getting the current user is also standard practice all over hadoop code.

sidseth:
Thanks for taking a look. Updated the PR, and rebased to master. Just got a confirmation that the CLA has been signed. (I signed it! - in case this triggers a recheck)

medb:
I see Hortonworks in corporate CLAs signers now.

I guess you need to post standalone "I signed it!" comment to trigger check by bot.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1665,6 +1666,12 @@ String trimToPrefixWithoutGlob(String path) {
     return fileStatuses;
   }
 
+  /** Helper method to get the UGI short user name */
+  private String getUgiUserName() throws IOException {
```
medb: Make this method static - it doesn't access any object state.

medb:
Merged: https://github.com/GoogleCloudPlatform/bigdata-interop/commit/e538ae0e7db682640f823f12c1226c2972ee9b9e

PR #126, opened 2018-09-27
Support to add credentials via hadoop credential provider path

prashantgolash:
Pull request to support reading credential secrets from Hadoop credential provider. Currently, this is one of the ways how S3 credentials are read (integrated into Hadoop main trunk).

For details on Hadoop credential provider, please refer: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/CredentialProviderAPI.html.

The client will configure the credential in core-site like this:
<property>
     <name>google.cloud.auth.hadoop.credential.provider.path</name>
     <value>jceks://hdfs@random_namenode/user/random_user/test.jceks</value>
   </property>

### DIFF
--- a/util/pom.xml
+++ b/util/pom.xml
@@ -29,6 +29,52 @@
   <artifactId>util</artifactId>
   <version>1.9.8-SNAPSHOT</version>
 
+  <properties>
+    <hadoop.one.version>0.23.11</hadoop.one.version>
+  </properties>
+
+  <profiles>
+    <profile>
+      <id>hadoop1</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-common</artifactId>
+          <version>${hadoop.one.version}</version>
+          <scope>provided</scope>
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>hadoop2</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-common</artifactId>
+          <version>${hadoop.two.version}</version>
+          <scope>provided</scope>
+        </dependency>
+      </dependencies>
+    </profile>
+
+    <profile>
+      <id>hadoop3</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-common</artifactId>
+          <version>${hadoop.three.version}</version>
+          <scope>provided</scope>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
+
+
   <dependencies>
     <dependency>
       <groupId>com.google.api-client</groupId>
--- a/util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
@@ -49,11 +49,18 @@ import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.GoogleLogger;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.security.alias.CredentialProvider;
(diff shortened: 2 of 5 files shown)


util/pom.xml
```diff
@@ -29,6 +29,52 @@
   <artifactId>util</artifactId>
   <version>1.9.8-SNAPSHOT</version>
 
+  <properties>
+    <hadoop.one.version>0.23.11</hadoop.one.version>
```
medb: Do you still use this Hadoop version, or this is just to make it compile?

util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
```diff
@@ -287,6 +297,77 @@ public Credential getCredentialFromJsonKeyFile(
...
+   * Initializes OAuth2 credential from a Hadoop credential Provider path.
+   * More details around Hadoop credential provider
+   * <a href="https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/CredentialProviderAPI.html">
+   * Hadoop Credential </a>
+   *
+   * @param credentialProviderPath Credential provider.
+   * @param scopes List of well-formed desired scopes to use with the credential.
+   * @param transport The HttpTransport used for authorization
+   */
+  public Credential getCredentialFromCredentialProvider(
```
medb: Will it be easier just to use [Configuration#getPassword](https://hadoop.apache.org/docs/current/api/org/apache/hadoop/conf/Configuration.html#getPassword(java.lang.String)) instead?

prashantgolash:
I signed here.

prashantgolash:
I signed it!

util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
prashantgolash: Currently, Configuration#getPassword create/initializes credential provider for each call. In this case, if we have n entries, each call will try to load "n" number of times, hence I kept it like this.

util/pom.xml
prashantgolash: I looked at https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common to check what corresponds to V1 for hadoop-common. It looks that the timeline matched with other v1 for e.g. hadoop-core, but if you know any other version, I would be happy to change to that. (BTW, we don't use hadoop1)

util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
medb: Yes, but it should be called only once per GHFS instantiation, not for every API call, isn't it?

prashantgolash: Yes, that's right. This will go only during GHFS instantiation. "n" meant to be the number of attributes in credentials (private_key, private_key_id, .... email etc), which will also happen in all the mappers. If we are fine with that, I can make the change to use Configuration#getPassword.

prashantgolash: @medb - can you let me know your thoughts?

medb: I'm inclined toward using Configuration#getPassword, but I plan to refactor credential-related code to be able to use Configuration#getPassword [here](https://github.com/GoogleCloudPlatform/bigdata-interop/blob/cf5c004c4e29dd487cc5e9ec185c0b2d60a06fc6/util/src/main/java/com/google/cloud/hadoop/util/EntriesCredentialConfiguration.java#L303).

This refactoring could impact this PR, that's why I propose to wait for couple days for refactoring, before making decision on this PR.

prashantgolash: Sure. Thanks for letting me know on this.

medb: Sorry for the delay, started to work on it but it got deprioritized, will try to finish it in next couple days.

medb:
@prashantgolash may you take a look if https://github.com/GoogleCloudPlatform/bigdata-interop/pull/129 addresses your use case as well?

medb:
Closing as obsolete, feel free to re-open if you think that related functionality is still missing in GCS connector.

PR #129, opened 2018-10-11
Re-introduce hadoop credentials support when specifying credentials

sidseth:
directly via Hadoop Configuration.
Was reverted in commit 2b28ab1f50dbb2d8e4a48ae30496451dd41af676 to
maintain hadoop1 compatibility.

### DIFF
--- a/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialConfiguration.java
+++ b/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialConfiguration.java
@@ -32,6 +32,16 @@ import org.apache.hadoop.conf.Configuration;
 public class HadoopCredentialConfiguration
     extends EntriesCredentialConfiguration implements Configurable {
 
+  /** An adapter to use our Configuration object as the config object for our superclass.
+   *  Exists here for backward compatibility
+   */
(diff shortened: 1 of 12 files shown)


sidseth:
@medb - could you please take a look. This re-introduces hadoop credential support for the config based credentials.
The approach taken is similar to the one used for GoogleFileSystemSpecific - to separate hadoop1, hadoop2 and hadoop3 implementations.
An alternate is to use reflection for the getPassword calls.

util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialConfiguration.java
```diff
@@ -32,32 +32,6 @@
 public class HadoopCredentialConfiguration
     extends EntriesCredentialConfiguration implements Configurable {
 
-  /**
-   * An adapter to use our Configuration object as the config object
-   * for our superclass.
-   */
-  public static class ConfigurationEntriesAdapter
```
medb: Let's leave this class here for backward compatibility (even if this class will be empty) and make it extend Hadoop version specific classes.

medb:
Yes, we need to be more specific about our published API.

Looks good, thanks for contribution!

sidseth:
Thanks for taking a look and committing.

PR #130, opened 2018-10-15
GCS Connector override file support

sadineni:
Added new property to pass override json file path. Added support to read key/value pairs from file and override them in Cionfigurtation object.


sadineni:
Closing this as no longer needed

PR #131, opened 2018-10-16
Parallelize glob search algorithms.

### LINKED ISSUE #123: globStatus call runs very slowly from Hadoop to GCS compared with gsutil
Run a glob listing on a big tree of around 890K entries.

1. Using Hadoop:
   ```bash
   hadoop fs -ls gs://bucketabc/1???/1/1/small-1.txt
   ```
   takes  3+ minutes;

2. Using Spark
   ```scala
   spark.read.format("com.databricks.spark.csv")
       .option("header", "false").option("delimiter", "\t").option("inferSchema", "true").option("mode", "DROPMALFORMED")
       .load("gs://bucketabc/1???/1/1/small-1.txt")
   ```
   takes 7+ minutes

3. While gsutil:
   ```bash
   gsutil ls gs://bucketabc/1???/1/1/small-1.txt
   ```
   only takes 8 seconds. 

It turns out that the full glob pattern has not been pushed down to GCS by the GCS connector. 

This problem could have significant impact on queries on data sets with huge number of partitions for either query performance consideration and/or as result of huge number of distributed processing tasks like MR and Spark.

A simple prototype fix has been observed to be able to close the significant perf gap.

andrej-moravian:
First phase to fix #123

### DIFF
--- a/gcs/CHANGES.txt
+++ b/gcs/CHANGES.txt
@@ -1,5 +1,10 @@
 1.9.9 - 2018-XX-XX
 
+  1. Add an option for running flat and regular glob search algorithms in
+     parallel. Returns a result of an algorithm that finishes first and cancels
+     another algorithm.
+
+       fs.gs.glob.concurent.enable (default: false)
 
 1.9.8 - 2018-10-03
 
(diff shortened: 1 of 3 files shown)


medb:
LGTM
Thanks for contribution!

PR #132, opened 2018-10-16
GCS Connector overrides support

sadineni:


medb:
May you add a note to `gcs/CHANGES.txt` file?

PR #133, opened 2018-10-18
Move Version info log line to DEBUG level.

sidseth:


sidseth:
The GHFS version log line shows up for every hadoop fs command. Moving to DEBUG level to keep the hadoop fs output clean. cc @medb for review.

PR #139, opened 2018-12-14
Make it possible to set `fs.gs.system.bucket` to an empty string

jphalip:
Make it possible to set `fs.gs.system.bucket` to an empty string in the case where there is a root bucket.


medb:
Thank you for contribution!

PR #140, opened 2018-12-17
Set default for "fs.gs.working.dir" to "/"

jphalip:
The code comment on line 149 was already saying that "/" was the default value, but it was actually not effectively set. This avoids having to manually set the property.


medb:
Working directory should be initialized here if property is not set:
https://github.com/GoogleCloudPlatform/bigdata-interop/blob/eb8e5a9f7783cad9ad3a7efb94aa95e7833ffe6d/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java#L1691

Maybe we should fix comment instead? What use case are you trying to fix with setting it to `/`?

jphalip:
I started from a fresh custom Hadoop install (i.e. not using Dataproc), and got this warning:

```
$ hadoop fs -ls gs://MY_BUCKET
Dec 17, 2018 4:22:13 AM com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase configure
WARNING: No working directory configured, using default: 'gs://BUCKET/'
```

The warning only went away after explicitly setting the `fs.gs.working.dir` property to `/` in `core-site.xml`. So it didn't seem to be automatically set for me.

PR #146, opened 2019-01-18
Add kmsKeyName to GoogleCloudStorageWriteChannel

udim:
Hi!

I work on github.com/apache/beam.
This PR will allow us to create new GCS objects with user-specified KMS keys.

Beam usage of GoogleCloudStorageWriteChannel: https://github.com/apache/beam/blob/b953645ed6db837d24284d7fe1fe091e7309f821/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/util/GcsUtil.java#L417-L426

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
@@ -42,6 +42,7 @@ public class GoogleCloudStorageWriteChannel
   private final String objectName;
   private final ObjectWriteConditions writeConditions;
   private final Map<String, String> metadata;
+  private String kmsKeyName;
 
   private GoogleCloudStorageItemInfo completedItemInfo = null;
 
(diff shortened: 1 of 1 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -141,4 +146,15 @@ public void handleResponse(StorageObject response) {
...
   }
+
+  /**
+   * Sets kmsKeyName used in the object create request for this channel.
+   *
+   * Use before calling ``initialize``.
+   *
+   * @param kmsKeyName Name of Cloud KMS key to use to encrypt the newly created object.
+   */
+  public void setKmsKeyName(String kmsKeyName) {
```
medb: Why setter is necessary?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -72,6 +73,7 @@ public GoogleCloudStorageWriteChannel(
     this.objectName = objectName;
     this.writeConditions = writeConditions;
     this.metadata = objectMetadata;
+    this.kmsKeyName = null;
```
medb: You can not modify existing constructor in new subminor GCS connector version (we don't plan to release new minor version of GCS connector yet) because this will break existing clients, including older Beam SDK versions.

You need to create a new constructor with `kmsKeyName` parameter and pass null to it from this constructor.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
udim: This was necessary when you couldn't set this value from the constructor, but you can now.

udim:
Thanks for the quick review!
What's the release schedule for this project?

medb:
@udim we do not have a set schedule for the release. But I think that we will release new version in couple weeks. When do you need it?

medb:
@udim new connector version with this change was released today:
https://github.com/GoogleCloudPlatform/bigdata-interop/releases/tag/v1.9.12

udim:
@medb Thanks! In use in https://github.com/apache/beam/pull/7682

PR #147, opened 2019-02-06
Implement Hadoop File Sytem concat method

tomwhite:
Adds an implementation of Hadoop FileSystem's concat method using GCS compose.

I have run the unit tests successfully on a GCS bucket, and I have run manual tests that concat thousands of files together.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -63,6 +63,7 @@ import com.google.common.base.Strings;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.flogger.GoogleLogger;
@@ -91,6 +92,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
 import org.apache.commons.codec.binary.Hex;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.ContentSummary;
@@ -199,6 +202,9 @@ public abstract class GoogleHadoopFileSystemBase extends GoogleHadoopFileSystemB
 
   /** Identifies this version of the GoogleHadoopFileSystemBase library. */
   public static final String GHFS_ID;
+
+  /** The maximum number of objects that can be composed in one operation. */
+  public static final int MAX_COMPOSE_OBJECTS = 32;
 
   static {
     VERSION =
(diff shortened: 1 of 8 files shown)


tomwhite:
I've signed the CLA.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/contract/AbstractGoogleContractConcatTest.java
```diff
@@ -0,0 +1,36 @@
...
+    Path[] srcs = new Path[numFiles];
+    for (int i = 0; i < numFiles; i++) {
+      Path srcFile = new Path(testPath, "" + i);
+      blocks[i] = dataset(TEST_FILE_LEN, i, 255);
+      createFile(getFileSystem(), srcFile, true, blocks[i]);
+      srcs[i] = srcFile;
+    }
+    Path target = new Path(testPath, "target");
+
+    createFile(getFileSystem(), target, false, new byte[0]);
```
aniket486: For my understanding, does that mean Hadoop api requires target path to exists before call to concat?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -547,6 +547,7 @@ public GoogleCloudStorageItemInfo composeObjects(
       final StorageResourceId destination,
       CreateObjectOptions options)
       throws IOException {
+    Preconditions.checkArgument(sources.size() <= 32, "Cannot compose more than 32 sources");
```
aniket486: Use the same constant so that it can be updated in one place. Alternatively, we can remove this Precondition.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -819,6 +825,37 @@ public FSDataOutputStream append(Path hadoopPath, int bufferSize, Progressable p
...
+
+    Preconditions.checkArgument(!srcPaths.contains(trgPath), "target must not be contained in sources");
+
+    List<List<URI>> partitions = Lists.partition(srcPaths, MAX_COMPOSE_OBJECTS - 1);
+    logger.atFine().log("GHFS.concat: %s, %d partitions", trg, partitions.size());
+    for (List<URI> partition : partitions) {
+      // We need to include the target in the list of sources to compose since
+      // the GCSFS compose operation will overwrite the target, whereas the Hadoop
+      // concat operation appends to the target.
+      List<URI> sources = Lists.newArrayList(trgPath);
```
medb: What will happen if target object does not exist but it's included in first `sources` partition? Does Hadoop concat semantic requires that target file should exist?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -819,6 +825,37 @@ public FSDataOutputStream append(Path hadoopPath, int bufferSize, Progressable p
...
+    logger.atFine().log("GHFS.concat: %s, %d sources", trg, psrcs.length);
+
+    Preconditions.checkArgument(psrcs.length > 0, "must be at least one source");
+
+    URI trgPath = getGcsPath(trg);
+    List<URI> srcPaths = Arrays.stream(psrcs).map(this::getGcsPath).collect(Collectors.toList());
+
+    Preconditions.checkArgument(!srcPaths.contains(trgPath), "target must not be contained in sources");
+
+    List<List<URI>> partitions = Lists.partition(srcPaths, MAX_COMPOSE_OBJECTS - 1);
```
medb: I think that it will be better to move partitioning logic and `MAX_COMPOSE_OBJECTS` constant in the GCS FS - it could be useful for other projects that use GCSIO library directly.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/contract/AbstractGoogleContractConcatTest.java
tomwhite: Yes it does. The [javadoc](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/api/org/apache/hadoop/fs/FileSystem.html#concat(org.apache.hadoop.fs.Path,%20org.apache.hadoop.fs.Path[])) is not very clear on this point, but the HDFS implementation certainly requires that the target exists. There is also a filesystem contract test for exactly this case (`AbstractContractConcatTest#testConcatMissingTarget`).

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
tomwhite: I moved the constant to `GoogleCloudStorage` to do this. (Also suggested by @medb.)

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
tomwhite: The operation will fail in that case because target files doesn't exist. The target is required to exist - even if empty - for Hadoop FS concat to succeed.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
tomwhite: I've moved the constant to the `GoogleCloudStorage` class.

I'm not sure the partitioning logic should be moved however, since it is tied to the Hadoop filesystem semantics. In particular, the target needs to be included in the call to compose, which may not be true of other projects implementing a compose-like operation.

medb: Thanks! Agree, it could be not intuitive if we will re-define GCS compose semantic in GCS IO library.

medb:
Thank you for the contribution!

PR #149, opened 2019-02-08
Clean up Markdown formatting [skip ci]

mbrukman:
* use Markdown formatting instead of HTML formatting
* added link to search for all questions tagged with `google-cloud-dataproc` on SO


medb:
Thanks for contribution!

Issue #151, opened 2019-02-21
GCS connector sends a high volume of ListObjects requests, triggering GCS 429 throttling error


githubwua:
While enabling Implicit Directories can help list all file objects, it has the side effect of sending too many ListObjects requests to GCS.  When this happens, GCS returns the following error:

```
Caused by: com.google.cloud.hadoop.repackaged.gcs.com.google.api.client.googleapis.json.GoogleJsonResponseException: 429 unknown 
<html><head><meta http-equiv="content-type" content="text/html; charset=utf-8"/><title>Sorry...</title><style> body { font-family: verdana, arial, sans-serif; background-color: #fff; color: #000; }</style></head><body><div><table><tr><td><b><font face=sans-serif size=10><font color=#4285f4>G</font><font color=#ea4335>o</font><font color=#fbbc05>o</font><font color=#4285f4>g</font><font color=#34a853>l</font><font color=#ea4335>e</font></font></b></td><td style="text-align: left; vertical-align: bottom; padding-bottom: 15px; width: 50%"><div style="border-bottom: 1px solid #dfdfdf;">Sorry...</div></td></tr></table></div><div style="margin-left: 4em;"><h1>We're sorry...</h1><p>... but your computer or network may be sending automated queries. To protect our users, we can't process your request right now. 
```

GcsFuse does mention about this caveat here:
https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/semantics.md#implicit-directories

Can we please roll back the change?

zeldrinn:
I'm seeing an issue that may or may not be related, via GCP Dataproc. Specifically, I have a Spark job running on Dataproc that reads and writes an RDD from a GCS bucket containing implicit directories, and starting on Feb 15, certain interactions with the bucket appears to generate an enormous number of `ListObject` operations (rising to 40K req/sec for over an hour). As far as I can tell, this spike in `ListObject` operations seems to happen at the point at which the RDD starts being written to the GCS bucket (<bucket>/<implicit_rdd_dir>/part_00001, <bucket>/<implicit_rdd_dir>/part_00002, <bucket>/<implicit_rdd_dir>/part_00003, etc.).

[...]

PR #152, opened 2019-02-21
Break and stop the pagination if we have the results

Fokko:
Hi Googlers,

I would like to suggest the following optimization. Right now the while loop will continue to fetch results while we already have the max results. For example, will become very inefficient when you're looking if a directory exists, and it will fetch all the pages.

This caused our cluster to do 10th of thousands of requests per second, which became quite expensive.

Regards, Fokko

### DIFF
--- a/gcs/CHANGES.txt
+++ b/gcs/CHANGES.txt
@@ -6,6 +6,9 @@
 
   3. Do not parallelize GCS list requests, because it leads to too high QPS.
 
+  4. Fix bug when GCS connector lists all files in directory instead of
+     specified limit.
+
 
 1.9.14 - 2019-02-13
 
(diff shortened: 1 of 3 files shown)


Fokko:
I signed it!

medb:
Thank you for contribution.

I think that we mitigated spike in the list requests in newly released GCS connector 1.9.15 (see issue #151), but this will be worthwhile improvement anyways, please allow us some time to review it.

Fokko:
I signed it!

Fokko:
Thanks for the quick response. I'll try to fix the tests somewhere today.

Fokko:
Thanks @medb

PR #159, opened 2019-04-07
Add configuration property to override GCS root url for VPC-SC

jasonmar:
This is meant to be used to set GCS root URL to https://restricted.googleapis.com/ for VPC-SC
Because GCS does not accept `Host: restricted.googleapis.com` a HttpRequestInitializer is used to override the Host header on all requests.

### DIFF
--- a/gcs/CHANGES.txt
+++ b/gcs/CHANGES.txt
@@ -23,6 +23,10 @@
   9. Log HTTP `429 Too Many Requests` responses from GCS at 1 per 10 seconds
      rate.
 
+  10. Add configuration property to force usage of VPC-SC via restricted.googleapis.com:
+
+       `fs.gs.vpcsc.enable` (default: false)
+
 
 1.9.14 - 2019-02-13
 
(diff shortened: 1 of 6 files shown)


medb:
Hi Jason,

May you test if this will work in VPCSC environment (possible test described [here]( https://github.com/GoogleCloudPlatform/bigdata-interop/pull/158#issuecomment-480553527))

jasonmar:
The GCS connector uses `com.google.api.services.storage.Storage` rather than `com.google.cloud.storage.Storage` found in the provided example so in this PR I've attempted to replicate `StorageOptions.setHost` and `StorageOptions.setHeaderProvider` with `HttpRequestInitializer` - I will set up a test VPC and post results of the curl test.

jasonmar:
I tested with curl and it works - I will create a Java test too to verify Host header is being set correctly.

salrashid123:
@jasonmar  re: curl; thats good news.  

you can also just enable verbose logging w/ `java.util.logging` handler as well as within the request initializer.  her'es an example repo ive got w/ some samples

- https://github.com/salrashid123/gcpsamples#logging-1

the bit w/ the exponentialbackoff shows how to add in `request.setCurlLoggingEnabled` just to test

medb:
Overriding `host` header could have unintended consequences and not recommended.

Preferred approach is to properly configure DNS instead for VPC SC access:
https://cloud.google.com/vpc/docs/configure-private-google-access-hybrid#configuring-dns

PR #160, opened 2019-04-11
Add GCS connector properties required for on-prem installation

medb:


gcs/INSTALL.md
```diff
@@ -37,14 +37,41 @@ Once you have the JSON key file, you must add the following property to
...
   <value>/path/to/keyfile</value>
   <description>
-    The JSON key file of the service account used for GCS
-    access when google.cloud.auth.service.account.enable is true.
+      The JSON key file of the service account used for GCS
+      access when google.cloud.auth.service.account.enable is true.
   </description>
 </property>
+<property>
+  <name>fs.gs.impl</name>
```
pmkc: This should be discoverable using the service registry (same with Abstract?) though in practice it might not always work.

gcs/INSTALL.md
```diff
@@ -37,14 +37,41 @@ Once you have the JSON key file, you must add the following property to
...
+  </description>
+</property>
+<property>
+  <name>google.cloud.auth.service.account.enable</name>
+  <value>/path/to/keyfile</value>
+  <description>
+      Whether to use a service account for GCS authorization.
+      If set to `false` then GCE VM metadata service will be used
+      for authorization.
+  </description>
```
pmkc: Alternatively you can set GOOGLE_APPLICATION_CREDENTIALS=/path/to

Add that to the description.

PR #169, opened 2019-05-31
Fix for null pointer exception in rename method for src variable

mgasiorowski:
I also do:

- Google java format, 
- Tests,


medb:
Thank you for fixing this.

PR #170, opened 2019-05-31
Update gcs-connector INSTALL.md

natekupp:
I ran into this issue with transitive dependencies while trying to use `gcs-connector` with a fresh local version of Spark 2.4.3. Hope the proposed solution makes sense! I'm not sure if that's the most canonical URL to get shaded jar versions; please let me know if another URL would be preferred.

It also wasn't immediately clear to me whether I needed to have a local Hadoop installation in order to make this work; I added some text to clarify that point.


natekupp:
I signed it!

Issue #171, opened 2019-06-04
BigQuery Connector to use different projects for dataset ownership and load job execution


digger:
Currently BQ connector supports two parameters that set Project ID to use when loading data into a table:
1. BigQueryConfiguration.PROJECT_ID_KEY
2. BigQueryConfiguration.OUTPUT_PROJECT_ID_KEY

The connector always uses just one of the parameters, running load jobs under the the same project ID that the table belongs to, while BigQueryHelper utility class and BQ itself support using one project as the dataset owner and the other for billing / job execution. 

The connector code that chooses project ID to use is here:
https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/output/BigQueryOutputConfiguration.java#L263
and here:
https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/output/BigQueryOutputConfiguration.java#L286

digger:
Pull request #175 is a proposed fix.

PR #172, opened 2019-06-04
Support dedicated load job project ID in BigQuery connector (#171)

digger:
Adds support to configuring BQ load jobs with a project ID to be used for the load job execution, in addition to the project ID that qualifies the target output table.


digger:
Made my email public to resolve the CLA check failure.

digger:
Removed my alternative emails from the profile attempting to pass the cla check.

digger:
can't pass the cla check, will try creating a new pull request

PR #173, opened 2019-06-04
Support dedicated load job project ID in BigQuery connector (#171)

digger:
Adds support to configuring BQ load jobs with a project ID to be used for the load job execution, in addition to the project ID that qualifies the target output table.


medb:
Superseded by #175

PR #182, opened 2019-06-14
[WIP] Tests for GoogleCloudStorage

adam-stasiak:
Added more tests for GoogleCloudStorage.


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTest.java
```diff
@@ -140,8 +126,20 @@ protected int doHash(byte[] bytes) {
             PerformanceCachingGoogleCloudStorageOptions.builder()
                 .setListCachingEnabled(true)
                 .build());
+    ThrottledGoogleCloudStorage throttledGoogleCloudStorage =
```
medb: There not need to test `ThrottledGoogleCloudStorage` because it's used only for testing. Let's remove `throttledGoogleCloudStorage` and `throttledGoogleCloudStorageWithOneOperation `, but keep test cases if they increase test coverage for other `GoogleCloudStorage` implementations.

PR #208, opened 2019-07-08
Add BatchHelper.allRequestsAreSentInSingleBatch test

medb:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/BatchHelperTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private BatchHelper.Factory batchFactory;
+
+  @Before
+  public void setUp() {
+    batchFactory = new BatchHelper.Factory();
+  }
+
+  @Test
+  public void allRequestsAreSentInSingleBatch() throws IOException {
```
functicons: Can we make the test more readable, either add comments or refactor. I prefer the following pattern when writing test cases:

setup(x, y, z...)
run(x, y, z...)
verify(x, y, z...)

it should be easy to see the cause and effect between setup/run and verify, e.g., I setup batch size as 2, and I added 2 request, so I expect 1 batch request.

medb: Added comments.

There are 3 requests because requests interceptor intercepts not only single batch request, but also individual requests inside the batch request.

PR #222, opened 2019-07-20
Migrate GoogleCloudStorageOptions to AutoValue

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -423,11 +423,12 @@ public ParentTimestampUpdateIncludePredicate(
...
      * Determine if updating directory timestamps should be ignored.
-     * @return True if the directory timestamp should not be updated. False to indicate it should
-     * be updated.
+     *
+     * @return True if the directory timestamp should not be updated. False to indicate it should be
+     *     updated.
      */
     @Override
-    public boolean shouldUpdateTimestamp(URI uri) {
+    public boolean test(URI uri) {
```
functicons: Why do we rename it? I feel "test" is a bad name, unclear what it actually means.

medb: This is a `test` method in `java.util.function.Predicate` - I just removed custom predicate class and switched to provided by JDK

medb:
> Will this change break existing user code?

Yes, potentially it could break user's code, but for major version release (2.0.0) we don't care about this.

PR #228, opened 2019-07-26
Disable support for reading "Content-Encoding: gzip" files by default

mgasiorowski:
Implementation + tests

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -111,10 +111,6 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
 
   private static final String USER_PROJECT_FIELD_NAME = "userProject";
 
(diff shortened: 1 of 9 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -600,6 +604,12 @@ public SeekableByteChannel open(
         throw GoogleCloudStorageExceptions.getFileNotFoundException(
             resourceId.getBucketName(), resourceId.getObjectName());
       }
+
+      if (readOptions.getFastFailOnGzipEncoding()
```
medb: Let's move this check in GoogleCloudStorageReadChannel, it should determine during initialization if opened file is GZIP encoded

mgasiorowski: @medb I moved check to GoogleCloudStorageReadChannel and rewrite tests, check if it is ok, please.

medb:
Superseded by https://github.com/GoogleCloudPlatform/bigdata-interop/pull/232

PR #229, opened 2019-07-29
Add Cooperative Locking load test

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -626,20 +625,46 @@ public void rename(URI src, URI dst) throws IOException {
...
+        coopLockOp.ifPresent(CoopLockOperationRename::unlock);
+        coopLockOp = Optional.empty();
+      }
+    }
+
+    FileInfo dstParentInfo = dstParent == null ? null : fileInfo.get(2);
+
     // Throw if src is a file and dst == GCS_ROOT
     if (!srcInfo.isDirectory() && dst.equals(GCS_ROOT)) {
+      coopLockOp.ifPresent(CoopLockOperationRename::unlock);
```
aniket486: Avoid code duplication like this with try/finally? 
Alternatively, you can implement AutoLockable with AutoCloseable interface.

medb: Done.

PR #233, opened 2019-08-03
Fix infinite loop in record reader

medb:
There is a problem with infinite loop in record reader in two cases:

1. BigQuery returned 0 records, but didn't create 0-record file
2. First task attempt failed. In that case on the second task attempt record reader will wait indefinitely, but no files will appear.

This PR based on the https://github.com/GoogleCloudPlatform/bigdata-interop/pull/26 PR with additional review fixes.


medb:
> Can we add test cases for max attempts?

Done

PR #240, opened 2019-08-10
Wrap internal exceptions in public API

### LINKED ISSUE #239: BQ connector shouldn't throw shaded internal exceptions.
This is from a StackOverflow [question](https://stackoverflow.com/questions/57414647/why-is-dataproc-using-this-weird-shaded-version-of-the-json-package-and-how-do-i). I think the problem is that, BQ connector shouldn't throw shaded internal exceptions, because it made users hard to catch the exception. Instead it should create its own exception class, e.g., BqConnectorException and set the shaded internal exceptions as cause.

medb:
Fixes #239

### DIFF
--- a/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/BigQueryHelper.java
+++ b/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/BigQueryHelper.java
@@ -13,10 +13,10 @@
  */
 package com.google.cloud.hadoop.io.bigquery;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.flogger.LazyArgs.lazy;
 
 import com.google.api.services.bigquery.Bigquery;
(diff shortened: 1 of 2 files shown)


functicons:
I don't quite understand this change. In the stackoverflow question, the user was trying to catch `GoogleJsonResponseException`, but in your change, you didn't catch it.

medb:
> GoogleJsonResponseException

It addresses SO issue, because `GoogleJsonResponseException` is a subclass of `IOException`.

functicons:
> > GoogleJsonResponseException
> 
> It addresses SO issue, because `GoogleJsonResponseException` is a subclass of `IOException`.

But that means we don't need this change, the user can simply catch `IOException`.

medb:
> But that means we don't need this change, the user can simply catch `IOException`.

Technically we need, because this change will make immediately thrown exceptions consistent between shaded and non-shaded JARs.

AndyClausen:
Thank you <3

PR #243, opened 2019-08-14
Add an option to FSCK specific operation

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsck.java
```diff
@@ -31,28 +32,34 @@
...
  *
- * <p>Usage: <code>
- *   hadoop jar /usr/lib/hadoop/lib/gcs-connector.jar
- *       com.google.cloud.hadoop.fs.gcs.CoopLockFsck --rollForward gs://my-bucket
- * </code>
+ * <p>Usage:
+ *
+ * <pre>{@code
+ * hadoop jar /usr/lib/hadoop/lib/gcs-connector.jar  com.google.cloud.hadoop.fs.gcs.CoopLockFsck \
+ *     --rollForward gs://my-bucket <operation-id>
```
aniket486: Consider adding a shell script wrapper utility for this.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsck.java
```diff
@@ -73,17 +80,23 @@ public static void main(String[] args) throws Exception {
 
   @Override
   public int run(String[] args) throws Exception {
-    checkArgument(
-        args.length == 2, "2 arguments should be specified, but were: %s", Arrays.asList(args));
-
     String command = args[0];
     checkArgument(FSCK_COMMANDS.contains(command), "Unknown %s command, should be %s", command);
 
+    int argsCount = COMMAND_CHECK.equals(command) ? 2 : 3;
```
aniket486: nit s/argsCount/expectedNumArgs/g

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
```diff
@@ -114,6 +121,16 @@ public int run() throws IOException {
       return 0;
```
aniket486: Move this to start of the method.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsck.java
medb: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
medb: In this case it will not print operations status, which is a purpose of the check command.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsck.java
medb: Ack, it could be tricky though because we don't distribute any scripts with GCS connector and GCS connector jar path/location differs from environment to environment.

PR #247, opened 2019-08-16
Remove timestamp updater and use GCS object time attributes

medb:
Based on https://github.com/GoogleCloudPlatform/bigdata-interop/pull/223 PR with review fixes:

- Get creation and modification times from GCS
- Remove parent folder `TimestampUpdater` and related code
- Create `GcsItemInfoTestBuilder` for tests


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
```diff
@@ -1643,222 +1619,6 @@ public void testFileCreationSetsAttributes() throws IOException {
         .isEqualTo("value1".getBytes(StandardCharsets.UTF_8));
   }
 
-  @Test
-  public void testFileCreationUpdatesParentDirectoryModificationTimestamp()
```
functicons: I think we need to keep these tests, verify the parent dir mtime is not updated.

medb: Because, after this PR, we will not have anything that will update it and this is not out goal to not have them updated (essentially we rely on whatever GCS is doing) - I'm not sure that we should test something that we do not touch.

gcs/CHANGES.md
```diff
@@ -73,6 +73,12 @@
 
         fs.gs.inputstream.support.gzip.encoding.enable (default: false)
 
+1.  Remove parent directory timestamp update feature and related properties:
```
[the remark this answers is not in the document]
medb: Done.

PR #267, opened 2019-10-14
Log upload session ID for failed resumable uploads

mgasiorowski:


medb:
This can not be implemented this way because `httpResponseUnparsed ` object will be not available if exception was thrown.

I have added URL that includes `upolad_id` to `google-http-java-client`:
https://github.com/googleapis/google-http-java-client/pull/1002

PR #268, opened 2019-10-15
Implement Hadoop FileSystem.access(...) method

mgasiorowski:


medb:
Exactly the same already implemented in the Hadoop [`FileSystem.access`](https://github.com/apache/hadoop/blob/b6fc72a0250ac3f2341ebe8a14d19b073e6224c8/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2592) method.

PR #279, opened 2020-02-12
Handle recovery of failed operations that didn't create log file

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
```diff
@@ -363,6 +347,60 @@ private void deleteAndRenameToRepairRenameOperation(
...
+    URI globUri =
+        gcsFs.getPathCodec().getPath(bucketName, globPath, /* allowEmptyObjectName= */ false);
+    FileStatus[] operationLocks = ghfs.globStatus(new Path(globUri));
+    checkState(
+        operationLocks.length < 2,
+        "operation %s should not have more than one log file",
+        operationRecord.getOperationId());
+
+    // Log file not created - nothing to repair
+    if (operationLocks.length == 0) {
```
hongyegong: Do we wanna return false too if multiple logs found?

operationLocks looks like a hard-to-read name also since how it implies # of log files anyway.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
```diff
@@ -114,6 +114,7 @@ public int run() throws IOException {
             .map(this::getOperationLockIfExpiredUnchecked)
             .filter(Optional::isPresent)
             .map(Optional::get)
+            .filter(operation -> !unlockOperationIfNoLogUnchecked(operation))
```
hongyegong: Can we make this not double negative way?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
medb: No, multiple log files mean that there are something wrong and we should fail whole thing.

Fixed name.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/CoopLockFsckRunner.java
medb: Done.

PR #300, opened 2020-02-19
Add properties to customize API root URLs

medb:


medb:
> Code LGTM, but do we have tests for these credential options somewhere?

Good catch, added tests.

PR #313, opened 2020-02-26
Remove list requests caching in performance cache

medb:
Reason: caching list requests has a little benefit in Hadoop/Spark workloads

### DIFF
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -32,6 +32,9 @@
 
 1.  Use Slf4j backend by default for Google Flogger.
 
+1.  Remove list requests caching in the `PerformanceCachingGoogleCloudStorage`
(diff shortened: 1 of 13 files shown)


hongyegong:
Feel like we almost removed entire caching functionality? Then what's the point of keeping this wrapper class?

Also you meant caching has very little benefit on hadoop workload right?

medb:
> Feel like we almost removed entire caching functionality? Then what's the point of keeping this wrapper class?
> 
> Also you meant caching has very little benefit on hadoop workload right?

Currently it caches 2 types of requests:
- GCS get requests
- GCS list requests

This PR removes only caching of GCS list requests, which is more complex and less efficient than GCS get requests, but it preserves logic for caching GCS get requests.

PR #315, opened 2020-02-28
Do not cache non-existent (not found) items in performance cache.

hongyegong:
This fixes the error that happens when create a new file with caching enabled. More specifically, now the cache will fail whenever getting the item info before creating the object, because cache doesn't get refreshed after object creation and get method - as part of the creation steps with caching enabled - will still get the old status and fail essentially.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/PrefixMappedItemCache.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/PrefixMappedItemCache.java
@@ -13,22 +13,19 @@
  */
 package com.google.cloud.hadoop.gcsio;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Strings.nullToEmpty;
 import static java.util.Comparator.naturalOrder;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Ticker;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Comparator;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
-import java.util.NavigableMap;
 import java.util.SortedMap;
 import java.util.TreeMap;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -551,6 +551,11 @@
               <artifactId>auto-value</artifactId>
               <version>${google.auto-value.version}</version>
             </path>
+            <path>
```
medb: Revert - this is only necessary for IntelliJ because it has a bug.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTest.java
```diff
@@ -1247,6 +1247,107 @@ public void testUpdateItemInfoUpdatesMetadata() throws IOException {
         BYTE_ARRAY_EQUIVALENCE);
   }
 
+  @Test
+  public void testGetBeforeAndAfterCreateObject() throws IOException {
```
medb: May you add tests to cover object rename/move and modification?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTest.java
```diff
@@ -1247,6 +1247,141 @@ public void testUpdateItemInfoUpdatesMetadata() throws IOException {
...
+
+    GoogleCloudStorageItemInfo createdItemInfo = rawStorage.getItemInfo(resourceId);
+    assertThat(createdItemInfo.exists()).isTrue();
+    assertThat(createdItemInfo.getSize()).isEqualTo(128);
+
+    byte[] objectContent = writeObject(rawStorage, resourceId, /* objectSize= */ 256);
+
+    GoogleCloudStorageItemInfo overwrittenItemInfo = rawStorage.getItemInfo(resourceId);
+    assertThat(overwrittenItemInfo.exists()).isTrue();
+    assertThat(overwrittenItemInfo.getSize()).isEqualTo(256);
```
medb: I have improved tests a little and seems found a bug. This line is failing with perf cache. May you take a look into it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/PrefixMappedItemCache.java
```diff
@@ -92,6 +92,10 @@ public synchronized GoogleCloudStorageItemInfo getItem(StorageResourceId id) {
    * @return the overwritten item, null if no item was overwritten.
    */
   public synchronized GoogleCloudStorageItemInfo putItem(GoogleCloudStorageItemInfo item) {
+    if (!item.exists()) {
```
medb: If we already invalidate cache on creation do we still need this condition?

hongyegong: This can actually help assure the correct behavior when call createEmptyObject(s) on gcsStorage object - so that I dont need to override each of these methods.

PR #320, opened 2020-03-06
Improve GCS gRPC upload efficiency, reduce buffering complexity.

mfschwartz:
Improve GCS gRPC upload efficiency, reduce buffering complexity.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -59,8 +59,6 @@ public final class GoogleCloudStorageGrpcWriteChannel
   // Default GCS upload granularity.
   public static final int GCS_MINIMUM_CHUNK_SIZE = 256 * 1024;
 
-  // Chunk size to use.
-  protected volatile int uploadChunkSize = UPLOAD_CHUNK_SIZE_DEFAULT;
 
   private final Object object;
   private final ObjectWriteConditions writeConditions;
@@ -96,9 +94,6 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
   @Override
   public void setDirectUploadEnabled(boolean enableDirectUpload) {
-    // TODO(b/135137933): Figure out if we need to support non-resumable uploads.
-    // It doesn't look as if the connector uses it at all, but there's code
-    // to support it, so maybe I'm missing something.
     if (enableDirectUpload) {
       throw new UnsupportedOperationException("Direct upload not supported.");
     }
@@ -106,16 +101,8 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
   @Override
   public void setUploadChunkSize(int chunkSize) {
-    Preconditions.checkArgument(chunkSize > 0, "Upload chunk size must be great than 0.");
-    Preconditions.checkArgument(
-        chunkSize % GCS_MINIMUM_CHUNK_SIZE == 0,
-        "Upload chunk size must be a multiple of GCS_MINIMUM_CHUNK_SIZE");
-    if ((chunkSize > GCS_UPLOAD_GRANULARITY) && (chunkSize % GCS_UPLOAD_GRANULARITY != 0)) {
-      logger.atWarning().log(
-          "Upload chunk size should be a multiple of %s for best performance, got %s",
-          GCS_UPLOAD_GRANULARITY, chunkSize);
-    }
-    uploadChunkSize = chunkSize;
+    // No-op for GCS gRPC - we always upload using a stream of MAX_WRITE_CHUNK_BYTES sized messages.
+    Preconditions.checkArgument(chunkSize > 0, "Upload chunk size must be greater than 0.");
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -96,26 +93,15 @@ public GoogleCloudStorageGrpcWriteChannel(
 
   @Override
   public void setDirectUploadEnabled(boolean enableDirectUpload) {
-    // TODO(b/135137933): Figure out if we need to support non-resumable uploads.
```
medb: Why we don't need this TODO?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -96,26 +93,15 @@ public GoogleCloudStorageGrpcWriteChannel(
...
-    Preconditions.checkArgument(
-        chunkSize % GCS_MINIMUM_CHUNK_SIZE == 0,
-        "Upload chunk size must be a multiple of GCS_MINIMUM_CHUNK_SIZE");
-    if ((chunkSize > GCS_UPLOAD_GRANULARITY) && (chunkSize % GCS_UPLOAD_GRANULARITY != 0)) {
-      logger.atWarning().log(
-          "Upload chunk size should be a multiple of %s for best performance, got %s",
-          GCS_UPLOAD_GRANULARITY, chunkSize);
-    }
-    uploadChunkSize = chunkSize;
+    // No-op for GCS gRPC - we always upload using a stream of MAX_WRITE_CHUNK_BYTES sized messages.
```
medb: Is there a possibility that optimal chunk size will change in the future, and it will be handy to have an option to adjust it in the client?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -185,42 +171,28 @@ public Object call() throws IOException, InterruptedException {
...
 
       if (responseObserver.hasError()) {
+        // TODO(b/150892988): Support resuming an upload after a transient error as follows:
+        //  1. Build a wrapper class around the ByteString (or subclass it), and keep track of the
+        //     start offset of each chunk.
+        //  2. Build a list of chunks, and (a) use it for rewinding when restarting at last
+        //     committed offset (per call to getCommittedWriteSize); (b) remove chunks from the list
+        //     that have been persisted.
+        //  3. Limit the list size to some number (possibly flag-controlled) of sent chunks.
         logger.atSevere().withCause(responseObserver.getError()).log("Error:");
```
medb: Remove this log statement because this is an anti-pattern - we need to either throw or log, but not both.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: I propose we not deal with non-resumable uploads in this connector; it could help with latency for small object writes, but I expect that's not the common HIC case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: The optimal performance will come from sending messages whose size is the max allowed by GCS (2 MiB).

I think chunk size probably matters more for other HDFS cases, where you're dealing with different write quantum alignment requirements for different underlying FS implementations. With GCS, what you want is to write in some integral multiple of 512 KiB, and you want to do it in the fewest number of largest allowed messages - so, 2 MiB makes sense.

If something about that reasoning changes we could always change the implementation.

Aside: In general I prefer to add flexibility for such configuration once there's a concrete reason to do so, not because there's a possibility that it might be useful in the future.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: Updated in my forked repo (https://github.com/mfschwartz/hadoop-connectors/blob/grpc-2.0.0/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java). Should I generate a new pull request with that update, or how do you like to do this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
medb: I think we still need direct upload, there are use cases for it, especially when clients use GCSIO library directly, not through HDFS API.

Also, it will be hard to use GCS channel by default if it throws exception when user enable direct upload.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
medb: I think that we added this option in the first place because at some point optimal write chunk was 8 MiB for GCS but later it become 2 MiB, so seems like there are a real possibility that it could change in the future.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
medb: No, when you push this new commit to the `grpc-2.0.0` branch in you repo it automatically added to this PR.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: That was the behavior before my change; all I did was remove the TODO.

What does it mean to be a "direct upload"? The setter in the base class doesn't define what it means, and it's not clear to me it has anything to do with resumable uploads.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: I agree it can change over time, but I would suggest the way we deal with that is to update this implementation.
I don't see value in letting customers control it. They may not understand the implications and could set an inappropriate value and then complain when they get poor performance, and we'll get an escalation to investigate.

medb: I think direct upload implemented with `uploadType=multipart` parameter in JSON API:
https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload

Essentially upload when you don't need to create a separate resumable upload session - it saves at least one round-trip.

If this is/will be supported in gRPC API too, then I think we should have this TODO and address it before GA.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
medb: I think direct upload implemented with `uploadType=multipart` parameter in JSON API:
https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload

Essentially upload when you don't need to create a separate resumable upload session - it saves at least one round-trip.

If this is/will be supported in gRPC API too, then I think we should have this TODO and address it before GA.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
medb: The problem is that as soon as we ship connector in Dataproc image it's impossible to change binaries but we can change configuration if there will be a need, so we usually opt-in to have an option to adjust configuration.

Customer already can change a lot of options that will negatively impact performance, so I think this is a secondary concern.

In this case, if you are confident that we will not need to change this in foreseeable future (at least couple years) than I'm OK to leave it as is.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: Ah, I see. Ok, I think it's valid to let the customer control that we could do a non-resumable upload this way. Updated the TODO about that, and made the method no longer throw.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: I've only seen the optimal upload chunk size change once in the 10 years I've been on the GCS project ;-)

PR #327, opened 2020-03-13
Remove deprecated methods

medb:


util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -128,17 +128,11 @@ public void handleResponse(S response) throws IOException {}
...
   }
 
-  @Deprecated
-  public void setUploadBufferSize(int bufferSize) {
-    setUploadChunkSize(bufferSize);
-  }
-
-  /** Sets size of upload buffer used. */
-  public void setUploadChunkSize(int chunkSize) {
+  private static int validateUploadChunkSize(int chunkSize) {
```
hongyegong: Maybe getUploadChunkSize with validate logic inside? A method that begins with validate seems to return true/false.

medb: Yeah, agree this is confusing, but I want to move it inside `AsyncWriteChannelOptions` in follow up CL.

PR #330, opened 2020-03-16
Use new gRPC channel for each read and write channel.

BrandonY:
In order to allow read and write throughput to scale well with parallel reading and writing, give each read and write channel its own, separate gRPC channel and stub.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -351,6 +244,202 @@ protected GoogleCloudStorageImpl() {
     this.storageOptions = GoogleCloudStorageOptions.builder().setAppName("test-app").build();
   }
 
+  // A function to encode metadata map values
+  private static String encodeMetadataValues(byte[] bytes) {
```
medb: This seems like not strictly code move - it will complicate merging of gRPC code into other branches, please revert.

BrandonY: Ah, looks like I accidentally formatted the file with IntelliJ, which is apparently rather opinionated about where to put functions. Reverted all the ordering changes.

PR #334, opened 2020-03-17
Extract credentials configuration validation into options class

medb:


util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
```diff
@@ -254,131 +250,92 @@ public static Credential getCredentialFromMetadataServiceAccount()
...
-      List<String> scopes,
-      HttpTransport transport,
-      String tokenServerUrl)
-      throws IOException, GeneralSecurityException {
+  public Credential getCredentialFromPrivateKeyServiceAccount(
+      List<String> scopes, HttpTransport transport) throws IOException, GeneralSecurityException {
     logger.atFine().log(
-        "getCredentialFromPrivateKeyServiceAccount(%s, %s, %s)",
-        serviceAccountEmail, privateKeyFile, scopes);
+        "getCredentialFromPrivateKeyServiceAccount(%s, %s) for '%s' from '%s'",
```
hongyegong: Lets keep the log format consistent as above. method name style or human readable style, either one

util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
```diff
@@ -254,131 +250,92 @@ public static Credential getCredentialFromMetadataServiceAccount()
...
         new GoogleCredential.Builder()
             .setTransport(transport)
             .setJsonFactory(JSON_FACTORY)
-            .setServiceAccountId(serviceAccountEmail)
+            .setServiceAccountId(options.getServiceAccountEmail())
             .setServiceAccountScopes(scopes)
-            .setServiceAccountPrivateKey(privateKey)
-            .setServiceAccountPrivateKeyId(privateKeyId);
-    return new GoogleCredentialWithRetry(builder, tokenServerUrl);
+            .setServiceAccountPrivateKey(privateKeyFromPkcs8(options.getServiceAccountPrivateKey()))
```
hongyegong: why not move privateKeyFromPkcs8 into getServiceAccountPrivateKey?

medb: Options class responsible only for holding and validating user configuration - transformation is better done in the place where it needed, because there could be other use cases that don't need this transformation.

util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
medb: It's still consistent - number of method parameters changed, so I moved useful information outside of the method declaration in the log statement.

PR #338, opened 2020-03-18
Include additional detail in exception text if server returns respons…

mfschwartz:
…e to InsertObject with missing/empty bucket name


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -112,6 +112,10 @@ public void setUploadChunkSize(int chunkSize) {
 
   @Override
   public void handleResponse(Object response) {
+    if (response.getBucket().isEmpty()) {
```
medb: Use `checkArgument` instead:
```java
checkArgument(
    !response.getBucket().isEmpty(), "Got response from service with empty/missing bucketName: %s" response);
```

mfschwartz: Done

PR #341, opened 2020-03-25
Support CSEK encryption in GCS connector.

hongyegong:


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -200,10 +201,25 @@ public void testHttpHeadersProperties_multipleHeaders() {
     GoogleCloudStorageFileSystemOptions options =
         GoogleHadoopFileSystemConfiguration.getGcsFsOptionsBuilder(config).build();
 
+    String encryptionAlgorithm = options.getCloudStorageOptions().getEncryptionAlgorithm();
```
medb: Delete this line

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -623,7 +625,9 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() {
 
       @Override
       protected Storage.Objects.Get createRequest() throws IOException {
-        return configureRequest(super.createRequest(), resourceId.getBucketName());
+        Storage.Objects.Get getObject = super.createRequest();
+        setEncryptionHeaders(getObject);
+        return configureRequest(getObject, resourceId.getBucketName());
```
medb: Could it be done in `configureRequest` method? Seems like it  could be a good feet if it applies to all requests.

hongyegong: It actually doesn't necessarily apply to all requests. for example, list or delete requests don't required those encryption headers to show the result. Only get/insert/rewrite/open need these encryption headers.

gcs/CONFIGURATION.md
```diff
@@ -138,6 +138,19 @@
...
+    The encryption algorithm to use. For [CSEK](https://cloud.google.com/storage/docs/encryption/customer-supplied-keys)
+    You must use the value AES256.
+
+*   `fs.gs.encryption.key` (not set by default)
+
+    An RFC 4648 Base64-encoded string of the source object's AES-256 encryption key.
+
+*   `fs.gs.encryption.key.hash` (not set by default)
+
+    An RFC 4648 Base64-encoded string of the SHA256 hash of the source object's encryption key.
```
medb: Add a new section for these properties. Maybe *Encryption configuration*?

gcs/CONFIGURATION.md
```diff
@@ -138,6 +138,19 @@
     fs.gs.storage.http.headers.another-custom-header=another_custom_value
     ```
 
+*   `fs.gs.encryption.algorithm` (not set by default)
```
medb: Is there are possibility that these properties will be applicable to other non-CSEK encryption methods? If no, should we replace `encryption` with `csek` in the properties names? Or gsutil also uses `encryption` for configuration properties?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: Should we use if condition to distinguish these requests in the configuration method then? It's still could be easier to implement if all logic will be in one place?  Something like `request instanceof Insert`.

gcs/CONFIGURATION.md
hongyegong: yes, encryption_key is actually gsutil's name convention as well. I guess since there is no alternative encryption for CSEK so calling it encryption should be fine for now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: Done.

gcs/CONFIGURATION.md
```diff
@@ -138,6 +138,8 @@
     fs.gs.storage.http.headers.another-custom-header=another_custom_value
     ```
 
+### Encryption
```
medb: Mention CSEK here, so customers can find it easier.
Maybe like this:
```
### Encryption (CSEK)
```

gcs/CONFIGURATION.md
```diff
@@ -138,6 +138,21 @@
...
+
+*   `fs.gs.encryption.algorithm` (not set by default)
+
+    The encryption algorithm to use. For [CSEK](https://cloud.google.com/storage/docs/encryption/customer-supplied-keys)
+    You must use the value AES256.
```
medb: Change to:
> Only `AES256` value supported.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -14,9 +14,7 @@
 
 package com.google.cloud.hadoop.fs.gcs;
 
-import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.GCS_CONFIG_PREFIX;
-import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.GCS_HTTP_HEADERS;
-import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.GCS_ROOT_URL;
+import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.*;
```
medb: Do not use start import - it's not recommended in Google Java Format.
Or even better, just format this PR with Google Java Formatter

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -377,7 +377,8 @@ public WritableByteChannel create(final StorageResourceId resourceId, CreateObje
           @Override
           public Storage.Objects.Insert createRequest(InputStreamContent inputStream)
               throws IOException {
-            return configureRequest(super.createRequest(inputStream), resourceId.getBucketName());
+            Storage.Objects.Insert insertObject = super.createRequest(inputStream);
```
medb: Revert

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -623,7 +624,9 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() {
 
       @Override
       protected Storage.Objects.Get createRequest() throws IOException {
-        return configureRequest(super.createRequest(), resourceId.getBucketName());
+        Storage.Objects.Get getObject = super.createRequest();
+        setEncryptionHeaders(getObject);
```
medb: Revert

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2032,9 +2034,47 @@ public GoogleCloudStorageItemInfo composeObjects(
...
+    if (storageOptions.getEncryptionAlgorithm() != null && storageOptions.getEncryptionKey() != null
+        && storageOptions.getEncryptionKeyHash() != null) {
+      headers.set("x-goog-encryption-algorithm", storageOptions.getEncryptionAlgorithm());
+      headers.set("x-goog-encryption-key", storageOptions.getEncryptionKey());
+      headers.set("x-goog-encryption-key-sha256", storageOptions.getEncryptionKeyHash());
+    } else {
+      return;
+    }
+
+    request.setRequestHeaders(headers);
```
medb: Move inside if condition and delete `else` branch.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2032,9 +2034,47 @@ public GoogleCloudStorageItemInfo composeObjects(
...
+      setEncryptionHeaders(request);
+      setDecryptionHeaders(request);
+    }
+
     return request;
   }
 
+  private <RequestT extends StorageRequest<?>> void setEncryptionHeaders(RequestT request) {
+    HttpHeaders headers = request.getRequestHeaders();
+    if (storageOptions.getEncryptionAlgorithm() != null && storageOptions.getEncryptionKey() != null
```
medb: Add this validation to options class - if customer will specify only one property we should fail Options class creation.

gcs/CHANGES.md
```diff
@@ -1,9 +1,11 @@
 ### 2.2.0 - 2020-XX-XX
 
-1. Delete deprecated methods.
+1.  Delete deprecated methods.
 
 1.  Update all dependencies to latest versions.
 
+1.  Support CSEK headers to encrypt and decrypt objects.
```
medb: Add properties to release notes (see other release notes for an example)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -1015,7 +1015,6 @@ private void copyInternal(
         configureRequest(
             gcs.objects().copy(srcBucketName, srcObjectName, dstBucketName, dstObjectName, null),
             srcBucketName);
-
```
medb: Revert

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -247,6 +262,11 @@ public GoogleCloudStorageOptions build() {
       checkArgument(
           (instance.getProxyUsername() == null) == (instance.getProxyPassword() == null),
           "both proxyUsername and proxyPassword should be null or not null together");
+      checkArgument(
+          instance.getEncryptionAlgorithm() != null
```
medb: Should we check that either all of the encryption properties are set or neither of them are set?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2032,9 +2031,45 @@ public GoogleCloudStorageItemInfo composeObjects(
...
+    if (request instanceof Storage.Objects.Rewrite || request instanceof Storage.Objects.Copy) {
+      setEncryptionHeaders(request);
+      setDecryptionHeaders(request);
+    }
+
     return request;
   }
 
+  private <RequestT extends StorageRequest<?>> void setEncryptionHeaders(RequestT request) {
+    if (storageOptions.getEncryptionAlgorithm() != null
```
medb: With validation in options class, will it be sufficient to just check if `getEncryptionAlgorithm` is set here?

hongyegong: That's logically true and makes sense for us with context, but I guess for someone who just reads this code, I think this will make more sense.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -247,6 +262,15 @@ public GoogleCloudStorageOptions build() {
       checkArgument(
           (instance.getProxyUsername() == null) == (instance.getProxyPassword() == null),
           "both proxyUsername and proxyPassword should be null or not null together");
+      checkArgument(
+          (instance.getEncryptionAlgorithm() != null
```
medb: You can simplify this with XOR operator:
```java
checkArgument(
    !(instance.getEncryptionAlgorithm() == null
        ^ instance.getEncryptionKey() == null
        ^ instance.getEncryptionKeyHash() == null),
    "encryptionAlgorithm, encryptionKey and encryptionKeyHash should be null or not null"
              + " together");
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: You may consider using `checkNotNull` wrappers to the other values, when setting them instead checking in if condition, because it could hide validation bug in the options class:

```java
headers.set("x-goog-encryption-key", checkNotNull(storageOptions.getEncryptionKey()));
headers.set("x-goog-encryption-key-sha256", checkNotNull(storageOptions.getEncryptionKeyHash()));
```

medb:
> Is it possible to create a self-sufficient integration test that generates key, creates encrypted w/ CSEK bucket and runs couple test cases against it?

@hongyegong what about integration test?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
hongyegong: This actually won't work because true ^ true = false and then false ^ true = true and then !true = false which is wrong.

Remain the same logic, might not seem most concise but clear enough.

hongyegong:
> > Is it possible to create a self-sufficient integration test that generates key, creates encrypted w/ CSEK bucket and runs couple test cases against it?
> 
> @hongyegong what about integration test?

Added integrations to test insert/get/rewrite. right now hardcoded the encryption key (but functional). Might consider to add functions to generate key and hash automatically for each run.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -209,6 +209,62 @@ public void testCopySingleItemWithRewrite() throws IOException {
...
+  }
+
+  @Test
+  public void testRewriteEncryptedObject() throws IOException {
+    GoogleCloudStorageImpl gcs =
+        makeStorage(
+            GoogleCloudStorageTestHelper.getStandardOptionBuilder()
+                .setCopyWithRewriteEnabled(true)
+                .setMaxBytesRewrittenPerCall(512 * 1024 * 1024)
+                .setEncryptionAlgorithm("AES256")
```
medb: May you extract encryption configuration in common method and reuse it in these tests?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -209,6 +209,62 @@ public void testCopySingleItemWithRewrite() throws IOException {
     assertObjectContent(gcs, copiedResourceId, partition, partitionsCount);
   }
 
+  @Test
+  public void testUploadAndGetObjectWithEncryptionKey() throws IOException {
```
medb: May you extract these tests in a separate `CmekEncryptionTest` class - we have a bad habit to put all test in a single class, but it's getting harder and harder to maintain.

medb:
> Added integrations to test insert/get/rewrite. right now hardcoded the encryption key (but functional). Might consider to add functions to generate key and hash automatically for each run.

Does it require any project or GCS bucket set-up? Or anybody can run these tests against any bucket in any project?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
hongyegong: > > Added integrations to test insert/get/rewrite. right now hardcoded the encryption key (but functional). Might consider to add functions to generate key and hash automatically for each run.
> 
> Does it require any project or GCS bucket set-up? Or anybody can run these tests against any bucket in any project?

Unlike CMEK which requires a GCS key ring, CSEK doesn't need any preconfiguration as long as key and key hash match themseleves.

hongyegong: Done

hongyegong:
> > Added integrations to test insert/get/rewrite. right now hardcoded the encryption key (but functional). Might consider to add functions to generate key and hash automatically for each run.
> 
> Does it require any project or GCS bucket set-up? Or anybody can run these tests against any bucket in any project?

Unlike CMEK which requires a GCS key ring, CSEK doesn't need any preconfiguration as long as key and key hash match themseleves.

PR #346, opened 2020-03-31
Use RedactedString for CSEK properties

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -181,6 +182,41 @@ public void throwIfNotValid() {
...
+        + ("durationMillis: " + getMaxWaitMillisForEmptyObjectCreation() + '\n')
+        + ("maxListItemsPerCall: " + getMaxListItemsPerCall() + '\n')
+        + ("maxRequestsPerBatch: " + getMaxRequestsPerBatch() + '\n')
+        + ("batchThreads: " + getBatchThreads() + '\n')
+        + ("maxHttpRequestRetries: " + getMaxHttpRequestRetries() + '\n')
+        + ("httpRequestConnectTimeout: " + getHttpRequestConnectTimeout() + '\n')
+        + ("httpRequestReadTimeout: " + getHttpRequestReadTimeout() + '\n')
+        + ("transportType: " + getTransportType() + '\n')
+        + ("proxyAddress: " + getProxyAddress() + '\n')
+        + ("proxyUsername: " + toSecretString(getProxyUsername()) + '\n')
```
medb, line 203: While we did it this way in the past, I think it will be better to invest to implementing [less error-prone and more maintainable solution](https://github.com/google/auto/issues/132#issuecomment-54777277).

medb: Done.

PR #350, opened 2020-04-14
add retries and backoff for read operation

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -212,41 +252,110 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
-        messageHasher.putBytes(res.getChecksummedData().getContent().toByteArray());
-        int calculatedChecksum = messageHasher.hash().asInt();
-        int expectedChecksum = res.getChecksummedData().getCrc32C().getValue();
-        if (calculatedChecksum != expectedChecksum) {
-          throw new IOException(
-              String.format(
-                  "For %s: Message checksum didn't match. Expected %s, got %s.",
-                  this, expectedChecksum, calculatedChecksum));
+      long lastReadPosition = position;
+      while (lastReadPosition == position || retriesAttempted > 0) {
```
medb, line 257: I think that we should not implement a custom high-level retries, unless we are sure that there are some cases that are not covered by [gRPC Retry Design](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).

hongyegong: I think this is different level rather than a high level of retry? This PR is more like a client i/o failure retry, similar to something we have for non-gRPC read channel now. The link you linked above is gRPC level retry, caused by server failure, v.s. http level retry for json API.

medb: If there are retries on gRPC level, what failures we are trying to address by this additional retries? I.e what kind of client I/O failures are not covered by gRPC retries?

hongyegong: I will reconsider the how much this retry could help potential read failures and put this PR under draft until it's fully figured.

PR #351, opened 2020-04-16
Resume write upon tansient error with BufferedInputStream

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -161,13 +165,18 @@ public void startUpload(PipedInputStream pipeSource) {
...
+    private final InputStream pipeSource;
+    private PushbackInputStream pushbackPipeSource;
     private final GoogleCloudStorageGrpcWriteChannel grpcWriteChannel;
+    private final int MAX_BYTES_PER_MESSAGE = MAX_WRITE_CHUNK_BYTES.getNumber();
+    // Byte array to hold the unread data.
+    private byte[] pushbackBuffer;
 
-    UploadOperation(GoogleCloudStorageGrpcWriteChannel grpcWriteChannel, InputStream pipeSource) {
-      // Buffer the input stream by 1 byte so we can peek ahead for the end of the stream.
+    UploadOperation(GoogleCloudStorageGrpcWriteChannel grpcWriteChannel,
```
medb, line 175: Format with Google Java Formatter

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
```diff
@@ -309,6 +309,40 @@ public void writeHandlesErrorOnQueryWriteStatusRequest() throws Exception {
     writeChannel.write(data.asReadOnlyByteBuffer());
   }
 
+  @Test
+  public void writeHandlesErrorOnInsertRequestWithUnnommitedData() throws Exception {
```
medb, line 313: Typo: `Un*c*ommi*t*ted`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
```diff
@@ -309,6 +309,40 @@ public void writeHandlesErrorOnQueryWriteStatusRequest() throws Exception {
...
+
+    ByteString data = createTestData(chunkSize * 3 / 2);
+    writeChannel.initialize();
+    writeChannel.write(data.asReadOnlyByteBuffer());
+
+    assertThrows(IOException.class, writeChannel::close);
+  }
+
+  @Test
+  public void writeHandlesErrorOnInsertRequestWithoutUncommitedData() throws Exception {
```
medb, line 330: Can we add integration tests for this?

hongyegong: Seems like this is not an integration test file and we are missing integration tests for gRPC write/read. I will create another PR to draft integration tests for both read/write.

medb: ack

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -198,30 +207,41 @@ private Object doResumableUpload() throws IOException, InterruptedException {
...
-        //     committed offset (per call to getCommittedWriteSize); (b) remove chunks from the list
-        //     that have been persisted.
-        //  3. Limit the list size to some number (possibly flag-controlled) of sent chunks.
-        throw new IOException(
-            String.format("Insert failed for '%s'", resourceId), responseObserver.getError());
-      }
+        if (responseObserver.hasError()) {
+          // If error happens, wrap the source input stream with PushbackInputStream so that
+          // latest read data chunk can be re-read and insert during the next insert call.
+          this.pushbackPipeSource = new PushbackInputStream(pipeSource, MAX_BYTES_PER_MESSAGE);
```
medb, line 214: I think it will be simpler and less error-prone to use [`BufferedInputStream`](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedInputStream.html) with [`mark`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#mark-int-)/[`reset`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#reset--) methods instead of manually maintaining retry cache with `PushbackInputStream`.

hongyegong: I considered this option, mark/reset seems to be favorable if entire data chunk from last insert needs to be committed again. However, I didn't figure out how to use mark/reset handles the case where only partial of the last data chunk was committed. With pushbackInputStream I can achieve this by unreading length of uncommitted data from pushbackBuffer and there doesn't seem to be a way to do this easily with BufferedInputStream.

medb: To re-read part of the marked data, you can use [`skip`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#skip-long-) method to skip n-bytes to the 1st byte that you want to re-read and continue read from there.

hongyegong: Updated. I overlooked the skip() method. It does simplify the implementation.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -161,13 +164,16 @@ public void startUpload(PipedInputStream pipeSource) {
 
   private class UploadOperation implements Callable<Object> {
     // Read end of the pipe.
-    private final BufferedInputStream pipeSource;
+    private final InputStream pipeSource;
```
medb, line 167: Do we still need `pipeSource` as a field?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -281,13 +294,11 @@ public void run() {
                   if (checksumsEnabled) {
                     Hasher chunkHasher = Hashing.crc32c().newHasher();
                     for (ByteBuffer buffer : chunkData.asReadOnlyByteBufferList()) {
-                      chunkHasher.putBytes(buffer);
-                    }
-                    for (ByteBuffer buffer : chunkData.asReadOnlyByteBufferList()) {
+                      chunkHasher.putBytes(buffer.duplicate());
```
medb, line 297: Why do we need `duplicate` call?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: This was actually removed locally. Forgot to git add before commit ...

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: becuz putBytes actually modify the putBytes. That's also why it was in 2 for loops originally. Just figured this way might be clearer.

PR #352, opened 2020-04-17
Add property to override GCS gRPC service name

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
```diff
@@ -202,6 +208,11 @@ public static Builder builder() {
      */
     public abstract Builder setGrpcChecksumsEnabled(boolean grpcChecksumsEnabled);
 
+    /**
+     * Sets the property to override the default GCS gRPC service name.
```
medb, line 212: Format with Google Java Formatter

hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -51,7 +51,7 @@ public StorageStub buildAsyncStub() {
   }
 
   private Map<String, Object> getGrpcServiceConfig() {
-    Map<String, Object> name = ImmutableMap.of("service", "google.storage.v1.Storage");
+    Map<String, Object> name = ImmutableMap.of("service", readOptions.getGrpcServiceName());
```
medb, line 54: I think that we should override this instead:
https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/630a99d2c78bfb1d7647d11af8381f3e1e757093/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java#L40

hongyegong: updated the property to be hostname.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -37,7 +37,7 @@ public StorageStubProvider(
 
   public StorageBlockingStub buildBlockingStub() {
     return StorageGrpc.newBlockingStub(
-        ComputeEngineChannelBuilder.forAddress("storage.googleapis.com", 443)
+        ComputeEngineChannelBuilder.forAddress(readOptions.getGrpcHostname(), 443)
```
medb, line 40: There are `ComputeEngineChannelBuilder.forTarget(String)` method that seems to allow override not only a hostname but scheme and port too. Can we sue it instead to have a greater flexibility?

hongyegong: It's a wrapper of .forTarget method, which utilizes handy GrpcUtil.authorityFromHostAndPort method to build target string.
.forTarget doesn't gain extra flexibility for this use case.

PR #357, opened 2020-04-22
Balance gRPC connections with bpool algorithm.

BrandonY:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -34,9 +50,82 @@
...
+                  }
+                  super.onClose(status, trailers);
+                }
+              }, headers);
+        }
+      };
+    }
+  }
+
+  class ChannelAndRequestCounter {
```
hongyegong, line 116: Wouldn't this class make more sense if it's called ChannelWithRequestCounter, a subclass of ManagedChannel with a counter in initializer?  Seems like it's essentially a channel with a counter around it, and you don't need to write something like "channel.channel" down below.

medb: Let's address this issue in separate PR.

PR #361, opened 2020-04-28
Add option to support grpc auto flowcontrol window (branch-2.1.x-grpc)

WeiranFang:
- Add option `fs.gs.grpc.auto.window.enable` in `GoogleHadoopFileSystemConfiguration`.
- Manually set this feature by using `InternalHandlerSettings` when flag is enabled.
- This feature will be enabled by default in future grpc-java release. So this is a temporary change used for benchmarking test.


hongyegong:
I believe you will have to update GoogleHadoopFileSystemConfigurationTest default value as well, otherwise test won't pass.

PR #363, opened 2020-04-30
Remove one data copy on read path.

BrandonY:


veblush:
BTW can we do the same thing on [line 241](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/781ecd99d2094f90c978d52b8b2926b4c424b79d/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java#L241)?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -149,6 +149,14 @@ private static IOException convertError(
     }
   }
 
+  /** Writes part of a ByteString into a ByteBuffer with as little copying as possible */
+  private final static void put(ByteString source, int offset, int size, ByteBuffer dest) {
+    ByteString croppedSource = source.substring(offset, offset+size);
+    for( ByteBuffer sourcePiece : croppedSource.asReadOnlyByteBufferList()) {
```
hongyegong, line 155: Why asReadOnlyByteBuffer() doesn't fit given the bytestring of given `size` to write is guaranteed to fit in dest according to the logic before calling this method. This seems unnecessary iteration?

BrandonY: From the [ByteString doc](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString.html#asReadOnlyByteBufferList--):

> By returning a list, implementations of this method may be able to avoid copying even when there are multiple backing arrays.

I'm not sure if I'm actually saving any extra copying with this, but based on that comment it seems possible.

hongyegong: "multiple backing arrays" is the case I am not sure either. anyway Let's give it a try.

PR #370, opened 2020-05-05
Add retries for getUploadID and getCommitedSize

hongyegong:
Right now there are errors in dfsio write jobs that are caused by failure to getUploadId from server and we are missing retries for it. Haven't seen any data upload failure for a particular data chunk after retry for that case is added so it might also help to retry getUploadId erros to get rid of some of upload failures caused by this.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -69,7 +69,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
   private static final Duration WRITE_STREAM_TIMEOUT = Duration.ofSeconds(20);
   // Maximum number of automatic retries for each data chunk
   // when writing to underlying channel raises error.
-  private static final int WRITE_CHUNK_RETRIES = 10;
+  private static final int UPLOAD_RETRIES = 10;
 
   private final StorageStub stub;
   private final StorageResourceId resourceId;
@@ -182,9 +182,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
       do {
-        ByteString chunkData = ByteString.EMPTY;
-        responseObserver =
-            new InsertChunkResponseObserver(uploadId, chunkData, writeOffset, objectHasher);
+        responseObserver = new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher);
         // TODO(b/151184800): Implement per-message timeout, in addition to stream timeout.
         stub.withDeadlineAfter(WRITE_STREAM_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
             .insertObject(responseObserver);
@@ -207,7 +205,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
           retriesAttempted = 0;
         }
 
-        if (retriesAttempted >= WRITE_CHUNK_RETRIES) {
+        if (retriesAttempted >= UPLOAD_RETRIES) {
           throw new IOException(
               String.format("Insert failed for '%s'", resourceId), responseObserver.getError());
         }
@@ -234,10 +232,9 @@ public final class GoogleCloudStorageGrpcWriteChannel
       // stream is closed.
       final CountDownLatch done = new CountDownLatch(1);
 
-      InsertChunkResponseObserver(
-          String uploadId, ByteString chunkData, long writeOffset, Hasher objectHasher) {
+      InsertChunkResponseObserver(String uploadId, long writeOffset, Hasher objectHasher) {
         this.uploadId = uploadId;
-        this.chunkData = chunkData;
+        this.chunkData = ByteString.EMPTY;
         this.writeOffset = writeOffset;
         this.objectHasher = objectHasher;
       }
@@ -357,6 +354,23 @@ public final class GoogleCloudStorageGrpcWriteChannel
       }
     }
 
+    private void runWithRetries(Runnable block, SimpleResponseObserver responseObserver)
+        throws IOException {
+      for (int attemptedRetries = 0; ; attemptedRetries++) {
+        block.run();
+        if (responseObserver.hasError()) {
+          if (attemptedRetries >= UPLOAD_RETRIES) {
+            throw new IOException(
+                String.format("Failed to start resumable upload for '%s'", resourceId),
+                responseObserver.getError());
+          }
+          continue;
+        } else {
+          break;
+        }
+      }
+    }
+
     /** Send a StartResumableWriteRequest and return the uploadId of the resumable write. */
     private String startResumableUpload() throws InterruptedException, IOException {
       InsertObjectSpec.Builder insertObjectSpecBuilder =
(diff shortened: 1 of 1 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -357,6 +354,23 @@ public void onCompleted() {
       }
     }
 
+    private void runWithRetries(Runnable block, SimpleResponseObserver responseObserver)
+        throws IOException {
+      for (int attemptedRetries = 0; ; attemptedRetries++) {
+        block.run();
+        if (responseObserver.hasError()) {
+          if (attemptedRetries >= UPLOAD_RETRIES) {
```
medb, line 362: Move this condition in the `for` loop and throw exception outside the loop, and return from the method if there are no errors:
```java
    private void runWithRetries(Runnable block, SimpleResponseObserver responseObserver)
        throws IOException {
      for (int attemptedRetries = 0; attemptedRetries < UPLOAD_RETRIES; attemptedRetries++) {
        block.run();
        if (!responseObserver.hasError()) {
          return;
        }
      }
      throw new IOException(
          String.format("Failed to start resumable upload for '%s'", resourceId),
          responseObserver.getError());
    }
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -357,6 +354,23 @@ public void onCompleted() {
       }
     }
 
+    private void runWithRetries(Runnable block, SimpleResponseObserver responseObserver)
+        throws IOException {
+      for (int attemptedRetries = 0; ; attemptedRetries++) {
+        block.run();
```
medb, line 360: Will this update status of the `responseObserver` object?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -386,16 +400,18 @@ private String startResumableUpload() throws InterruptedException, IOException {
...
-      }
+      runWithRetries(
+          () -> {
+            stub.withDeadlineAfter(START_RESUMABLE_WRITE_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
+                .startResumableWrite(request, responseObserver);
+            try {
+              responseObserver.done.await();
+            } catch (InterruptedException e) {
+              // Do not throw here since it will eventually throw if all retries fail.
+              e.printStackTrace();
```
medb, line 411: `InterruptedException` is different than `IOException` and usually this means that such exception should not be swallowed, you need to handle it like this: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/00e3f553ff49a189518206b4d531f6a9906b4d09/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java#L942:L945

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -386,16 +400,18 @@ private String startResumableUpload() throws InterruptedException, IOException {
...
-      stub.withDeadlineAfter(START_RESUMABLE_WRITE_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
-          .startResumableWrite(request, responseObserver);
-      responseObserver.done.await();
-
-      if (responseObserver.hasError()) {
-        throw new IOException(
-            String.format("Failed to start resumable upload for '%s'", resourceId),
-            responseObserver.getError());
-      }
+      runWithRetries(
```
medb, line 403: Can you use [ResilientOperation](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/util/src/main/java/com/google/cloud/hadoop/util/ResilientOperation.java) instead?

hongyegong: Done. It doesn't reduce amount of codes but I guess it's good to have nice features like backoff etc.,

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: N/A anymore, but I think it should because the object in lambda is by reference.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Deleted.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -407,18 +423,65 @@ private long getCommittedWriteSize(String uploadId) throws InterruptedException,
...
-      responseObserver.done.await();
+      return responseObserver.getResponse().getCommittedSize();
+    }
 
-      if (responseObserver.hasError()) {
-        throw new IOException(
-            String.format("Failed to get committed write size '%s'", resourceId),
-            responseObserver.getError());
+    /** Simple class to create a {@link CheckedCallable} from a {@link QueryWriteStatusRequest}. */
+    private class GrpcQueryWriteStatusRequestExecutor<T>
```
medb, line 444: Can you use lambda instead of creating custom class to execute retries?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -78,6 +84,16 @@
...
+  // Helper delegate for turning IOExceptions from API calls into higher-level semantics.
+  private ApiErrorExtractor errorExtractor = ApiErrorExtractor.INSTANCE;
+  // Determine if a given IOException is due to rate-limiting.
+  private RetryDeterminer<IOException> rateLimitedRetryDeterminer =
+      RetryDeterminer.createRateLimitedRetryDeterminer(errorExtractor);
+  // Object to use to perform sleep operations
+  private Sleeper sleeper = Sleeper.DEFAULT;
+  // BackOff objects are per-request, use this to make new ones.
+  private BackOffFactory backOffFactory = BackOffFactory.DEFAULT;
+
```
medb, line 96: If you initialize these fields with the same values all the time, then you can just inline them in the call site. Or even better, you can create `ResilientOperation.retry` method that uses default values.

hongyegong: Done. Left rateLimitedRetryDeterminer to be there because it needs more code to initialize and can be shared btw multiple methods.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Doubt I can do so, the retry method requires a CheckedCallable to be passed as a wrapper of the code to be executed. Secondly CheckedCallable actually provide a good generic for exceptions to be thrown, which is a little bit messy to do with lambda.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -78,6 +84,12 @@
   private final ImmutableMap<String, String> metadata;
   private final boolean checksumsEnabled;
 
+  // Helper delegate for turning IOExceptions from API calls into higher-level semantics.
+  private ApiErrorExtractor errorExtractor = ApiErrorExtractor.INSTANCE;
+  // Determine if a given IOException is due to rate-limiting.
+  private RetryDeterminer<IOException> rateLimitedRetryDeterminer =
```
medb, line 90: I recalled that this will not work because `ApiErrorExtractor` relies on parsing of JSON API responses. Let's revert to your initial approach with retry method.

PR #376, opened 2020-05-07
Temporary disabled GrpcAutoWindowEnabled

veblush:
`InternalHandlerSettings` is not part of gRPC Java public API so it's discouraging to use it in the gcsio release. So this will be disabled temporary.


veblush:
Instead of this, let's roll back https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/361 for simplicity.

medb:
The simplest will be to leave it as is :)

PR #380, opened 2020-05-11
Enable grpc retry mechanism

hongyegong:
enableRetry() should be called on ChannelBuilder to enable grpc retry mechanism. Right now there are a lot of "UNAVAILABLE" error status that should be eliminated by enabling this retry.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
@@ -143,6 +143,7 @@ public class StorageStubProvider {
                 isNullOrEmpty(readOptions.getGrpcServerAddress())
                     ? DEFAULT_GCS_GRPC_SERVER_ADDRESS
                     : readOptions.getGrpcServerAddress())
+            .enableRetry()
             .defaultServiceConfig(getGrpcServiceConfig())
             .intercept(counter)
             .build();


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -143,6 +143,7 @@ private ChannelAndRequestCounter buildManagedChannel() {
                 isNullOrEmpty(readOptions.getGrpcServerAddress())
                     ? DEFAULT_GCS_GRPC_SERVER_ADDRESS
                     : readOptions.getGrpcServerAddress())
+            .enableRetry()
```
medb, line 146: Are there any options that we can configure for retries? Like number of retries, retry delay, etc?

hongyegong: This is configured in method getGrpcServiceConfig() down below. However without setting this, I believe the settings down there would be no-op after I looked through grpc-java impl.

medb: Cool!
Can we delete some high-level retries that we added in java code? Or they are still needed?

hongyegong: Potentially yes. Now we are unclear about what error status write requests (that we added retries for), e.g. queryWriteStatus, startResumableWrite return upon errors. Our current grpc retry config only covers retries for error statuses UNAVAILABLE/RESOURCE_EXHAUSTED. I would rather remove any retries we have now or add it under grpc retry policy after doing some experiments.

medb: Ack.

PR #382, opened 2020-05-13
Clear data cache of current request in cancelRequest call

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -285,6 +285,10 @@ private void requestObjectMedia(OptionalInt bytesToRead) throws IOException {
   }
 
   private void cancelCurrentRequest() {
+    this.bufferedContent = null;
```
medb, line 288: Usually we use `this.` prefix only in constructors and setters, please remove if possible.

PR #384, opened 2020-05-15
Raise errors in getResponse() for failed calls.

BrandonY:
In the gRPC write channel, if the server returns an error, the client currently ends up attempting to read a nonexistent response message and failing out. This change has it instead raise an IOException with a meaningful error message.

It might make sense to return the gRPC status error directly, but wrapping it allows us to add useful additional error details, like the name of the object and the upload ID.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -181,11 +181,6 @@ public final class GoogleCloudStorageGrpcWriteChannel
       // the writer at the other end will not hang indefinitely.
       try (InputStream toClose = pipeSource) {
         return doResumableUpload();
-      } catch (IOException e) {
-        throw e;
-      } catch (Exception e) {
-        throw new IOException(
-            String.format("Caught exception during upload for '%s'", resourceId), e);
       }
     }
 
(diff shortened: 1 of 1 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -325,7 +328,10 @@ private ByteString readRequestData() throws IOException {
             });
       }
 
-      public Object getResponse() {
+      public Object getResponse() throws IOException {
+	if (hasError()) {
```
hongyegong, line 332: I think instead of/in addition to doing this here, you also want to add this check above at while (!responseObserver.isFinished()), to not let while loop stop upon error. Instead it should only stop if and only if it finishes without error or maxmimum retry reached in which case exception is thrown.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -175,18 +176,20 @@ public void startUpload(PipedInputStream pipeSource) {
...
     @Override
-    public Object call() throws IOException, InterruptedException {
+    public Object call() throws IOException {
       // Try-with-resource will close this end of the pipe so that
       // the writer at the other end will not hang indefinitely.
       try (InputStream toClose = pipeSource) {
         return doResumableUpload();
+      } catch (IOException e) {
+        throw e;
       } catch (Exception e) {
```
medb, line 186: Are there any non-IOExceptions that thrown in `doResumableUpload` method by gRPC?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
BrandonY: Ah, good point. I'll have it check only whether the upload was finalized instead.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
BrandonY: Not to my knowledge. I can remove the clause.

PR #386, opened 2020-05-16
remove autowindow settings from connector

hongyegong:
Removed autowindow settings from connector bcuz it will be by default to true in grpc library from next release. Plus it's an internal API that should not be called directly from client, which could even cause problem if directly call it based on [this comment](https://github.com/grpc/grpc-java/issues/7007#issuecomment-625368425).

Also suspect that the fact this brings extra grpc internal package dependency may cause the problem when working/building with hadoop job somehow, which causes some `class not found` exception, given the fact adding back this grpc internal settings is one of the only 3 changes that are made after GCS connector 2.1.3 (above exception is not seen) and before the GCS connector RC7 (above exception is seen).


medb:
> You should remove this property from `GoogleHadoopFileSystemConfigurationTest` too. But how presubmit passed if this property wasn't removed from the test?

nvm, missed it in PR

PR #391, opened 2020-06-04
Add an option to use Java NIO Pipe in Cloud Storage output stream

### LINKED ISSUE #104: Support Apache Accumulo on GCS
Presumably, Apache Accumulo requires random file access which is not efficiently handled by GCS connector right now.

Relevant discussion: https://github.com/GoogleCloudPlatform/bigdata-interop/issues/103

medb:
In contrast to `PipedInputStream`/`PipedOutputStream` pipe, Java NIO `Pipe` supports multiple writers and does not throw *"Pipe broken"* exception in this case (see #103 for an example).

Note that according to `TestDFSIO` HCFS test Java NIO `Pipe` has 10% lower throughput when writing to GCS than current implementation - that's why Java NIO `Pipe` is not used by default.

Fixes #104 (?)

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -289,7 +289,7 @@ public class GoogleHadoopFileSystemConfiguration {
 
   /** Configuration key for setting pipe type. */
   public static final HadoopConfigurationProperty<PipeType> GCS_OUTPUT_STREAM_PIPE_TYPE =
-      new HadoopConfigurationProperty<>("fs.gs.outputstream.pipe.type", PipeType.NIO_CHANNEL_PIPE);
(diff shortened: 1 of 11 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -286,6 +287,10 @@
   public static final HadoopConfigurationProperty<Integer> GCS_OUTPUT_STREAM_PIPE_BUFFER_SIZE =
       new HadoopConfigurationProperty<>("fs.gs.outputstream.pipe.buffer.size", 1024 * 1024);
 
+  /** Configuration key for setting pipe type. */
+  public static final HadoopConfigurationProperty<PipeType> GCS_OUTPUT_STREAM_PIPE_TYPE =
+      new HadoopConfigurationProperty<>("fs.gs.outputstream.pipe.type", PipeType.NIO_CHANNEL_PIPE);
```
hongyegong, line 292: Shouldn't default be IO_STREAM_PIPE?

medb: Good catch, yes, it should be.

PR #400, opened 2020-06-11
Support per-user/group service account impersonation

hongyegong:
This change enables users to specify service accounts to impersonate keyed by user name or group name, when accessing GCS, mainly to facilitate GCS connector based multi-tenancy.

### DIFF
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -68,6 +68,20 @@
     (`roles/iam.serviceAccountTokenCreator`) on the service account to
     impersonate.
 
+ 1.  Add properties to impersonate a service account through user or group name:
+
+     ```
+     fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
+     fs.gs.auth.group.impersonation.<GROUP_NAME> (not set by default)
+     ```
+
+     If any of these properties is set, an access token will be generated for
+     the service account associated with specified user name or group name in
+     order to access GCS. The caller who issues a request for the access token
+     must have been granted the Service Account Token Creator role
+     (`roles/iam.serviceAccountTokenCreator`) on the service account to
+     impersonate.
+
 ### 2.1.1 - 2020-03-11
 
 1.  Add upload cache to support high-level retries of failed uploads. Cache size
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -75,8 +75,8 @@
 1.  Add properties to impersonate a service account through user or group name:
 
     ```
-    fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
-    fs.gs.auth.group.impersonation.<GROUP_NAME> (not set by default)
+    fs.gs.auth.impersonation.service.account.user.<USER_NAME> (not set by default)
+    fs.gs.auth.impersonation.service.account.group.<GROUP_NAME> (not set by default)
     ```
 
     If any of these properties is set, an access token will be generated for
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -68,19 +68,23 @@
     (`roles/iam.serviceAccountTokenCreator`) on the service account to
     impersonate.
 
- 1.  Add properties to impersonate a service account through user or group name:
+1.  Throw `ClosedChannelException` in `GoogleHadoopOutputStream.write` methods
+    if stream already closed. This fixes Spark Streaming jobs checkpointing to
+    Cloud Storage.
 
-     ```
-     fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
-     fs.gs.auth.group.impersonation.<GROUP_NAME> (not set by default)
-     ```
+1.  Add properties to impersonate a service account through user or group name:
 
-     If any of these properties is set, an access token will be generated for
-     the service account associated with specified user name or group name in
-     order to access GCS. The caller who issues a request for the access token
-     must have been granted the Service Account Token Creator role
-     (`roles/iam.serviceAccountTokenCreator`) on the service account to
-     impersonate.
+    ```
+    fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
+    fs.gs.auth.group.impersonation.<GROUP_NAME> (not set by default)
+    ```
+
+    If any of these properties is set, an access token will be generated for
+    the service account associated with specified user name or group name in
+    order to access GCS. The caller who issues a request for the access token
+    must have been granted the Service Account Token Creator role
+    (`roles/iam.serviceAccountTokenCreator`) on the service account to
+    impersonate.
 
 ### 2.1.1 - 2020-03-11
 
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -75,8 +75,8 @@
 1.  Add properties to impersonate a service account through user or group name:
 
     ```
-    fs.gs.auth.impersonation.service.account.user.<USER_NAME> (not set by default)
-    fs.gs.auth.impersonation.service.account.group.<GROUP_NAME> (not set by default)
+    fs.gs.auth.impersonation.service.account.for.user.<USER_NAME> (not set by default)
+    fs.gs.auth.impersonation.service.account.for.group.<GROUP_NAME> (not set by default)
     ```
 
     If any of these properties is set, an access token will be generated for
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -1508,21 +1508,21 @@ public abstract class GoogleHadoopFileSystemBase extends FileSystem
             .build()
             .getCloudStorageOptions();
 
-    String serviceAccountFromUserOrGroup =
-        getMatchedServiceAccountToImpersonateFromGroup(
-                options.getUserImpersonationServiceAccounts(),
-                ImmutableList.of(UserGroupInformation.getCurrentUser().getShortUserName()))
-            .orElse(
-                getMatchedServiceAccountToImpersonateFromGroup(
-                        options.getGroupImpersonationServiceAccounts(),
-                        ImmutableList.copyOf(UserGroupInformation.getCurrentUser().getGroupNames()))
-                    .orElse(null));
-
     String serviceAccountToImpersonate =
         IMPERSONATION_SERVICE_ACCOUNT_SUFFIX
             .withPrefixes(CONFIG_KEY_PREFIXES)
             .get(config, config::get);
     if (isNullOrEmpty(serviceAccountToImpersonate)) {
+      String serviceAccountFromUserOrGroup =
+          getMatchedServiceAccountToImpersonateFromGroup(
+                  options.getUserImpersonationServiceAccounts(),
+                  ImmutableList.of(UserGroupInformation.getCurrentUser().getShortUserName()))
+              .orElse(
+                  getMatchedServiceAccountToImpersonateFromGroup(
+                          options.getGroupImpersonationServiceAccounts(),
+                          ImmutableList.copyOf(
+                              UserGroupInformation.getCurrentUser().getGroupNames()))
+                      .orElse(null));
       serviceAccountToImpersonate = serviceAccountFromUserOrGroup;
     }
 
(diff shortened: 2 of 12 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1531,6 +1545,21 @@ private Credential getCredential(Configuration config)
...
+   */
+  private static boolean matchedUserOrGroup(String name) throws IOException {
+    String userName = UserGroupInformation.getCurrentUser().getShortUserName();
+    String[] groupNames = UserGroupInformation.getCurrentUser().getGroupNames();
+
+    if (userName.equals(name) || Arrays.asList(groupNames).contains(name)) {
+      return true;
+    }
+
+    return false;
```
medb, line 1560: Use simple `return` instead:
```
return userName.equals(name) || Arrays.asList(groupNames).contains(name);
```

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1531,6 +1545,21 @@ private Credential getCredential(Configuration config)
     return Optional.of(impersonatedCredential.createScoped(CredentialFactory.GCS_SCOPES));
   }
 
+  /**
+   * @param name login user name or group name
+   * @return true if name matches with currently login user or any group.
+   */
+  private static boolean matchedUserOrGroup(String name) throws IOException {
+    String userName = UserGroupInformation.getCurrentUser().getShortUserName();
+    String[] groupNames = UserGroupInformation.getCurrentUser().getGroupNames();
```
medb, line 1554: Aren't this a little bit risky if we allow such ambiguity in security configuration? Should we clearly distinguish between the two? Also, may you check how S3 is configuring this, so we will have as close user-facing configuration to them as possible (if it make sense)?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1508,6 +1509,19 @@ private Credential getCredential(Configuration config)
             .withPrefixes(CONFIG_KEY_PREFIXES)
             .get(config, config::get);
 
+    if (isNullOrEmpty(serviceAccountToImpersonate)) {
+      Map<String, String> serviceAccountMapping =
```
medb, line 1513: Can we get this map from options object using `getImpersonationServiceAccounts()` method instead?

medb:
May you add a description of feature(s) that this PR implements?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -1370,6 +1377,58 @@ private static long getMaxRemainingResults(
     return maxResults - numResults;
   }
 
+  /**
+   * TODO(hgong): Maybe a pre-filter can be done to filter out only the keys that match certain
+   * object name prefix pattern.
+   */
+  private String getMatchedObjectNamePrefix(Map<String, String> prefixes, String targetPrefix) {
```
medb, line 1384: Prefix-based auth seems a very different from user/group-auth in terms of implementation. May you implement it in separate follow up PR? It will be easier to reason about implementation this way.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: Resolved. Having separate property for each solved this problem

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: reverted the change, but kept the prefix configuration property name to make follow up change easier.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1508,14 +1508,22 @@ private Credential getCredential(Configuration config)
...
         GoogleHadoopFileSystemConfiguration.getGcsFsOptionsBuilder(config)
             .build()
             .getCloudStorageOptions();
+    if (isNullOrEmpty(serviceAccountToImpersonate)) {
+      serviceAccountToImpersonate = getMatchedServiceAccountToImpersonateFromUser(options);
+    }
+
+    if (isNullOrEmpty(serviceAccountToImpersonate)) {
+      serviceAccountToImpersonate = getMatchedServiceAccountToImpersonateFromGroup(options);
+    }
```
medb, line 1521: If you make this methods to return `Optional` then you can chain them:
```java
serviceAccountToImpersonate =
    getMatchedServiceAccountToImpersonateFromUser(options)
        .orElse(getMatchedServiceAccountToImpersonateFromGroup(options))
        .orElse(serviceAccountToImpersonate);
```

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1508,14 +1508,22 @@ private Credential getCredential(Configuration config)
             .withPrefixes(CONFIG_KEY_PREFIXES)
             .get(config, config::get);
```
medb, line 1509: You should move an above `IMPERSONATION_SERVICE_ACCOUNT_SUFFIX....get(...)` statement to be a last one that assigns to `serviceAccountToImpersonate` variable, because otherwise this generic impersonation configuration will override per-user/group configuration.

gcs/CHANGES.md
```diff
@@ -72,6 +72,20 @@
     if stream already closed. This fixes Spark Streaming jobs checkpointing to
     Cloud Storage.
 
+1.  Add properties to impersonate a service account through user or group name:
+
+    ```
+    fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
```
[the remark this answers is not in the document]
hongyegong: I guess this is not that bad.. given we will need to define another property for non-impersonation user/group/prefix auth anyway.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1531,6 +1539,30 @@ private Credential getCredential(Configuration config)
...
+      GoogleCloudStorageOptions options) throws IOException {
+    Map<String, String> serviceAccountMapping = options.getGroupImpersonationServiceAccounts();
+    List<String> groupNames = Arrays.asList(UserGroupInformation.getCurrentUser().getGroupNames());
+    for (Map.Entry<String, String> entry : serviceAccountMapping.entrySet()) {
+      if (groupNames.contains(entry.getKey())) {
+        return entry.getValue();
+      }
+    }
+    return null;
+  }
```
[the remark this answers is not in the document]
hongyegong: Nice, this is neat!

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: I believe the current chain will do the job.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: this cannot be chained like this unfortunately as orElse will return value type not another Optional. But it indeed can be recursively nested.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: Done.

util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialConfiguration.java
```diff
@@ -165,6 +167,24 @@
   public static final HadoopConfigurationProperty<String> IMPERSONATION_SERVICE_ACCOUNT_SUFFIX =
       new HadoopConfigurationProperty<>(".auth.impersonation.service.account");
 
+  /**
+   * Key prefix for the user identifier associated with the service account to impersonate when
+   * accessing GCS.
+   */
+  public static final HadoopConfigurationProperty<Map<String, String>>
+      USER_IMPERSONATION_IDENTIFIER_PREFIX =
```
medb, line 175: Rename to `IMPERSONATION_SERVICE_ACCOUNT_FOR_USER_SUFFIX` for consistency with IMPERSONATION_SERVICE_ACCOUNT_SUFFIX`?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1503,25 +1503,28 @@ private Credential getCredential(Configuration config)
...
                 getMatchedServiceAccountToImpersonateFromGroup(
                         options.getGroupImpersonationServiceAccounts(),
                         ImmutableList.copyOf(UserGroupInformation.getCurrentUser().getGroupNames()))
-                    .orElse(serviceAccountToImpersonate));
+                    .orElse(null));
+
+    String serviceAccountToImpersonate =
+        IMPERSONATION_SERVICE_ACCOUNT_SUFFIX
+            .withPrefixes(CONFIG_KEY_PREFIXES)
+            .get(config, config::get);
```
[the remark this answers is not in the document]
hongyegong: This is actually not matter of readability, it will change the logic to make this least priority. Current logic is to let the SA specified by impersonation.service.account supersede the ones specified in mappings.

gcs/CHANGES.md
```diff
@@ -75,8 +75,8 @@
 1.  Add properties to impersonate a service account through user or group name:
 
     ```
-    fs.gs.auth.user.impersonation.<USER_NAME> (not set by default)
-    fs.gs.auth.group.impersonation.<GROUP_NAME> (not set by default)
+    fs.gs.auth.impersonation.service.account.user.<USER_NAME> (not set by default)
```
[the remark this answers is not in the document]
hongyegong: It will look a little bit clumsy.. but I guess acceptable given that we have name like `fs.gs.max.wait.for.empty.object.creation.ms`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1503,19 +1503,33 @@ private Credential getCredential(Configuration config)
...
+      String serviceAccountFromUserOrGroup =
+          getMatchedServiceAccountToImpersonateFromGroup(
+                  options.getUserImpersonationServiceAccounts(),
+                  ImmutableList.of(UserGroupInformation.getCurrentUser().getShortUserName()))
+              .orElse(
+                  getMatchedServiceAccountToImpersonateFromGroup(
+                          options.getGroupImpersonationServiceAccounts(),
+                          ImmutableList.copyOf(
+                              UserGroupInformation.getCurrentUser().getGroupNames()))
+                      .orElse(null));
```
cyxxy, line 1525: Can we reformat so the indentations for user-match --> group-match --> null are matched.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -88,6 +88,10 @@
   /** Default setting for GCS HTTP request headers. */
   public static final ImmutableMap<String, String> HTTP_REQUEST_HEADERS_DEFAULT = ImmutableMap.of();
 
+  /** Default setting for impersonation service accounts. */
+  public static final ImmutableMap<String, String> IMPERSONATION_SERVICE_ACCOUNTS =
```
cyxxy, line 92: Please add a "DEFAULT" postfix to the variable name.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: Agreed it's not readable enough. Extracted those two unreadable method calls to two variables instead so logic will be more clear.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
hongyegong: Default will be an empty mapping and will be ignored at credential fetching time. If we actually put some default "string" then we need to actually differentiate the default string and actual user/group name which would rather be an empty check.

PR #402, opened 2020-06-21
Rewrite retry mechanism for gRPC write in gcsio

hongyegong:
This change mainly solves 2 issues:

1. After resumable write failed due to transient error, e.g. DEADLINE_EXCEED, committed size queried from server can be less than the size that has been uploaded thru client request. Current implementation doesn't handle this case. New implementation handles this by cached the past `numberOfRequestsToRetain` number of data chunk for later resume if needed.
2. This also fixes a bug of current retry mechanism that the bufferedInputStream could be reset to an invalid position

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -169,6 +169,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
     // Read end of the pipe.
     private final BufferedInputStream pipeSource;
     private final int MAX_BYTES_PER_MESSAGE = MAX_WRITE_CHUNK_BYTES.getNumber();
+    // A map mapping from write offset to retries that have been made to the corresponding offset.
     private final Map<Long, Integer> retryMap = new HashMap<>();
 
     UploadOperation(InputStream pipeSource) {
@@ -225,6 +226,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
       private final long writeOffset;
       private final String uploadId;
+      // A map holding cached data chunks, keyed by the offset to start writing each data chunk.
       private final TreeMap<Long, ByteString> dataChunkMap;
       private volatile boolean objectFinalized = false;
       // The last transient error to occur during the streaming RPC.
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -24,6 +24,7 @@ import com.google.cloud.hadoop.util.AsyncWriteChannelOptions;
 import com.google.cloud.hadoop.util.BaseAbstractGoogleAsyncWriteChannel;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
 import com.google.common.io.BaseEncoding;
@@ -43,6 +44,9 @@ import com.google.protobuf.Int64Value;
 import com.google.protobuf.UInt32Value;
 import com.google.protobuf.util.Timestamps;
 import io.grpc.Status;
+import io.grpc.Status.Code;
+import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
 import io.grpc.stub.ClientCallStreamObserver;
 import io.grpc.stub.ClientResponseObserver;
 import io.grpc.stub.StreamObserver;
@@ -51,6 +55,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
@@ -70,6 +76,14 @@ public final class GoogleCloudStorageGrpcWriteChannel
   // Maximum number of automatic retries for each data chunk
   // when writing to underlying channel raises error.
   private static final int UPLOAD_RETRIES = 10;
+
+  // Number of insert requests to retain, in case we need to rewind and resume an upload. Using too
+  // small of a number could risk being unable to resume the write if the resume point is an
+  // 'already-discarded buffer; and setting the value too high wastes RAM. Note: We could have a
+  // more complex implementation that periodically queries the service to find out the last
+  // committed offset, to determine what's safe to discard, but that would also impose a performance
+  // penalty.
+  private static int numberOfRequestsToRetain = 5;
 
   private final StorageStub stub;
   private final StorageResourceId resourceId;
@@ -179,7 +193,11 @@ public final class GoogleCloudStorageGrpcWriteChannel
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
       do {
-        responseObserver = new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher);
+        // Holds list of most recent number of numberOfRequestsToRetain requests, so upload can be
+        // rewound and re-sent upon non-transient errors.
+        List<InsertObjectRequest.Builder> requestList = new ArrayList<>();
+        responseObserver =
+            new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher, requestList);
         // TODO(b/151184800): Implement per-message timeout, in addition to stream timeout.
         stub.withDeadlineAfter(WRITE_STREAM_TIMEOUT.toMillis(), MILLISECONDS)
             .insertObject(responseObserver);
@@ -191,30 +209,14 @@ public final class GoogleCloudStorageGrpcWriteChannel
               String.format("Resumable upload failed for '%s'", getResourceString()), e);
         }
 
-        if (responseObserver.hasError()) {
-          long committedSize = getCommittedWriteSize(uploadId);
-          // If the last upload completely failed then reset to where it's marked before last
-          // insert. Otherwise, uploaded data chunk needs to be skipped before reading from buffered
-          // input stream.
-          pipeSource.reset();
-          if (committedSize > writeOffset) {
-            long uploadedDataChunkSize = committedSize - writeOffset;
-            pipeSource.skip(uploadedDataChunkSize);
-            writeOffset += uploadedDataChunkSize;
-          }
-          ++retriesAttempted;
+        if (responseObserver.hasTransientError()) {
+          writeOffset = getCommittedWriteSize(uploadId);
         } else {
           writeOffset += responseObserver.bytesWritten();
-          retriesAttempted = 0;
         }
+      } while (!responseObserver.hasFinalized());
 
-        if (retriesAttempted >= UPLOAD_RETRIES) {
-          throw new IOException(
-              String.format("Insert failed for '%s'", resourceId), responseObserver.getError());
-        }
-      } while (!responseObserver.hasFinalized() || responseObserver.hasError());
-
-      return responseObserver.getResponse();
+      return responseObserver.getResponseOrThrow();
     }
 
     /** Handler for responses from the Insert streaming RPC. */
@@ -223,46 +225,88 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
       private final long writeOffset;
       private final String uploadId;
+      private final List<InsertObjectRequest.Builder> requestList;
       private volatile boolean objectFinalized = false;
+      // The last transient error to occur during the streaming RPC.
+      private Throwable transientError = null;
+      // The last non-transient error to occur during the streaming RPC.
+      private Throwable nonTransientError = null;
       // The last error to occur during the streaming RPC. Present only on error.
       private IOException error;
       // The response from the server, populated at the end of a successful streaming RPC.
       private Object response;
       private ByteString chunkData;
       private Hasher objectHasher;
+      private boolean isReadyHandleExecuted = false;
 
       // CountDownLatch tracking completion of the streaming RPC. Set on error, or once the request
       // stream is closed.
       final CountDownLatch done = new CountDownLatch(1);
 
-      InsertChunkResponseObserver(String uploadId, long writeOffset, Hasher objectHasher) {
+      InsertChunkResponseObserver(
+          String uploadId,
+          long writeOffset,
+          Hasher objectHasher,
+          List<InsertObjectRequest.Builder> requestList) {
         this.uploadId = uploadId;
         this.chunkData = ByteString.EMPTY;
         this.writeOffset = writeOffset;
         this.objectHasher = objectHasher;
+        this.requestList = requestList;
       }
 
       @Override
-      public void beforeStart(ClientCallStreamObserver<InsertObjectRequest> requestObserver) {
+      public void beforeStart(final ClientCallStreamObserver<InsertObjectRequest> requestObserver) {
         requestObserver.setOnReadyHandler(
             new Runnable() {
               @Override
               public void run() {
-                if (objectFinalized) {
+                if (objectFinalized || isReadyHandleExecuted) {
                   // onReadyHandler may be called after we've closed the request half of the stream.
                   return;
                 }
 
+                isReadyHandleExecuted = true;
+
+                InsertObjectRequest.Builder requestBuilder = null;
+                // Create new request builder if the writeOffset grows over inserts. Otherwise
+                // something went wrong and resume write from a cached request builder.
+                if (requestList.size() > 0
+                    && Iterables.getLast(requestList).getWriteOffset() >= writeOffset) {
+                  requestBuilder = getSavedRequestBuilder(requestBuilder);
+                  if (requestBuilder == null) {
+                    return;
+                  }
+                } else {
+                  requestBuilder = buildRequestFromPipeData();
+                  if (requestBuilder == null) {
+                    return;
+                  }
+                }
+
+                requestObserver.onNext(requestBuilder.build());
+                if (objectFinalized) {
+                  // Close the request half of the streaming RPC.
+                  requestObserver.onCompleted();
+                }
+              }
+
+              // Handles the case where we're not resuming a write, and instead we need to read data
+              // from the input pipe.
+              private InsertObjectRequest.Builder buildRequestFromPipeData() {
+                InsertObjectRequest.Builder requestBuilder;
                 try {
                   chunkData = readRequestData();
                 } catch (IOException e) {
-                  error =
-                      new IOException(
-                          String.format("Failed to read chunk for '%s'", resourceId), e);
-                  return;
+                  nonTransientError =
+                      new RuntimeException(
+                          String.format(
+                              "InsertChunkResponseObserver.beforeStart for uploadId %s caught %s",
+                              uploadId, e));
+                  return null;
                 }
 
-                InsertObjectRequest.Builder requestBuilder =
+                requestBuilder =
                     InsertObjectRequest.newBuilder()
                         .setUploadId(uploadId)
                         .setWriteOffset(writeOffset);
@@ -294,49 +338,86 @@ public final class GoogleCloudStorageGrpcWriteChannel
                                 UInt32Value.newBuilder().setValue(objectHasher.hash().asInt())));
                   }
                 }
-                requestObserver.onNext(requestBuilder.build());
-
-                if (objectFinalized) {
-                  // Close the request half of the streaming RPC.
-                  requestObserver.onCompleted();
+                if (requestList.size() == numberOfRequestsToRetain) {
+                  requestList.remove(0);
                 }
+                requestList.add(requestBuilder);
+                return requestBuilder;
               }
 
               private ByteString readRequestData() throws IOException {
-                // Mark the input stream in case this request fails so that read can be recovered
-                // from where it's marked.
-                pipeSource.mark(MAX_BYTES_PER_MESSAGE);
                 ByteString data =
                     ByteString.readFrom(ByteStreams.limit(pipeSource, MAX_BYTES_PER_MESSAGE));
 
                 objectFinalized =
-                    data.size() < MAX_BYTES_PER_MESSAGE || pipeSource.available() <= 0;
+                    data.size() < MAX_BYTES_PER_MESSAGE;
                 return data;
               }
             });
       }
 
-      public Object getResponse() throws IOException {
-        if (hasError()) {
-          throw getError();
+      // Handles the case when a writeOffset of data read previously is being processed. This
+      // happens if a transient failure happens while uploading, and can be resumed by querying the
+      // current committed offset.
+      private InsertObjectRequest.Builder getSavedRequestBuilder(
+          InsertObjectRequest.Builder requestBuilder) {
+        // Resume will only work if the first request builder in the cache carries a offset not
+        // greater than the current writeOffset.
+        if (requestList.get(0).getWriteOffset() <= writeOffset) {
+          for (int i = 0; i < requestList.size(); i++) {
+            // Take current request builder if the writerOffset of it is equal to the current
+            // writeOffset. Take the previous one if it's greater than current writeOffset. This
+            // will work as server will allow to re-upload a previously uploaded part of the object
+            // and ignore the bytes preceding the last committed offset.
+            InsertObjectRequest.Builder currentRequestBuilder = requestList.get(i);
+            if (currentRequestBuilder.getWriteOffset() == writeOffset) {
+              requestBuilder = currentRequestBuilder;
+              break;
+            } else if (currentRequestBuilder.getWriteOffset() > writeOffset && i > 0) {
+              requestBuilder = requestList.get(i - 1);
+              break;
+            }
+          }
+        }
+        if (requestBuilder == null) {
+          nonTransientError =
+              new RuntimeException(
+                  String.format(
+                      "Didn't have enough data buffered for attempt to resume upload for uploadID"
+                          + " %s: last committed offset=%s, earliest buffered offset=%s.",
+                      uploadId, writeOffset, requestList.get(0).getWriteOffset()));
+          return null;
+        }
+        return requestBuilder;
+      }
+
+      public Object getResponseOrThrow() throws IOException {
+        if (hasNonTransientError()) {
+          throw new IOException(
+              String.format("Resumable upload failed for '%s'", getResourceString()),
+              nonTransientError);
         }
         return checkNotNull(response, "Response not present for '%s'", resourceId);
       }
 
-      boolean hasError() {
-        return error != null || response == null;
+      boolean hasTransientError() {
+        return transientError != null || response == null;
+      }
+
+      boolean hasNonTransientError() {
+        return nonTransientError != null;
+      }
+
+      Throwable getNonTransientError() {
+        return nonTransientError;
       }
 
       int bytesWritten() {
         return chunkData.size();
       }
 
-      public IOException getError() {
-        return checkNotNull(error, "Error not present for '%s'", resourceId);
-      }
-
       boolean hasFinalized() {
-        return objectFinalized;
+        return objectFinalized || hasNonTransientError();
       }
 
       @Override
@@ -348,13 +429,28 @@ public final class GoogleCloudStorageGrpcWriteChannel
       public void onError(Throwable t) {
         Status s = Status.fromThrowable(t);
         String statusDesc = s == null ? "" : s.getDescription();
-        error =
-            new IOException(
-                String.format(
-                    "Caught exception for '%s', while uploading to uploadId %s at writeOffset %d."
-                        + " Status: %s",
-                    resourceId, uploadId, writeOffset, statusDesc),
-                t);
+
+        if (t.getClass() == StatusException.class || t.getClass() == StatusRuntimeException.class) {
+          Code code =
+              t.getClass() == StatusException.class
+                  ? ((StatusException) t).getStatus().getCode()
+                  : ((StatusRuntimeException) t).getStatus().getCode();
+          if (code == Code.DEADLINE_EXCEEDED
+              || code == Code.RESOURCE_EXHAUSTED
+              || code == Code.INTERNAL
+              || code == Code.UNAVAILABLE) {
+            transientError = t;
+          }
+        }
+        if (transientError == null) {
+          nonTransientError =
+              new Throwable(
+                  String.format(
+                      "Caught exception for '%s', while uploading to uploadId %s at writeOffset %d."
+                          + " Status: %s",
+                      resourceId, uploadId, writeOffset, statusDesc),
+                  t);
+        }
         done.countDown();
       }
 
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -24,7 +24,6 @@ import com.google.cloud.hadoop.util.AsyncWriteChannelOptions;
 import com.google.cloud.hadoop.util.BaseAbstractGoogleAsyncWriteChannel;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
 import com.google.common.io.BaseEncoding;
@@ -55,9 +54,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -179,7 +193,11 @@ private Object doResumableUpload() throws IOException {
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
       do {
-        responseObserver = new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher);
+        // Holds list of most recent number of numberOfRequestsToRetain requests, so upload can be
+        // rewound and re-sent upon non-transient errors.
+        List<InsertObjectRequest.Builder> requestList = new ArrayList<>();
```
mfschwartz: I realize this came from my original implementation but: ArrayList.remove is inefficient - requires copying elements. How about switching to a LinkedList?

hongyegong: There are two commits on this PR. You must be looking at first commit not all commits. I actually figured that TreeMap will be a good fit here and updated the code accordingly.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +338,86 @@ public void run() {
...
+          for (int i = 0; i < requestList.size(); i++) {
+            // Take current request builder if the writerOffset of it is equal to the current
+            // writeOffset. Take the previous one if it's greater than current writeOffset. This
+            // will work as server will allow to re-upload a previously uploaded part of the object
+            // and ignore the bytes preceding the last committed offset.
+            InsertObjectRequest.Builder currentRequestBuilder = requestList.get(i);
+            if (currentRequestBuilder.getWriteOffset() == writeOffset) {
+              requestBuilder = currentRequestBuilder;
+              break;
+            } else if (currentRequestBuilder.getWriteOffset() > writeOffset && i > 0) {
```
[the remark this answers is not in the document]
hongyegong: Please checkout the latest commit, the logic here changed quite a lot.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +338,86 @@ public void run() {
...
+      private InsertObjectRequest.Builder getSavedRequestBuilder(
+          InsertObjectRequest.Builder requestBuilder) {
+        // Resume will only work if the first request builder in the cache carries a offset not
+        // greater than the current writeOffset.
+        if (requestList.get(0).getWriteOffset() <= writeOffset) {
+          for (int i = 0; i < requestList.size(); i++) {
+            // Take current request builder if the writerOffset of it is equal to the current
+            // writeOffset. Take the previous one if it's greater than current writeOffset. This
+            // will work as server will allow to re-upload a previously uploaded part of the object
+            // and ignore the bytes preceding the last committed offset.
```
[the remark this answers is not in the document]
hongyegong: Not sure it's feasible to trigger this particular condition during retry in integration env. I'll try to consolidate this in grpc doc/code.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +338,86 @@ public void run() {
...
+              break;
+            }
+          }
+        }
+        if (requestBuilder == null) {
+          nonTransientError =
+              new RuntimeException(
+                  String.format(
+                      "Didn't have enough data buffered for attempt to resume upload for uploadID"
+                          + " %s: last committed offset=%s, earliest buffered offset=%s.",
```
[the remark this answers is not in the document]
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +338,86 @@ public void run() {
...
 
-      public Object getResponse() throws IOException {
-        if (hasError()) {
-          throw getError();
+      // Handles the case when a writeOffset of data read previously is being processed. This
+      // happens if a transient failure happens while uploading, and can be resumed by querying the
+      // current committed offset.
+      private InsertObjectRequest.Builder getSavedRequestBuilder(
+          InsertObjectRequest.Builder requestBuilder) {
+        // Resume will only work if the first request builder in the cache carries a offset not
```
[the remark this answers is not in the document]
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +338,86 @@ public void run() {
...
             });
       }
 
-      public Object getResponse() throws IOException {
-        if (hasError()) {
-          throw getError();
+      // Handles the case when a writeOffset of data read previously is being processed. This
+      // happens if a transient failure happens while uploading, and can be resumed by querying the
+      // current committed offset.
+      private InsertObjectRequest.Builder getSavedRequestBuilder(
```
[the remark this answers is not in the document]
hongyegong: I actually already updated this method name to cachedRequest instead of savedRequest. Again PTAL all commits.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -223,46 +225,88 @@ private Object doResumableUpload() throws IOException {
 
       private final long writeOffset;
       private final String uploadId;
+      private final List<InsertObjectRequest.Builder> requestList;
```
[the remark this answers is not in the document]
hongyegong: N/A

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -223,46 +225,88 @@ private Object doResumableUpload() throws IOException {
...
+      private Throwable transientError = null;
+      // The last non-transient error to occur during the streaming RPC.
+      private Throwable nonTransientError = null;
       // The last error to occur during the streaming RPC. Present only on error.
       private IOException error;
       // The response from the server, populated at the end of a successful streaming RPC.
       private Object response;
       private ByteString chunkData;
       private Hasher objectHasher;
+      private boolean isReadyHandleExecuted = false;
```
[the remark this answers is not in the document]
hongyegong: I'll just delete it, this was a line of code I added for a broken test.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -179,7 +193,11 @@ private Object doResumableUpload() throws IOException {
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
       do {
-        responseObserver = new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher);
+        // Holds list of most recent number of numberOfRequestsToRetain requests, so upload can be
+        // rewound and re-sent upon non-transient errors.
+        List<InsertObjectRequest.Builder> requestList = new ArrayList<>();
+        responseObserver =
+            new InsertChunkResponseObserver(uploadId, writeOffset, objectHasher, requestList);
         // TODO(b/151184800): Implement per-message timeout, in addition to stream timeout.
```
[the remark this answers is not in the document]
hongyegong: This is actually quite often seen in GCS connector code, usually go by `TODO(b/xxxxxxxx)`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -71,6 +77,14 @@
   // when writing to underlying channel raises error.
   private static final int UPLOAD_RETRIES = 10;
 
+  // Number of insert requests to retain, in case we need to rewind and resume an upload. Using too
+  // small of a number could risk being unable to resume the write if the resume point is an
+  // 'already-discarded buffer; and setting the value too high wastes RAM. Note: We could have a
```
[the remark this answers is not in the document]
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -223,7 +225,12 @@ private Object doResumableUpload() throws IOException {
 
       private final long writeOffset;
       private final String uploadId;
+      private final TreeMap<Long, ByteString> dataChunkMap;
       private volatile boolean objectFinalized = false;
```
mfschwartz, line 229: Why do we need volatile? Only a single thread operates in this class at a time, right?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -223,7 +225,12 @@ private Object doResumableUpload() throws IOException {
 
       private final long writeOffset;
       private final String uploadId;
+      private final TreeMap<Long, ByteString> dataChunkMap;
```
mfschwartz, line 228: Add a comment, "Maps from x to y ..." (i.e., what is the Long, what is the ByteString)?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +328,145 @@ public void run() {
...
+                                    uploadId, curWriteOffset));
+                      } else {
+                        retryMap.put(curWriteOffset, count + 1);
+                      }
+                      break;
+                    } else if (entry.getKey() > writeOffset) {
+                      chunkData = prevDataChunk;
+                      requestBuilder =
+                          buildRequestFromWriteOffsetAndDataChunk(prevWriteOffset, prevDataChunk);
+                      int count =
```
mfschwartz, line 382: How about moving this check above the loop - it's loop-invariant, so doing that would make the loop shorter/more readable.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +328,145 @@ public void run() {
...
                   }
                 }
-                requestObserver.onNext(requestBuilder.build());
-
-                if (objectFinalized) {
-                  // Close the request half of the streaming RPC.
-                  requestObserver.onCompleted();
+                if (dataChunkMap.size() == numberOfRequestsToRetain) {
+                  dataChunkMap.remove(dataChunkMap.firstKey());
+                  retryMap.remove(dataChunkMap.firstKey());
```
mfschwartz, line 333: Why do we need two different maps from offset to chunk?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +328,145 @@ public void run() {
...
+                  TreeMap<Long, ByteString> dataChunkMap) {
+                // Resume will only work if the first request builder in the cache carries an offset
+                // not greater than the current writeOffset.
+                InsertObjectRequest.Builder requestBuilder = null;
+                if (dataChunkMap.size() > 0 && dataChunkMap.firstKey() <= writeOffset) {
+                  long prevWriteOffset = 0;
+                  ByteString prevDataChunk = null;
+                  for (Map.Entry<Long, ByteString> entry : dataChunkMap.entrySet()) {
+                    // Take the previous entry if it's greater than current writeOffset.
+                    if (entry.getKey() == writeOffset) {
```
mfschwartz, line 361: Why do we need to handle the case where entry.getKey() == writeOffset separately from the case where entry.getKey() > writeOffset? I think you just need to iterate until you get to the first entry whose offset is <= last persisted offset + write size (which means the cached request contains the offset where we want to resume writing). I think the code would be more readable if you have this one condition instead of two.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -178,8 +192,12 @@ private Object doResumableUpload() throws IOException {
       long writeOffset = 0;
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
+      // Holds list of most recent number of numberOfRequestsToRetain requests, so upload can be
+      // rewound and re-sent upon non-transient errors.
+      TreeMap<Long, ByteString> dataChunkMap = new TreeMap<>();
```
mfschwartz, line 197: I assume the reason you chose a TreeMap is because you can then key on the offset. But since you always just iterate over it (rather than do a single lookup to find the needed chunk), I think this is overkill and kind of confusing; all you really need is a LinkedList, so you know you will iterate in offset order.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: The count is actually related to iteration itself. We have no idea upfront which <offset, datachunk> will be retired and which are not until we iterate and find the first entry that meed resume requirement.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: I am not sure I understand correctly "whose offset is <= last persisted offset + write size". I think regardless we need to diff btw '<' and '=='. So just to be clear, each entry in the map means <write offset, datachunk that will be written from this write offset>. So ideally the new writeOffset that comes with a new insert would be larger than all keys in the caching map. If not, there are two cases 1) new writeOffset is even smaller than the first key of the caching map, in which case we will just consider as non-resumable and as a in-transient error 2) this is the case we can resume from the caching map, if the new writeOffset is equal to any of previous key that means we can redo insertion with exactly that offset and the corresponding data chunk. However if there is no exact match, we want to find the first key(offset) in the map that is greater that this new offset but we do not use this one to redo insert because there will be piece of last data chunk that's gonna be missing(remember the current new writeOffset means committed size we queried from server). Instead we want to use the entry <offset, datachunk> one before the offset that's larger than new writeOffset. Hope this helps to explain...

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Second is keep track of how many tries have been made to each offset position. Otherwise the logic could end up trying a failed offset with associated data chunk endlessly.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Docstring added.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: But we need to keep both value of offset, not just order, along with cached data chunk. Just the fact that data chunks are pushed in increasing offset order will not be enough to pick the right data chunk to resume write.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Not necessarily, for example the code in beforestart() will be running in a different thread and it could be we want the status of write is reflected in main memory once it's been marked as true in any thread.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: This approach will potentially make us retry too many times (in the worst case we'll retry UPLOAD_RETRIES per chunk) and that the number of retries varies with object size being uploaded.

Let's please keep this simpler. Impose the UPLOAD_RETRIES limit at the top level, i.e., make that the maximum number of times we retry for a transient error. That will be easier to understand and consistent with the way retry handling is implemented in all other GCS clients. And it will let you get rid of this second map, making the code easier to read and reason about.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: Oh, I see, your map no longer stores the pre-built request objects, but rather just the bytes (which is why you then also need to store the offset). In my implementation I held on to the request objects. The way you are doing it forces you to re-scan the data every time you re-transmit (to compute the checksums).

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -64,12 +69,20 @@
   // Default GCS upload granularity.
   static final int GCS_MINIMUM_CHUNK_SIZE = 256 * 1024;
 
-  private static final Duration START_RESUMABLE_WRITE_TIMEOUT = Duration.ofMinutes(10);
-  private static final Duration QUERY_WRITE_STATUS_TIMEOUT = Duration.ofMinutes(10);
-  private static final Duration WRITE_STREAM_TIMEOUT = Duration.ofMinutes(20);
+  private static final Duration START_RESUMABLE_WRITE_TIMEOUT = Duration.ofSeconds(10);
+  private static final Duration QUERY_WRITE_STATUS_TIMEOUT = Duration.ofSeconds(10);
```
medb, line 73: Isn't this too low? Can we make is larger just to make sure that we are resilient to network issues? Maybe at least 1 minute.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -156,6 +169,8 @@ public void startUpload(InputStream pipeSource) {
     // Read end of the pipe.
     private final BufferedInputStream pipeSource;
     private final int MAX_BYTES_PER_MESSAGE = MAX_WRITE_CHUNK_BYTES.getNumber();
+    // A map mapping from write offset to retries that have been made to the corresponding offset.
+    private final Map<Long, Integer> retryMap = new HashMap<>();
```
medb, line 173: Rename to `offsetRetryMap` and use `AtomicLong` as a value because you can increment it without changing value in the map.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -223,7 +226,13 @@ private Object doResumableUpload() throws IOException {
 
       private final long writeOffset;
       private final String uploadId;
+      // A map holding cached data chunks, keyed by the offset to start writing each data chunk.
+      private final TreeMap<Long, ByteString> dataChunkMap;
```
medb, line 230: Why do we need 2 `dataChunkMap`s?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -253,13 +267,35 @@ public void run() {
...
                 }
 
+                InsertObjectRequest.Builder requestBuilder = null;
+                // Create new request builder if the writeOffset grows over inserts. Otherwise
+                // something went wrong and resume write from a cached request builder.
+                if (dataChunkMap.size() > 0 && dataChunkMap.lastKey() >= writeOffset) {
+                  requestBuilder = buildRequestFromCachedDataChunk(dataChunkMap);
+                } else {
+                  requestBuilder = buildRequestFromPipeData();
+                }
```
medb, line 277: Convert to ternary operator.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -253,13 +267,35 @@ public void run() {
...
                 } catch (IOException e) {
-                  error =
-                      new IOException(
-                          String.format("Failed to read chunk for '%s'", resourceId), e);
-                  return;
+                  nonTransientError =
+                      new RuntimeException(
+                          String.format(
+                              "InsertChunkResponseObserver.beforeStart for uploadId %s caught %s",
+                              uploadId, e));
```
medb, line 297: Pass `e` as a parameter of the `RuntimeException` not `format` method, so its stack trace will be preserved.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +330,145 @@ public void run() {
...
               private ByteString readRequestData() throws IOException {
-                // Mark the input stream in case this request fails so that read can be recovered
-                // from where it's marked.
-                pipeSource.mark(MAX_BYTES_PER_MESSAGE);
                 ByteString data =
                     ByteString.readFrom(ByteStreams.limit(pipeSource, MAX_BYTES_PER_MESSAGE));
-
-                objectFinalized =
-                    data.size() < MAX_BYTES_PER_MESSAGE || pipeSource.available() <= 0;
+                pipeSource.mark(1);
```
medb, line 344: Why do we mark with `1`?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -178,8 +193,12 @@ private Object doResumableUpload() throws IOException {
       long writeOffset = 0;
       int retriesAttempted = 0;
       Hasher objectHasher = Hashing.crc32c().newHasher();
+      // Holds list of most recent number of numberOfRequestsToRetain requests, so upload can be
+      // rewound and re-sent upon non-transient errors.
+      TreeMap<Long, ByteString> dataChunkMap = new TreeMap<>();
```
medb, line 198: Are there any guarantees that `getCommittedWriteSize` will always return one of the previously sent offsets? I.e. is it always aligned to 2 MiB (chunk size)?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -294,49 +330,145 @@ public void run() {
...
+              // querying the current committed offset.
+              private InsertObjectRequest.Builder buildRequestFromCachedDataChunk(
+                  TreeMap<Long, ByteString> dataChunkMap) {
+                // Resume will only work if the first request builder in the cache carries an offset
+                // not greater than the current writeOffset.
+                InsertObjectRequest.Builder requestBuilder = null;
+                if (dataChunkMap.size() > 0 && dataChunkMap.firstKey() <= writeOffset) {
+                  long prevWriteOffset = 0;
+                  ByteString prevDataChunk = null;
+                  for (Map.Entry<Long, ByteString> entry : dataChunkMap.entrySet()) {
```
medb, line 361: Because you already use TreeSet, may you use binary search methods to find offset that you need instead of iterating over them?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -348,13 +480,28 @@ public void onNext(Object response) {
...
-                        + " Status: %s",
-                    resourceId, uploadId, writeOffset, statusDesc),
-                t);
+
+        if (t.getClass() == StatusException.class || t.getClass() == StatusRuntimeException.class) {
+          Code code =
+              t.getClass() == StatusException.class
+                  ? ((StatusException) t).getStatus().getCode()
+                  : ((StatusRuntimeException) t).getStatus().getCode();
+          if (code == Code.DEADLINE_EXCEEDED
```
medb, line 489: Create a `HashSet` of transient error codes and check if `code` is in this set instead of using many `||` statements.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -348,13 +480,28 @@ public void onNext(Object response) {
...
+              || code == Code.RESOURCE_EXHAUSTED
+              || code == Code.INTERNAL
+              || code == Code.UNAVAILABLE) {
+            transientError = t;
+          }
+        }
+        if (transientError == null) {
+          nonTransientError =
+              new Throwable(
```
medb, line 498: Set `t` as a cause of this `Throwable` and change exception type of all thrown exceptions to `IOException` - I think this is more appropriate.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -64,12 +69,20 @@
...
-  private static final int UPLOAD_RETRIES = 10;
+  private static final int UPLOAD_RETRIES = 5;
+
+  // Number of insert requests to retain, in case we need to rewind and resume an upload. Using too
+  // small of a number could risk being unable to resume the write if the resume point is an
+  // already-discarded buffer; and setting the value too high wastes RAM. Note: We could have a
+  // more complex implementation that periodically queries the service to find out the last
+  // committed offset, to determine what's safe to discard, but that would also impose a performance
+  // penalty.
+  private static int numberOfRequestsToRetain = 5;
```
medb, line 85: Convert name to capitalized snake case, because this is a constant.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: +1, I started a previous conversation about this.
I suggest we set up a meeting because we seem to be revisiting the same questions.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: The list is only 5 long, so binary search is less efficient than just walking the list linearly.
I also previously suggested a TreeSet is overkill - all we need here is a simple LinkedList. Henry didn't want to do that because he wants to get the offset from the tree key but if you instead use as LinkedList of request messages (rather than just the byte strings) you have the offsets in the request messages and it has the other important advantage that you avoid re-scanning the full chunk data whenever re-transmitting.

Again, I think it's time for a meeting. I feel the current implementation is both overly complex and inefficient.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Updated. Although this is supposely not kind of requests that should take a minute before triggering retry.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Removed this map for now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Discussed offline, removed the retryMap for now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: This is to revert to old approach to peek available bytes to avoid making a new request for nothing.
It was removed to avoid confusion because we relied on another reset to resume write, since we don't do that anymore hence revert to old way.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: That's not assumed in the implementation, gRPC write updates with offsets like 2, 4, 6, 8Mibs and the size returned from server can be 7Mib, in that case we resume uploading from offset 6Mibs again and Michael confirmed that GCS is smart enough to start with 7 and drop the data from 6-7.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Discussed offline. Will keep TreeSet and plain iteration for the sake of simplicity and readability.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Created a static Set for these codes for maintainability.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: As discussed offline, this map can be removed and will enforce the maximum retries of UPLOAD_RETRIES at top lvl.

hongyegong:
> This PR reduces test coverage by 0.1%, please add test to cover untested functionality.

Added more test to cover more edge cases. Current unit test design is quite inflexible to add some test cases and needs a refactor.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mfschwartz: It shouldn't take a minute, but 10 seconds was kind of short. You could use 30 seconds or just leave it at 1 min.

PR #408, opened 2020-06-29
Fix per-user/group SA impersonation

medb:
Changes:

1. Do not override per-user/group impersonation configuration with generic impersonation.
1. Code cleanup

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -30,7 +30,9 @@ import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration
 import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.GCS_WORKING_DIRECTORY;
 import static com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemConfiguration.PERMISSIONS_TO_REPORT;
 import static com.google.cloud.hadoop.gcsio.CreateFileOptions.DEFAULT_NO_OVERWRITE;
+import static com.google.cloud.hadoop.util.HadoopCredentialConfiguration.GROUP_IMPERSONATION_SERVICE_ACCOUNT_SUFFIX;
 import static com.google.cloud.hadoop.util.HadoopCredentialConfiguration.IMPERSONATION_SERVICE_ACCOUNT_SUFFIX;
+import static com.google.cloud.hadoop.util.HadoopCredentialConfiguration.USER_IMPERSONATION_SERVICE_ACCOUNT_SUFFIX;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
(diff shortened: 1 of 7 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1494,65 +1498,72 @@ private Credential getCredential(Configuration config)
...
-              options.getUserImpersonationServiceAccounts(),
-              ImmutableList.of(UserGroupInformation.getCurrentUser().getShortUserName()));
-      Optional<String> serviceAccountToImpersonateFromGroup =
-          getMatchedServiceAccountToImpersonateFromGroup(
-              options.getGroupImpersonationServiceAccounts(),
-              ImmutableList.copyOf(UserGroupInformation.getCurrentUser().getGroupNames()));
-      serviceAccountToImpersonate =
-          serviceAccountToImpersonateFromUser.orElse(
-              serviceAccountToImpersonateFromGroup.orElse(null));
+      Configuration config, GoogleCloudStorageFileSystemOptions gcsFsOptions, Credential credential)
```
hongyegong, line 1509: I thought designated service account will always take precedence?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1494,65 +1498,72 @@ private Credential getCredential(Configuration config)
...
+                    () ->
+                        Optional.ofNullable(
+                            IMPERSONATION_SERVICE_ACCOUNT_SUFFIX
+                                .withPrefixes(CONFIG_KEY_PREFIXES)
+                                .get(config, config::get)))
+            .map(Supplier::get)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .filter(sa -> !isNullOrEmpty(sa))
+            .findFirst();
```
hongyegong, line 1537: Not super familiar with lambda in java, but will this always return some form of Optional object even if any filter above returns nothing?

medb: Yes, it will return `Optinal.empty()` if nothing is found.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
medb: I think that it's the best if the narrower scope impersonation configuration takes precedence over broader-scope impersonation. I.e user impersonation overrides group impersonation which in turn overrides default/generic impersonation.

Is this what you thought too?

hongyegong: Agreed with narrower scope should take precedence. The original design was if a single SA is set up thru IMPERSONATION_SERVICE_ACCOUNT_SUFFIX this will take precedence over everything(I somehow have such impression that you also agreed on this in one of the comments you had on my PR).

But I guess this is not a critical order anyway, since I would doubt it happens often that properties of both two categories will be set at the same time. It's helpful to document the precedence explicitly tho.

hongyegong: If we keep like this way maybe add something like "Note that if any of these two properties is set, it will override the value of
`fs.gs.auth.impersonation.service.account` if set." down below the Service account impersonation section.

medb: Done.

PR #411, opened 2020-07-09
Use System Lambda's withEnvironmentVariable

stefanbirkner:
System Lambda is independent of the test framework and therefore no
obstacle for changing the test framework.
System Lambda is more local, too. It is only used for the single
statement that needs environment variables.

### DIFF
--- a/util-hadoop/src/test/java/com/google/cloud/hadoop/util/HadoopCredentialConfigurationTest.java
+++ b/util-hadoop/src/test/java/com/google/cloud/hadoop/util/HadoopCredentialConfigurationTest.java
@@ -159,7 +159,7 @@ public class HadoopCredentialConfigurationTest {
 
     GoogleCredentialWithRetry credential =
         (GoogleCredentialWithRetry)
-        withEnvironmentVariable(CREDENTIAL_ENV_VAR, getStringPath("test-credential.json"))
+            withEnvironmentVariable(CREDENTIAL_ENV_VAR, getStringPath("test-credential.json"))
                 .execute(() -> credentialFactory.getCredential(TEST_SCOPES));
 
     assertThat(credential.getServiceAccountId()).isEqualTo("test-email@gserviceaccount.com");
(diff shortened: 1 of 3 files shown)


util-hadoop/src/test/java/com/google/cloud/hadoop/util/HadoopCredentialConfigurationTest.java
```diff
@@ -158,15 +155,15 @@ public void metadataServiceIsUsedByDefault() throws Exception {
...
-
     CredentialFactory credentialFactory = getCredentialFactory();
-
-    GoogleCredentialWithRetry credential =
-        (GoogleCredentialWithRetry) credentialFactory.getCredential(TEST_SCOPES);
-
-    assertThat(credential.getServiceAccountId()).isEqualTo("test-email@gserviceaccount.com");
-    assertThat(credential.getServiceAccountPrivateKeyId()).isEqualTo("test-key-id");
+    withEnvironmentVariable(CREDENTIAL_ENV_VAR, getStringPath("test-credential.json"))
+        .execute(() -> {
```
medb, line 160: Thank you for your contribution!

One suggestion for future feature - It will be good if `execute` or other method returns result so it can be consumed and/or asserted outside of the lambda - similar pattern to the JUnit `assertThrows` method:
```java
Exception e = assertThrows(Exception.class, () -> ...)
```

Having this functionality will allow to make tests more readable, because it will be obvious what exact commands need env variables to be set.

stefanbirkner:
@medb I fixed the conflict by rebasing against master.

util-hadoop/src/test/java/com/google/cloud/hadoop/util/HadoopCredentialConfigurationTest.java
medb: @stefanbirkner FYI

medb: Thank you! LGTM

stefanbirkner:
Thanks for fixing the formatting.

PR #412, opened 2020-07-10
adding quotaExceeded method in ApiErrorExtractor

danielxjd:
adding an additional method in ApiErrorExtractor to help determine if an IOException is caused by quotaExceeded. This additional method would help a lot for the development of Apache Beam.

### DIFF
--- a/util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
@@ -47,6 +47,8 @@ public class ApiErrorExtractor {
   public static final String RATE_LIMITED_REASON = "rateLimitExceeded";
   public static final String USER_RATE_LIMITED_REASON = "userRateLimitExceeded";
 
+  public static final String Quota_Exceeded_REASON = "quotaExceeded";
+
   // These come with "The account for ... has been disabled" message.
   public static final String ACCOUNT_DISABLED_REASON = "accountDisabled";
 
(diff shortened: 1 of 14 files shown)


util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
```diff
@@ -205,6 +207,18 @@ public boolean rateLimited(IOException e) {
     return false;
   }
 
+  /**
+   * Determines if a given Throwable is caused by Quota Exceeded. Recursively checks
+   * getCause() if outer exception isn't an instance of the correct class.
+   */
+  public boolean quotaExceeded(IOException e) {
+    ErrorInfo errorInfo = getErrorInfo(e);
+    if(errorInfo !=null){
```
medb, line 216: Please, format this change with google java formatter: https://github.com/google/google-java-format

danielxjd: Reformatted with the IntelliJ google-java-format plugin

medb: Please format only changes lines, not whole file: https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py

danielxjd: Sorry for that, I have created a new PR for the change, will close this one once that get passed.

PR #413, opened 2020-07-13
Add quotaExceeded method to ApiErrorExtractor

danielxjd:


util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
```diff
@@ -205,6 +207,18 @@ public boolean rateLimited(IOException e) {
...
+  /**
+   * Determines if a given Throwable is caused by Quota Exceeded. Recursively checks getCause() if
+   * outer exception isn't an instance of the correct class.
+   */
+  public boolean quotaExceeded(IOException e) {
+    ErrorInfo errorInfo = getErrorInfo(e);
+    if (errorInfo != null) {
+      return QUOTA_EXCEEDED_REASON.equals(errorInfo.getReason());
+    }
+    return false;
```
medb, line 219: Replace with one-liner:

```java
return errorInfo != null && QUOTA_EXCEEDED_REASON.equals(errorInfo.getReason());
```

util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
```diff
@@ -205,6 +207,18 @@ public boolean rateLimited(IOException e) {
     return false;
   }
 
+  /**
+   * Determines if a given Throwable is caused by Quota Exceeded. Recursively checks getCause() if
+   * outer exception isn't an instance of the correct class.
+   */
+  public boolean quotaExceeded(IOException e) {
```
medb, line 214: What status code GCS returns for requests with `quotaExceeded` reason? Is it `429`?

danielxjd: I think the error code is 403, same as the rateLimitedExceeded

medb: `rateLimitedExceeded` has `429` code, I see that other methods in this class also check status code, so we may want to do the same for this new method too.

danielxjd: ok, I will add a check for status code on this method

danielxjd: I am not sure about whether the getReason is enough to determine the error type, since there is not a check for code for rateLimitedExceeded error.

danielxjd: And also from [Google Cloud Error Messages](https://cloud.google.com/bigquery/docs/error-messages) the http code for both rateLimited and quotaExceeded error has the error code of 403.

danielxjd: And from [Google Cloud Error Message](https://cloud.google.com/bigquery/docs/error-messages), both error have error code of 403.

medb: okay, let's leave it as is, and we will update all methods to check status code in the future.

danielxjd:
I am not familiar with the codecov/patch test. Is there any suggestion on what should I do to pass the test?

util/src/main/java/com/google/cloud/hadoop/util/ApiErrorExtractor.java
```diff
@@ -205,6 +207,15 @@ public boolean rateLimited(IOException e) {
...
```
medb, line 214: Please, add a test for this method in [ApiErrorExtractorTest](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/util/src/test/java/com/google/cloud/hadoop/util/ApiErrorExtractorTest.java).

medb:
Thank you for your contribution!

PR #417, opened 2020-07-17
Change exception message in ListFileInfo.

aalexx-S:
Let the exception message in ListFileInfo includes underlying
exception messages.


medb:
May you add an example of exception message that is not clear enough without this change?

aalexx-S:
For example, if I list at root directory and it fails:
```
$ hadoop fs -ls gs://bucket/
ls: Failed to listFileInfo for 'gs://bucket'
```
The error message doesn't expose the cause of the error.

PR #418, opened 2020-07-21
Do not set IfGenerationMatch parameter for creation with overwrite

hongyegong:


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
```diff
@@ -364,6 +368,43 @@ public void testOverwrite()
     // Try to create the same file again with overwrite == false.
     assertThrows(
         IOException.class, () -> ghfsHelper.writeFile(hadoopPath, text, 1, /* overwrite= */ false));
+
+    // Try to create the same file again with overwrite == true.
+    ghfsHelper.writeFile(hadoopPath, text, 1, /* overwrite= */ true);
```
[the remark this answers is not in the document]
hongyegong: It's calling public API essentially. Let's maybe keep this for now just for test consistency, we use this helper method to call underlying Create(...) elsewhere.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -360,7 +360,7 @@ public WritableByteChannel create(final StorageResourceId resourceId, CreateObje
     Optional<Long> overwriteGeneration =
         resourceId.hasGenerationId()
             ? Optional.of(resourceId.getGenerationId())
-            : Optional.of(getWriteGeneration(resourceId, options.overwriteExisting()));
+            : getWriteGeneration(resourceId, options.overwriteExisting());
```
medb, line 363: Seems like we don't need to check existence of the object here and just depend on `options.overwriteExisting()` condition:
```java
options.overwriteExisting() ? Optional.absent() : Optional.of(0);
```

hongyegong: We do need to take care of the file-already-exists case, in which chase we want to set some generation ID if overwrite is false, I think better to just consolidate into one method.

hongyegong: This sounds about right but getWriteGeneration() method actually fails request directly without making request to GCS if overwrite set to false, I think better to just consolidate into one method and save one request.

Here I tend to maintain the original logic as much as possible before I am clear about the impact of set/unset ifMatchGeneration on other requests.

medb: It actually does not save a request, because `getWriteGeneration()` always makes at least couple requests in the `getItemInfo(resourceId)` method.

I guess the only edge case, is if this logic somehow allows to handle directory overwrite with file, i.e. what happens when you write `gs://a/b/c` file if directory `gs://a/b/c/` already exists? If it's not handled, then we can just remove `getWriteGeneration()` altogether (if it used only here).

hongyegong: This method is also used by compose method as well. Removal of it causes some irrelevant tests to fail. 

About the edge case you mentioned we actually handle this case - by failing the attempt - at the beginning stage of the creation, so we have this check`checkNoDirectoryConflict` in create method of GoogleCloudStorageFileSystem.

medb: Maybe compose has the same bug?

Regarding `create` changes in this PR, because it sets existing object generation in overwrite request, it means that it can not overwrite existing objects concurrently, so it does not fix all the use cases for overwrite.

hongyegong: I'll leave compose as is - always trying to set generation, for now for this PR.

With that said, I will let generateWriteGeneration merely do what it means to do - generate generation, rather than mixed with other responsibilities like check overwrite and throw exception etc. For creation it's now updated more like what you suggested earlier but remain assigning generation# for non-overwrite case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -358,9 +358,13 @@ public WritableByteChannel create(final StorageResourceId resourceId, CreateObje
      */
 
     Optional<Long> overwriteGeneration =
-        resourceId.hasGenerationId()
-            ? Optional.of(resourceId.getGenerationId())
-            : getWriteGeneration(resourceId, options.overwriteExisting());
+        options.overwriteExisting()
+            ? Optional.absent()
+            : Optional.of(getWriteGeneration(resourceId));
```
medb, line 363: Aren't setting generation to non-zero here will overwrite object if `options.overwriteExisting()` is `false`?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -358,9 +358,13 @@ public WritableByteChannel create(final StorageResourceId resourceId, CreateObje
      */
 
     Optional<Long> overwriteGeneration =
-        resourceId.hasGenerationId()
-            ? Optional.of(resourceId.getGenerationId())
-            : Optional.of(getWriteGeneration(resourceId, options.overwriteExisting()));
+        options.overwriteExisting()
```
tdas, line 361: I think the comment above should be appropriately updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: Updated

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
```diff
@@ -364,6 +369,112 @@ public void testOverwrite()
...
```
medb, line 374: You need to write some different `text` here to test that files actually overridden with a new content.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
```diff
@@ -364,6 +369,112 @@ public void testOverwrite()
...
+    URI path = GoogleCloudStorageFileSystemIntegrationTest.getTempFilePath();
+    Path hadoopPath = ghfsHelper.castAsHadoopPath(path);
+    assertThrows(FileNotFoundException.class, () -> ghfs.getFileStatus(hadoopPath));
+
+    String text = "Hello World!";
+    ExecutorService executorService = Executors.newFixedThreadPool(2);
+
+    Future<Integer> future1 =
+        executorService.submit(
+            () -> ghfsHelper.writeFile(hadoopPath, text, 1, /* overwrite= */ true));
```
medb, line 431: Write different content and validate that file content equals to either of them.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
```diff
@@ -364,6 +369,112 @@ public void testOverwrite()
...
+    // Get a temp path and ensure that it does not already exist.
+    URI path = GoogleCloudStorageFileSystemIntegrationTest.getTempFilePath();
+    Path hadoopPath = ghfsHelper.castAsHadoopPath(path);
+    assertThrows(FileNotFoundException.class, () -> ghfs.getFileStatus(hadoopPath));
+
+    String text = "Hello World!";
+    ExecutorService executorService = Executors.newFixedThreadPool(2);
+
+    Future<Integer> future1 =
+        executorService.submit(
```
medb, line 430: Submit both futures at the same time: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#invokeAll-java.util.Collection-

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
```diff
@@ -364,6 +369,112 @@ public void testOverwrite()
...
+            () -> ghfsHelper.writeFile(hadoopPath, text, 1, /* overwrite= */ true));
+    Future<Integer> future2 =
+        executorService.submit(
+            () -> ghfsHelper.writeFile(hadoopPath, text, 1, /* overwrite= */ true));
+    // wait until result will be ready
+    Integer numBytesWritten1 = future1.get();
+    Integer numBytesWritten2 = future2.get();
+    assertThat(numBytesWritten1).isEqualTo(text.getBytes(UTF_8).length);
+    assertThat(numBytesWritten2).isEqualTo(text.getBytes(UTF_8).length);
```
medb, line 471: Same comments as above

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -358,9 +357,7 @@ public WritableByteChannel create(final StorageResourceId resourceId, CreateObje
      */
 
     Optional<Long> overwriteGeneration =
-        resourceId.hasGenerationId()
```
medb, line 361: I think that we may need to preserve `resourceId.hasGenerationId()` logic so caller can control if they want to overwrite a specific generation and fail on concurrent writes, or always overwrite regardless of concurrent overwrites. Does any code in connector depends on this logic? Does connector sets generation in `resourceId` parameter at any time?

util/src/main/java/com/google/cloud/hadoop/util/BaseAbstractGoogleAsyncWriteChannel.java
```diff
@@ -260,6 +262,14 @@ private T waitForCompletionAndThrowIfUploadFailed() throws IOException {
       exception.addSuppressed(e);
       throw exception;
     } catch (ExecutionException e) {
+      GoogleJsonResponseException exception =
+          ApiErrorExtractor.getJsonResponseException(e.getCause());
+      if (exception != null && exception.getStatusCode() == 412) {
```
medb, line 267: We need to create a new method in `ApiErrorExtractor` to check this (see other methods in this class for an example)

util/pom.xml
```diff
@@ -100,5 +100,10 @@
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
```
medb, line 104: `util` project can not depend on Hadoop jars.

hongyegong: Deleted.

util/src/main/java/com/google/cloud/hadoop/util/BaseAbstractGoogleAsyncWriteChannel.java
hongyegong: N/A anymore

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: Per discussion this will be kept.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
hongyegong: Done.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
hongyegong: Done.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/HadoopFileSystemTestBase.java
hongyegong: Done.

PR #419, opened 2020-07-24
Return deep copy of FileStatus objects from InMemoryGlobberFileSystem

Gaurangi94:
`FileStatus` objects returned by methods in `InMemoryGlobberFileSystem` can be modified by the caller. We want to keep state of the `InMemoryGlobberFileSystem` immutable.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
@@ -108,11 +108,9 @@ class InMemoryGlobberFileSystem extends FileSystem {
     }
 
     FileStatus[] fileStatusArr = new FileStatus[fileStatuses.size()];
-    int i = 0;
-
-    for (FileStatus fs : fileStatuses)
+    for (int i = 0; i < fileStatuses.size(); i++)
     {
-      fileStatusArr[i++] = new FileStatus(fs);
+      fileStatusArr[i++] = new FileStatus(fileStatuses.get(i));
     }
 
     return fileStatusArr;
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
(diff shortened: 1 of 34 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
```diff
@@ -105,7 +105,7 @@ public Path getWorkingDirectory() {
       throw new FileNotFoundException(
           String.format("Path '%s' (qualified: '%s') does not exist.", f, qualifiedPath));
     }
-    return fileStatuses.toArray(new FileStatus[0]);
+    return cloneFileStatusList(fileStatuses);
```
hongyegong, line 108: This you can simply do something like:

`fileStatuses.stream().map(fileStatus -> new FileStatus(fileStatus)).collect(Collectors.toList());`

to save a method work.

Gaurangi94: Made the clone method inline as per our discussion

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/InMemoryGlobberFileSystem.java
```diff
@@ -105,7 +106,16 @@ public Path getWorkingDirectory() {
...
-    return fileStatuses.toArray(new FileStatus[0]);
+
+    FileStatus[] fileStatusArr = new FileStatus[fileStatuses.size()];
+    int i = 0;
+
+    for (FileStatus fs : fileStatuses)
+    {
+      fileStatusArr[i++] = new FileStatus(fs);
+    }
+
```
medb, line 117: Use regular for loop, because anyway you need an index:

```java
for (int i = 0; i < fileStatusArr.length; i++) {
  fileStatusArr[i] = new FileStatus(fileStatuses.get(i));
}
```

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
```diff
@@ -665,6 +666,41 @@ public void createFile(Path filePath, byte[] data) throws IOException {
...
+    byte[] data = "data".getBytes(StandardCharsets.UTF_8);
+
+    PathFilter FILE_FILTER = path -> path.getName().endsWith(".json");
+
+    createFile(new Path("/parent/date/2020/07/17/0/file1.xml"), data);
+    createFile(new Path("/parent/date/2020/07/17/0/file1.json"), data);
+    createFile(new Path("/parent/date/2020/07/18/0/file2.xml"), data);
+    createFile(new Path("/parent/date/2020/07/18/0/file2.json"), data);
+    createFile(new Path("/parent/date/2020/07/19/0/file3.xml"), data);
+    createFile(new Path("/parent/date/2020/07/19/0/file3.json"), data);
```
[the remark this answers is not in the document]
Gaurangi94: we dont want xml files in our results. I have added that

PR #420, opened 2020-07-28
Insert object data chunks with single stream before resume

hongyegong:
- This PR only puts all inserts before resume into a single stream to save unnecessary overhead around maintaining multiple streams btw client and server, especially for large object upload.

- Resume logic is updated accordingly.

With this optimization, in experiment the gRPC write in connector shows ~55% performance boost for 4GB data upload, ~25% faster for 2GB data upload. The performance increases more as larger data is uploaded as expected.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -176,6 +176,10 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
     private Hasher objectHasher;
     private String uploadId;
+    private int retriesAttempted = 0;
+    // Holds list of most recent number of NUMBER_OF_REQUESTS_TO_RETAIN requests, so upload can be
+    // rewound and re-sent upon transient errors.
+    private TreeMap<Long, ByteString> dataChunkMap = new TreeMap<>();
 
     UploadOperation(InputStream pipeSource) {
       this.pipeSource = new BufferedInputStream(pipeSource, MAX_BYTES_PER_MESSAGE);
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -266,6 +288,35 @@ private InsertObjectRequest buildInsertRequest(
       return requestBuilder.build();
     }
 
+    // Handles the case when a writeOffset of data read previously is being processed.
+    // This happens if a transient failure happens while uploading, and can be resumed by
+    // querying the current committed offset.
+    private InsertObjectRequest buildRequestFromCachedDataChunk(
```
mfschwartz, line 297: How about calling this buildRequestFromBufferedDataChunk? Caching makes people wonder about cache consistency, whereas here you're really just operating off of disk buffers.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -266,6 +288,35 @@ private InsertObjectRequest buildInsertRequest(
...
+          if (entry.getKey() + entry.getValue().size() > writeOffset) {
+            Long writeOffsetToResume = entry.getKey();
+            ByteString chunkData = entry.getValue();
+            request = buildInsertRequest(writeOffsetToResume, chunkData);
+            break;
+          }
+        }
+      }
+      if (request == null) {
+        throw new IOException(
```
mfschwartz, line 313: If this happens the exception will propagate all the way back to the caller. An alternative would be to catch it somewhere in this class and retry from the beginning. That would make uploads more robust for this case (although hopefully it's a very uncommon case).

I guess we could just wait and see if it's a problem before doing that, though.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Make sense, updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: +1 do it until we actual see these kind of exception. 

I think even we do see this happens not uncommon in our tests, 1st thing we could try is to increase the buffer size instead of retrying all over again. The unretriable kindof errors need to be handled by client side anyway.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -188,270 +195,168 @@ public Object call() throws IOException {
...
-          throw new IOException(
-              String.format(
-                  "Too many retry attempts. Resumable upload failed for '%s'", resourceId));
+      requestStreamObserver.onCompleted();
+
+      try {
+        responseObserver.done.await();
+        if (responseObserver.hasTransientError()) {
+          retriesAttempted++;
+          return doResumableUpload();
```
medb, line 247: This is users recursion for retries, can we use iteration instead - it should be much easier to follow iterative code.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -188,270 +195,168 @@ public Object call() throws IOException {
...
+        Thread.currentThread().interrupt();
+        throw new IOException(
+            String.format("Resumable upload failed for '%s'", getResourceString()), e);
+      }
 
       return responseObserver.getResponseOrThrow();
     }
 
+    private InsertObjectRequest buildInsertRequest(
+        long writeOffset, ByteString dataChunk, boolean resumeMode) {
```
medb, line 259: What does `resumeMode` means? Is this mean recovery of failed upload? If so, let's chose better name to reflect this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -188,270 +195,168 @@ public Object call() throws IOException {
...
+          InsertObjectRequest.newBuilder().setUploadId(uploadId).setWriteOffset(writeOffset);
+
+      if (dataChunk.size() > 0) {
+        ChecksummedData.Builder requestDataBuilder =
+            ChecksummedData.newBuilder().setContent(dataChunk);
+        if (checksumsEnabled) {
+          Hasher chunkHasher = Hashing.crc32c().newHasher();
+          for (ByteBuffer buffer : dataChunk.asReadOnlyByteBufferList()) {
+            chunkHasher.putBytes(buffer);
+          }
```
medb, line 270: Extract into `getChunkHash` method

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -188,270 +195,168 @@ public Object call() throws IOException {
...
+          if (!resumeMode) {
+            for (ByteBuffer buffer : dataChunk.asReadOnlyByteBufferList()) {
+              objectHasher.putBytes(buffer);
+            }
```
medb, line 274: Extract into `updateObjectHash` method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -188,270 +195,168 @@ public Object call() throws IOException {
...
-        stub.withDeadlineAfter(WRITE_STREAM_TIMEOUT.toMillis(), MILLISECONDS)
-            .insertObject(responseObserver);
-        try {
-          responseObserver.done.await();
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          throw new IOException(
-              String.format("Resumable upload failed for '%s'", getResourceString()), e);
+      long writeOffset = retriesAttempted > 0 ? getCommittedWriteSize(uploadId) : 0;
+      objectHasher = retriesAttempted > 0 ? objectHasher : Hashing.crc32c().newHasher();
```
medb, line 208: Can you initialize hasher in constructor if `checksumsEnabled` is true?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Renamed resumeFromFailedInsert to better reflect the case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Updated, more so because of multiple places handling errors, recursion is actually more readable to me in that case. Ended up having to added nested while loops and more branches with iteration.

PR #428, opened 2020-08-10
Improve failure message for 'hadoop fs -ls' command

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -1048,7 +1048,11 @@ private StorageResourceId getPrefixId(URI prefix) {
         if (e instanceof InterruptedException) {
           Thread.currentThread().interrupt();
         }
-        throw new IOException(String.format("Failed to listFileInfo for '%s': %s", path, e.getMessage()), e);
+        throw new IOException(
+            String.format(
+                "Failed to listFileInfo for '%s': %s",
+                path, (e instanceof ExecutionException ? e.getCause() : e).getMessage()),
```
hongyegong, line 1054: maybe geMessage to print error description is not necessary? I believe by default printing e as String will include error description as well.

medb: Good point, done.

PR #430, opened 2020-08-12
[grpc] Set GRPC_CHECKSUMS_ENABLED_DEFAULT to false in GCSIO

veblush:
By our decision to turn this off by default. It is now consistent with `GCS_GRPC_CHECKSUMS_ENABLE`.


veblush:
Thanks Igor, the test is updated to pass the test.

PR #433, opened 2020-08-20
Set location when getting BigQuery job status

### LINKED ISSUE #431: .setLocation() is required to retrieve jobs located in a non-US location.
https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/cca54ae3fc0059f2296812fdeeed9dfe4a511b75/bigquery/src/main/java/com/google/cloud/hadoop/io/bigquery/BigQueryHelper.java#L360

githubwua:
When getting a BigQuery job created outside of the default US location, we need to explicitly provide the job location in the get job request.

Fixes #431

### DIFF
--- a/bigquery/CHANGES.md
+++ b/bigquery/CHANGES.md
@@ -2,6 +2,8 @@
 
 1.  Update all dependencies to latest versions.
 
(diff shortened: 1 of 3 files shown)


medb:
Thank you for contribution!

PR #434, opened 2020-08-24
Tune exponential backoff configuration for Cloud Storage requests.

medb:


util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -234,7 +234,17 @@ public void initialize(HttpRequest request) {
     // IOExceptions such as "socket timed out" of "insufficient bytes written" will follow a
     // straightforward backoff.
     HttpBackOffIOExceptionHandler exceptionHandler =
-        new HttpBackOffIOExceptionHandler(new ExponentialBackOff());
+        new HttpBackOffIOExceptionHandler(
```
hongyegong, line 237: Instead of hardcoding a bunch of magic numbers here, I think it's better to make all these constants with variable names if not configurable at all.

medb: Yes, I plan to add properties to configure them in follow up CL. But this is already strictly better than what we have now.

PR #436, opened 2020-08-31
Unified backoff for exception and unsuccessful response handlers

hongyegong:


util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -62,6 +63,18 @@
   // Only used for testing.
   private Sleeper sleeperOverride;
 
+  private final BackOff defaultBackOff =
```
medb, line 66: I don't think that you can share backoff instance, it could be stateless. You need to extract this in the static method instead.

hongyegong: Done

PR #437, opened 2020-09-01
Match the default request range for gRPC reads with Json API

hongyegong:
Terasort workflow succeeds to run again after this change. This change includes

1. Make request default range match with JSON API's behavior.
2. Do not call hasNext on request iterator  when requestContext has been canceled when detect more available content.


hongyegong:
> Can we add a test that verifies new behavior?

Added a test for requesting different sizes.

medb (approved):

PR #441, opened 2020-09-14
Add interceptor to inject remote address to non-ok status discription

WeiranFang:
For debugging purposes, when client receives non-ok status, we want to know the remote peer address.

To achieve this, we need add additional client interceptor to introspect the remote address before error got thrown out, and also inject the address string into the status description.

+cc @veblush

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -407,7 +407,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
                   String.format(
                       "Caught exception for '%s', while uploading to uploadId %s at writeOffset %d."
                           + " Status: %s",
-                      resourceId, uploadId, writeOffset, s.toString()),
+                      resourceId, uploadId, writeOffset, s),
                   t);
         }
         done.countDown();
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -376,7 +376,6 @@ public final class GoogleCloudStorageGrpcWriteChannel
       @Override
       public void onError(Throwable t) {
         Status s = Status.fromThrowable(t);
-        String statusDesc = s == null ? "" : s.getDescription();
 
         if (t.getClass() == StatusException.class || t.getClass() == StatusRuntimeException.class) {
           Code code =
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -393,7 +392,7 @@ public void onError(Throwable t) {
                   String.format(
                       "Caught exception for '%s', while uploading to uploadId %s at writeOffset %d."
                           + " Status: %s",
-                      resourceId, uploadId, writeOffset, statusDesc),
```
hongyegong, line 411: This assumes `s` to be always non-null, is this guaranteed tho?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -146,6 +174,9 @@ private ChannelAndRequestCounter buildManagedChannel() {
             .enableRetry()
             .defaultServiceConfig(getGrpcServiceConfig())
             .intercept(counter)
+            // TODO(weiranf): Temporarily added for debugging,
+            //                will be removed once GCS/DirectPath is stable.
+            .intercept(new RemoteAddressDebuggingInterceptor())
```
hongyegong, line 179: Is the effect of the interceptor here to populate remote address into read/write error status? Make sure you have manually tested it to work.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
WeiranFang: Yes I believe Status.fromThrowable won't return null. And we may want the system to throw NullPointerException if that happens as it's unexpected. WDYT?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
WeiranFang: I have tested manually with fake error status. For read, the exception always got caught with right message. For write, I only see errors got logged for non-transient errors. And I noticed that [getResponseOrThrow](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java#L369) only throws exception for non-transient errors. For transient errors, the client just retries without displaying any loggings. Is that expected?

veblush:
Thanks! This would be helpful to nail the backend!

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: I also think it wouldnt. Let’s keep it this way for now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
hongyegong: Yes this is expected, it will eventually throw non retriable error if all retries also fail.

hongyegong (approved):

PR #447, opened 2020-09-17
Expand scope for service account when authenticating from outside GCE

hongyegong:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1472,30 +1472,34 @@ private Credential getCredential(
       throws IOException, GeneralSecurityException {
     Credential credential = null;
 
+    // In case of impersonating service account, service account scopes need to be expanded in order
+    // to request access token from IAM.
+    ImmutableList<String> scopes =
```
medb, line 1477: Will it be simpler to just make scopes configurable and let users specify scopes that they need?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1472,30 +1472,34 @@ private Credential getCredential(
...
+        isImpersonatingServiceAccount(config)
+            ? CredentialFactory.CLOUD_SCOPES
```
medb, line 1479: Cloud scopes are too broad, for this, I think.

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: I thought about that too, however I feel like this introduces another layer of complexity for user, it maybe only worth it until we have more use cases other than impersonation that need different scopes.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
hongyegong: Is there any side effect for being broad tho? This is actually a valid scope to set for many of cross GCP requests based on some request docs.

medb: It works, but this is less secure than specifying the only necessary scopes.

hongyegong: As resolved offline, update default scope for credential to CLOUD_PLATFORM as best practice.

medb (approved):

PR #448, opened 2020-09-18
Migrate CreateObjectOptions to auto value

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/CreateObjectOptions.java
```diff
@@ -19,126 +19,81 @@
...
-   * @param metadata A dictionary of metadata to apply to created objects
-   */
-  public CreateObjectOptions(
-      boolean overwriteExisting, String contentType, Map<String, byte[]> metadata) {
-    this(overwriteExisting, contentType, DEFAULT_CONTENT_ENCODING, metadata, true);
+  public static Builder builder() {
+    return new AutoValue_CreateObjectOptions.Builder()
+        .setContentEncoding(null)
+        .setContentType(CONTENT_TYPE_DEFAULT)
+        .setEnsureEmptyObjectsMetadataMatch(true)
```
hongyegong, line 41: Shouldn't this default to be false? So that later you create default for empty object you don't need to set it back to false?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -90,12 +95,9 @@ private boolean validateBucketName(String bucketName) {
       return false;
     }
 
-    if (bucketName.length() > 63) {
-      return false;
-    }
+    return bucketName.length() <= 63;
```
hongyegong, line 98: Not on your change.. but it's really weird seeing this random number pops up.  maybe move the comment down below before this line?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/CreateObjectOptions.java
```diff
@@ -19,126 +19,81 @@
...
-   * empty objects to true.
-   * 
-   *
-   * @param overwriteExisting True to overwrite any existing objects with the same name.
-   * @param metadata A dictionary of metadata to apply to created objects.
-   */
-  public CreateObjectOptions(boolean overwriteExisting, Map<String, byte[]> metadata) {
-    this(overwriteExisting, DEFAULT_CONTENT_TYPE, metadata);
-  }
+  public static final CreateObjectOptions DEFAULT_NO_OVERWRITE = builder().build();
```
hongyegong, line 33: Should we still keep a general DEFAULT option here? Am thinking not all methods require the overwrite semantic. Or you think it's better to explicitly indicate overwrite or non-overwrite in default? It's just feel weird that in context that a real default option is required and DEFAULT_OVERWRITE is used. One option I can think of is having a DEFAULT (with overwirte set to true) and a OPTION_NO_OVERWRITE(with no default semantic in it)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -109,7 +111,7 @@ private boolean validateObjectName(String objectName) {
 
   @Override
   public synchronized WritableByteChannel create(StorageResourceId resourceId) throws IOException {
-    return create(resourceId, CreateObjectOptions.DEFAULT);
+    return create(resourceId, CreateObjectOptions.DEFAULT_OVERWRITE);
```
hongyegong, line 114: I guess here is an example of the scenario I mentioned  above - where a DEFAULT option is required but the overwrite one is used.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -51,6 +51,11 @@
  */
 public class InMemoryGoogleCloudStorage implements GoogleCloudStorage {
 
+  private static final CreateObjectOptions EMPTY_OBJECT_CREATE_OPTIONS =
```
hongyegong, line 54: I see you created this several times, should it just go to CreateObjectOptions.java as a public static variable then?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/CreateObjectOptions.java
medb: Yes, I want to be explicit because it's quite important configuration that can cause data loss.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
medb: I will do it in follow-up CL, but this is a testing code, so it's not so critical.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/CreateObjectOptions.java
medb: Yeah, I checked how this change came about, and it seems that it's safer to set it `true` because it will guarantee that created object has the same metadata that we need.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
medb: `DEFAULT` was overwrite option in previously, I just make it explicit.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
medb: This is a test class - it has a lot of duplicate code, but I guess this is WAI, because you do not want silently change tests, when you changing prod code, otherwise tests are less reliable.

hongyegong (approved):

PR #452, opened 2020-09-21
Do not parallelize GoogleCloudStorageFileSystem.listFileInfo

medb:
Parallelizing `listFileInfo` calls leads to excessive list requests and does not provide significant performance boost.


hongyegong (commented):
How about similar pattern used in `getFileInfoInternal` in the same file? Should that also be got rid of for sake of performance?

medb:
> How about similar pattern used in `getFileInfoInternal` in the same file? Should that also be got rid of for sake of performance?

No, there parallelization makes sense because it does not list all files in the directory.

hongyegong (approved):

PR #456, opened 2020-09-23
Refactor list object methods to use ListObjectOptions

medb:
This PR also renames `create` method used to create buckets in `GoogleCloudStorage` class to `createBucket` - this reduces confusion with `create` methods overloaded for object creation.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
@@ -64,13 +64,13 @@ public class ForwardingGoogleCloudStorage implements GoogleCloudStorage {
 
   @Override
   public void createBucket(String bucketName) throws IOException {
-    logger.atFine().log("%s.create(%s)", delegateClassName, bucketName);
+    logger.atFine().log("%s.createBucket(%s)", delegateClassName, bucketName);
     delegate.createBucket(bucketName);
   }
 
   @Override
   public void createBucket(String bucketName, CreateBucketOptions options) throws IOException {
-    logger.atFine().log("%s.create(%s, %s)", delegateClassName, bucketName, options);
+    logger.atFine().log("%s.createBucket(%s, %s)", delegateClassName, bucketName, options);
     delegate.createBucket(bucketName, options);
   }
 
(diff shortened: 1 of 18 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
```diff
@@ -63,15 +63,15 @@ public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOpti
   }
 
   @Override
-  public void create(String bucketName) throws IOException {
+  public void createBucket(String bucketName) throws IOException {
     logger.atFine().log("%s.create(%s)", delegateClassName, bucketName);
```
hongyegong, line 67: Should we update the logging string as well? Here and possibly anywhere else.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -30,24 +30,22 @@
...
 
   /**
-   * Value indicating all objects should be returned from GCS, no limit.
+   * Pseudo path delimiter.
+   *
+   * <p>GCS does not implement full concept of file system paths but it does expose some notion of a
+   * delimiter that can be used with ${code Storage.Objects.List} to control which items are listed.
    */
-  public static final long MAX_RESULTS_UNLIMITED = -1;
+  String PATH_DELIMITER = "/";
```
hongyegong, line 40: Why these constant modifiers got removed?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
medb: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
medb: They are applied by default in Java interfaces - Intellij linter shows them as redundant warning.

hongyegong (approved):

PR #460, opened 2020-09-30
Allow Null Credentials

### LINKED ISSUE #459: GCS connector should be able to make unauth'd requests to GCS public buckets
### Feature Request
Allow GCS connector to make un-auth'd requests to GCS (e.g. list / download file from public bucket) with minimal configuration.
This could be exposed as a new config `fs.gs.auth.none=true`. This is mostly helpful for lightweight CI and demos using public buckets.

### Use Case
When installing GCS connector in a docker image (in my case I am preparing an apache airflow image and need to use gcs connector to stage spark jars on GCS when using SparkSubmitOperator) I should be able to easily verify it's installation in CI script without having to give credentials.

For example my ci script might run :
```shell
docker run --rm "${AIRFLOW_IMAGE}"  \
  bash -c "hadoop fs -D fs.gs.auth.none=true -ls gs://gcp-public-data-landsat && GCS Connector Installation Validated!"
```

jaketf:
closes #459

### DIFF
--- a/util/src/main/java/com/google/cloud/hadoop/util/CredentialOptions.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/CredentialOptions.java
@@ -28,7 +28,7 @@ public abstract class CredentialOptions {
 
   static final boolean SERVICE_ACCOUNT_ENABLED_DEFAULT = true;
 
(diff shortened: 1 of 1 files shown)


medb:
This is already implemented, see https://github.com/GoogleCloudDataproc/hadoop-connectors/issues/459#issuecomment-701866543

PR #461, opened 2020-10-02
Added exponential backoff retry for gRPC write

hongyegong:
This change does:
1. Updated current plain retry to exponential backoff retry.
2. Added exception propogation back and retry as well for two non-gRPC streaming APIs (Resolving [b/169692122](https://b.corp.google.com/issues/169692122)).

### DIFF
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -138,10 +138,6 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.github.rholder</groupId>
-      <artifactId>guava-retrying</artifactId>
-    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -138,6 +138,10 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.rholder</groupId>
+      <artifactId>guava-retrying</artifactId>
+    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
@@ -273,6 +277,7 @@
               <artifactSet>
                 <includes>
                   <include>com.fasterxml.jackson.core:jackson-core</include>
+                  <include>com.github.rholder</include>
                   <include>com.google.api-client</include>
                   <include>com.google.api.grpc</include>
                   <include>com.google.apis</include>
@@ -300,6 +305,7 @@
                   <shadedPattern>com.google.cloud.hadoop.repackaged.gcs.com</shadedPattern>
                   <includes>
                     <include>com.fasterxml.**</include>
+                    <include>com.github.rholder.retry.**</include>
                     <include>com.google.api.**</include>
                     <include>com.google.auth.**</include>
                     <include>com.google.cloud.audit.**</include>
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -265,6 +265,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
         }
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
+        throw new RuntimeException(e);
       }
 
       return responseObserver.getResponseOrThrow();
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -19,6 +19,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.google.storage.v1.ServiceConstants.Values.MAX_WRITE_CHUNK_BYTES;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.stream.Collectors.toMap;
 
 import com.github.rholder.retry.Retryer;
@@ -198,13 +199,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
     }
 
     private Object doResumableUpload() throws IOException {
-      Retryer<Boolean> retryer =
-          RetryerBuilder.<Boolean>newBuilder()
-              .retryIfExceptionOfType(IOException.class)
-              .retryIfRuntimeException()
-              .withWaitStrategy(WaitStrategies.exponentialWait(2, 20_000, MILLISECONDS))
-              .withStopStrategy(StopStrategies.stopAfterAttempt(UPLOAD_RETRIES))
-              .build();
+      Retryer<Boolean> retryer = getRetryer();
 
       // Send the initial StartResumableWrite request to get an uploadId.
       uploadId = startResumableUpload();
@@ -465,13 +460,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
       SimpleResponseObserver<StartResumableWriteResponse> responseObserver =
           new SimpleResponseObserver<>();
-      Retryer<Boolean> retryer =
-          RetryerBuilder.<Boolean>newBuilder()
-              .retryIfExceptionOfType(IOException.class)
-              .retryIfRuntimeException()
-              .withWaitStrategy(WaitStrategies.exponentialWait(2, 20_000, MILLISECONDS))
-              .withStopStrategy(StopStrategies.stopAfterAttempt(UPLOAD_RETRIES))
-              .build();
+      Retryer<Boolean> retryer = getRetryer();
       try {
         retryer.call(
             () -> {
@@ -503,13 +492,7 @@ public final class GoogleCloudStorageGrpcWriteChannel
 
       SimpleResponseObserver<QueryWriteStatusResponse> responseObserver =
           new SimpleResponseObserver<>();
-      Retryer<Boolean> retryer =
-          RetryerBuilder.<Boolean>newBuilder()
-              .retryIfExceptionOfType(IOException.class)
-              .retryIfRuntimeException()
-              .withWaitStrategy(WaitStrategies.exponentialWait(2, 20_000, MILLISECONDS))
-              .withStopStrategy(StopStrategies.stopAfterAttempt(UPLOAD_RETRIES))
-              .build();
+      Retryer<Boolean> retryer = getRetryer();
       try {
         retryer.call(
             () -> {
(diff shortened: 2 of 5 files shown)


medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -470,16 +503,33 @@ private long getCommittedWriteSize(String uploadId) throws IOException {
...
-            try {
-              responseObserver.done.await();
-            } catch (InterruptedException e) {
-              Thread.currentThread().interrupt();
-              throw new RuntimeException(
-                  String.format("Failed to get committed write size for '%s'", getResourceString()),
-                  e);
-            }
+      Retryer<Boolean> retryer =
+          RetryerBuilder.<Boolean>newBuilder()
```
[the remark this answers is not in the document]
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -449,16 +465,33 @@ private String startResumableUpload() throws IOException {
...
-                .startResumableWrite(request, responseObserver);
-            try {
-              responseObserver.done.await();
-            } catch (InterruptedException e) {
-              Thread.currentThread().interrupt();
-              throw new RuntimeException(
-                  String.format("Failed to start resumable upload for '%s'", getResourceString()),
-                  e);
-            }
+      Retryer<Boolean> retryer =
```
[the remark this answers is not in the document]
hongyegong: All retryers are configured same in this method, but need to be different instances.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -193,76 +198,87 @@ public Object call() throws IOException {
...
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          throw new IOException(
-              String.format("Resumable upload failed for '%s'", getResourceString()), e);
-        }
+      Retryer<Boolean> retryer =
+          RetryerBuilder.<Boolean>newBuilder()
+              .retryIfExceptionOfType(IOException.class)
+              .retryIfRuntimeException()
+              .withWaitStrategy(WaitStrategies.exponentialWait(2, 20_000, MILLISECONDS))
```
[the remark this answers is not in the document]
hongyegong: 2 seconds as this framework doesn't support float multiplier as http client library does.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -193,76 +198,87 @@ public Object call() throws IOException {
...
-          }
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          throw new IOException(
-              String.format("Resumable upload failed for '%s'", getResourceString()), e);
-        }
+      Retryer<Boolean> retryer =
+          RetryerBuilder.<Boolean>newBuilder()
+              .retryIfExceptionOfType(IOException.class)
+              .retryIfRuntimeException()
```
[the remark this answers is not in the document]
hongyegong: Good point, the runtime exp comes from CounddownLatch thread interruption. Am not sure if this is a good case to retry tho,  maybe let's remove the runtime retry until we fully figure out when/if it can happen and what the impact is.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -193,76 +198,87 @@ public Object call() throws IOException {
...
-          if (responseObserver.hasTransientError()) {
-            continue;
-          }
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          throw new IOException(
-              String.format("Resumable upload failed for '%s'", getResourceString()), e);
-        }
+      Retryer<Boolean> retryer =
+          RetryerBuilder.<Boolean>newBuilder()
```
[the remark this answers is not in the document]
hongyegong: Extract it into a method instead.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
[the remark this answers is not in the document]
hongyegong: Done

gcs/pom.xml
```diff
@@ -138,6 +138,10 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
```
medb, line 141: Does not look like we need this dependency here.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -171,7 +176,8 @@ public void startUpload(InputStream pipeSource) {
 
     private Hasher objectHasher;
     private String uploadId;
-    private int retriesAttempted = 0;
+    private long writeOffset = 0;
+    InsertChunkResponseObserver responseObserver;
```
medb, line 180: Declare as private? Or add `@VisisbleForTesting`?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done.

gcs/pom.xml
hongyegong: Right, wasn't sure if the shade includes below needs explicit dependency here.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -193,76 +199,81 @@ public Object call() throws IOException {
...
-        }
-
-        return responseObserver.getResponseOrThrow();
+      Retryer<Boolean> retryer = getRetryer();
+
+      // Send the initial StartResumableWrite request to get an uploadId.
+      uploadId = startResumableUpload();
+      try {
+        retryer.call(
+            () -> {
```
medb, line 208: Can we extract this in the a single `doResumableUploadAttempt` method and retry it:
```java
retryer.call(this::doResumableUploadAttempt)
```

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: Done, it does make code look cleaner.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -187,82 +193,81 @@ public void startUpload(InputStream pipeSource) {
...
+      }
 
-        return responseObserver.getResponseOrThrow();
+      try {
+        responseObserver.done.await();
+        if (responseObserver.hasTransientError()) {
+          throw new IOException(responseObserver.transientError);
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
```
medb, line 267: Is this intentional to swallow exception here?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: No, here we should throw runtime exception which is the kind that shouldn't be retried.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -449,16 +455,27 @@ private String startResumableUpload() throws IOException {
...
+              stub.withDeadlineAfter(START_RESUMABLE_WRITE_TIMEOUT.toMillis(), MILLISECONDS)
+                  .startResumableWrite(request, responseObserver);
+              try {
+                responseObserver.done.await();
+                if (responseObserver.hasError()) {
+                  throw new IOException(responseObserver.getError());
+                }
+              } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
```
medb, line 471: ditto

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -187,82 +193,82 @@ public void startUpload(InputStream pipeSource) {
...
 
-        return responseObserver.getResponseOrThrow();
+      try {
+        responseObserver.done.await();
+        if (responseObserver.hasTransientError()) {
+          throw new IOException(responseObserver.transientError);
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new RuntimeException(e);
```
medb, line 268: Should it be something like, to preserve original failure and resource:
```
throw new RuntimeException(
    String.format("Failed to start resumable upload for '%s'", getResourceString()), e);
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -470,20 +487,39 @@ private long getCommittedWriteSize(String uploadId) throws IOException {
...
+              stub.withDeadlineAfter(QUERY_WRITE_STATUS_TIMEOUT.toMillis(), MILLISECONDS)
+                  .queryWriteStatus(request, responseObserver);
+              try {
+                responseObserver.done.await();
+                if (responseObserver.hasError()) {
+                  throw new IOException(responseObserver.getError());
+                }
+              } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
```
medb, line 503: ditto

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
hongyegong: All non-IOExceptions will be catched by the try-catch around retryer and the original failure and resource will be preserved.

medb: In this case it could be better to not catch `InterruptedException` here and special case it in `catch` around `retry` call?

hongyegong: I think this works for me, updated.

medb (approved):

PR #468, opened 2020-10-12
Do not create parent directories for new objects

### LINKED ISSUE #356: Writing to GCS prefix requires permissions in parent "directories" of the bucket
When writing to GCS from Spark using this library to for example `gs://somebucket/prefix1/prefix2/prefix3` it seems like either Spark or this library is trying to create the parent "directories" first.

We limit write access of our Spark applications to a prefix that's specific to the Spark application so they only have access to objects that start with their full prefix (the role binding for `roles/storage.admin` uses `resource.name.startsWith('projects/_/buckets/projectname/objects/prefix1/prefix2/prefix3/` as its condition).

We get the following errors:
```
serviceaccount@projectname.iam.gserviceaccount.com does not have storage.objects.create access to somebucket/prefix1/prefix2/."
serviceaccount@projectname.iam.gserviceaccount.com does not have storage.objects.create access to somebucket/prefix1/."
```
So it seems like either Spark or this library is treating GCS like a regular filesystem and trying to create directories first before writing the data which is obviously not the correct behavior.
Is this something that can be fixed?

medb:
Explicit parent directories objects are not necessary because they will be inferred automatically if object exists.

Partially fixes #356

### DIFF
--- a/bigquery/src/test/java/com/google/cloud/hadoop/io/bigquery/GsonBigQueryInputFormatTest.java
+++ b/bigquery/src/test/java/com/google/cloud/hadoop/io/bigquery/GsonBigQueryInputFormatTest.java
@@ -217,6 +217,7 @@ public class GsonBigQueryInputFormatTest {
 
     // Write values to file.
     Path mockPath = new Path("gs://test_bucket/path/test");
+    ghfs.mkdirs(new Path("gs://test_bucket/"));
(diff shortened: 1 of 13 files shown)


hongyegong (changes requested):
I think this change deserves a release notes?

medb:
> I think this change deserves a release notes?

Done.

hongyegong (approved):

PR #477, opened 2020-10-20
Revert fs.gs.overwrite.generation.mismatch.ignore feature

medb:


hongyegong (approved):

majdyz:
Hi @medb, what was the reason of this PR get reverted ?  Is the issue of the concurrent writes that drives this config to be added already fixed from the backend ?

medb:
@majdyz we reverted this feature because it unsafe and can lead to data loss. Also it's is not necessary for Delata Lake integration after https://github.com/delta-io/delta/pull/560 too.

PR #485, opened 2020-11-04
Unified the constructors of GCSImpl to respect gRPC flag

hongyegong:
/gcbrun

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -267,24 +267,11 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
    *
    * @param gcs Preconstructed Storage to use for I/O.
    */
-  public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage gcs) {
-    Preconditions.checkNotNull(options, "options must not be null");
-
-    logger.atFine().log("GCS(%s)", options.getAppName());
-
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -267,24 +267,11 @@ public GoogleCloudStorageImpl(
...
-  public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage gcs) {
-    Preconditions.checkNotNull(options, "options must not be null");
-
-    logger.atFine().log("GCS(%s)", options.getAppName());
-
-    options.throwIfNotValid();
-
-    this.storageOptions = options;
-
+  public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage gcs) throws IOException {
```
medb, line 270: I think that it should be implemented differently - we should reuse prod constructor in test constructor, not opposite.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: Are both of these public API that can be used in prod?

medb: No, one of them has `@VisibleForTesting` annotation which means that it intended to be used in tests.

PR #489, opened 2020-11-06
Add HCFS benchmark tool

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+                LongSummaryStatistics openLatencyNs = newLongSummaryStatistics(openLatencyNsList);
+                LongSummaryStatistics seekLatencyNs = newLongSummaryStatistics(seekLatencyNsList);
+                LongSummaryStatistics readLatencyNs = newLongSummaryStatistics(readLatencyNsList);
+                LongSummaryStatistics closeLatencyNs = newLongSummaryStatistics(closeLatencyNsList);
+
+                ThreadLocalRandom random = ThreadLocalRandom.current();
+                byte[] readBuffer = new byte[readSize];
+
+                initLatch.countDown();
+                startLatch.await();
```
hongyegong, line 320: Why not `awaitUninterruptibly` here but elsewhere?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+                          readCallBytes.accept(bytesRead);
+                        }
+                        readCallTimeNs.accept(System.nanoTime() - readCallStart);
+                      } while (bytesRead >= 0);
+
+                      readFileBytes.accept(fileBytesRead);
+                      readFileTimeNs.accept(System.nanoTime() - readStart);
+                    }
+                  }
+                } finally {
```
hongyegong, line 220: Is the try-finally block needed here? Does simply stopLatch.countDown() right after the big for loop won't work?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+
+    Path testFile = new Path(args.get("--file"));
+
+    warmup(
+        args,
+        () ->
+            benchmarkRandomRead(
+                fs,
+                testFile,
+                /* numOpen= */ 5,
```
hongyegong, line 270: How are these number chosen? (Asking bcuz it looks very differently than `read`)

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+                        for (int k = 0; k < numReads; k++) {
+                          long seekStart = System.nanoTime();
+                          input.seek(seekPos);
+                          seekLatencyNs.accept(System.nanoTime() - seekStart);
+
+                          long readStart = System.nanoTime();
+                          int numRead = input.read(readBuffer);
+                          readLatencyNs.accept(System.nanoTime() - readStart);
+
+                          if (numRead != readSize) {
```
hongyegong, line 339: Why this needs an error output? Doesn't this mean that it reached the EOF of something like that?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+              }));
+    }
+    executor.shutdown();
+
+    awaitUninterruptibly(initLatch);
+    long startTime = System.nanoTime();
+    startLatch.countDown();
+    awaitUninterruptibly(stopLatch);
+    double runtimeSeconds = nanosToSeconds(System.nanoTime() - startTime);
+    long operations = combineStats(readLatencyNsList).getCount();
```
hongyegong, line 368: How is combining the latencies converted to number of operations?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -0,0 +1,468 @@
...
+                  }
+                } finally {
+                  stopLatch.countDown();
+                }
+                return null;
+              }));
+    }
+    executor.shutdown();
+
+    awaitUninterruptibly(initLatch);
```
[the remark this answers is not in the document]
medb: Yes, and also it ensures that all test threads will start execution at the same time. For example if you will test with 1,000 threads first can start read significantly earlier than last one.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
medb: Because this it run inside the Callable that throws exception anyway, so `.await()` does not force you to add try/catch

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
medb: No, it means that because of some reason no data was actually read, which theoretically a bug when you test a random read preformance.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
medb: Each latency in the list has `count` so it will be just a sum of all individual threads counts.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
medb: Just something that does not run for too long to not make benchmarks too long. Read test always reads whole file, so there less room to reduce warmup phase time.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
medb: Yes, without it if exception will be thrown benchmark will never terminate and will not print this exception.

hongyegong (approved):

PR #490, opened 2020-11-10
Added basic integration tests for gRPC

hongyegong:
This PR adds some basic integration test cases when gRPC enabled, i.e. object read/writes to guard against any future breaking gRPC related changes to core functionality. It doesn't try to cover all the test cases we are covering in other GCSIntegrationTest class, e.g. bucket/object list/delete/copy related tests.

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
@@ -9,17 +9,14 @@ import static org.junit.Assert.assertThrows;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorage;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageImpl;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageItemInfo;
-import com.google.cloud.hadoop.gcsio.ListObjectOptions;
 import com.google.cloud.hadoop.gcsio.StorageResourceId;
-import com.google.common.collect.Lists;
-import com.google.common.flogger.GoogleLogger;
+import com.google.cloud.hadoop.gcsio.integration.GoogleCloudStorageTestHelper.TestBucketHelper;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SeekableByteChannel;
 import java.util.Arrays;
-import java.util.List;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
@@ -28,19 +25,16 @@ public class GoogleCloudStorageGrpcIntegrationTest {
 
   // Prefix this name with the prefix used in other gcs io integrate tests once it's whitelisted by
   // GCS to access gRPC API.
-  private static final String BUCKET_NAME = "gcs-grpc-team-dataproc-it";
+  private static final String BUCKET_NAME_PREFIX = "gcs-grpc-team-dataproc-it";
 
-  private static GoogleCloudStorage createGoogleCloudStorage() {
-    try {
-      return new GoogleCloudStorageImpl(
-          GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true).build(),
-          GoogleCloudStorageTestHelper.getCredential());
-    } catch (IOException e) {
-      GoogleLogger logger = GoogleLogger.forEnclosingClass();
-      logger.atWarning().withCause(e).log(
-          "Caught exception during GCS (%s) buckets creation", BUCKET_NAME);
-      return null;
-    }
+  private static final TestBucketHelper BUCKET_HELPER = new TestBucketHelper(BUCKET_NAME_PREFIX);
+
+  private static final String BUCKET_NAME = BUCKET_HELPER.getUniqueBucketName("shared");
+
+  private static GoogleCloudStorage createGoogleCloudStorage() throws IOException {
+    return new GoogleCloudStorageImpl(
+        GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true).build(),
+        GoogleCloudStorageTestHelper.getCredential());
   }
 
   @BeforeClass
@@ -52,14 +46,7 @@ public class GoogleCloudStorageGrpcIntegrationTest {
   @AfterClass
   public static void cleanupBuckets() throws IOException {
     GoogleCloudStorage rawStorage = createGoogleCloudStorage();
-    // randomize buckets order in case concurrent clean ups are running
-    List<GoogleCloudStorageItemInfo> objectsToDelete =
-        rawStorage.listObjectInfo(
-            BUCKET_NAME, /* objectNamePrefix= */ null, ListObjectOptions.DEFAULT_FLAT_LIST);
-
-    rawStorage.deleteObjects(
-        Lists.transform(objectsToDelete, GoogleCloudStorageItemInfo::getResourceId));
-    rawStorage.deleteBuckets(Arrays.asList(BUCKET_NAME));
+    BUCKET_HELPER.cleanup(rawStorage);
   }
 
   @Test
@@ -106,7 +93,7 @@ public class GoogleCloudStorageGrpcIntegrationTest {
   }
 
   @Test
-  public void testCreateInvalidObject() {
+  public void testCreateInvalidObject() throws IOException {
     GoogleCloudStorage rawStorage = createGoogleCloudStorage();
     StorageResourceId objectToCreate =
         new StorageResourceId(BUCKET_NAME, "testCreateInvalidObject_InvalidObject\n");
@@ -125,7 +112,7 @@ public class GoogleCloudStorageGrpcIntegrationTest {
   }
 
   @Test
-  public void testOpenNonExistentItem() {
+  public void testOpenNonExistentItem() throws IOException {
     GoogleCloudStorage rawStorage = createGoogleCloudStorage();
     assertThrows(
         FileNotFoundException.class,
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
@@ -9,16 +9,15 @@ import static org.junit.Assert.assertThrows;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorage;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageImpl;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageItemInfo;
+import com.google.cloud.hadoop.gcsio.ListObjectOptions;
 import com.google.cloud.hadoop.gcsio.StorageResourceId;
 import com.google.cloud.hadoop.gcsio.ThrottledGoogleCloudStorage;
 import com.google.cloud.hadoop.gcsio.ThrottledGoogleCloudStorage.StorageOperation;
-import com.google.cloud.hadoop.gcsio.integration.GoogleCloudStorageTestHelper.TestBucketHelper;
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
-import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.flogger.GoogleLogger;
 import com.google.common.util.concurrent.RateLimiter;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -27,18 +26,16 @@ import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SeekableByteChannel;
 import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.List;
 import org.junit.AfterClass;
-import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class GoogleCloudStorageGrpcIntegrationTest {
 
-  // This prefix will be replaced by the prefix used in other gcs io integrate tests once it's
-  // whitelisted by GCS to access gRPC API.
-  private static final String BUCKET_NAME_PREFIX = "gcs-grpc-team-dataproc-it";
-
-  private static final Supplier<TestBucketHelper> BUCKET_HELPER =
-      Suppliers.memoize(() -> new TestBucketHelper(BUCKET_NAME_PREFIX));
+  // Prefix this name with the prefix used in other gcs io integrate tests once it's whitelisted by
+  // GCS to access gRPC API.
+  private static final String BUCKET_NAME = "gcs-grpc-team-dataproc-it";
 
   private static final LoadingCache<GoogleCloudStorage, String> SHARED_BUCKETS =
       CacheBuilder.newBuilder()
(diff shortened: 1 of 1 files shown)


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+  private static final Supplier<TestBucketHelper> BUCKET_HELPER =
+      Suppliers.memoize(() -> new TestBucketHelper(BUCKET_NAME_PREFIX));
+
+  private static final LoadingCache<GoogleCloudStorage, String> SHARED_BUCKETS =
+      CacheBuilder.newBuilder()
+          .build(
+              new CacheLoader<GoogleCloudStorage, String>() {
+                @Override
+                public String load(GoogleCloudStorage gcs) throws Exception {
+                  return BUCKET_HELPER.get().getUniqueBucketName("shared") + "_" + gcs.hashCode();
```
medb, line 49: Why do we need bucket per GCS instance?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+            // Allow 2 create or delete bucket operation every second. This will hit rate limits,
+            // but GCS now has back-offs implemented for bucket operations.
+            RateLimiter.create(2),
+            new GoogleCloudStorageImpl(
+                GoogleCloudStorageTestHelper.getStandardOptionBuilder()
+                    .setGrpcEnabled(true)
+                    .build(),
+                GoogleCloudStorageTestHelper.getCredential()),
+            EnumSet.of(StorageOperation.DELETE_BUCKETS, StorageOperation.CREATE_BUCKET));
+    this.rawStorage = gcs;
```
medb, line 67: Tests initialize their setup usually in `@Before` method, not constructor.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  @Before
+  public void setUp() throws IOException {
+    rawStorage.createBucket(getSharedBucketName());
```
medb, line 72: Tests should reuse buckets, because it takes time to create bucket and increases tests flakiness.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  @AfterClass
+  public static void cleanupBuckets() throws IOException {
+    BUCKET_HELPER.get().cleanup(Iterables.getLast(SHARED_BUCKETS.asMap().keySet()));
+  }
+
+  private String getSharedBucketName() {
```
medb, line 80: Why do we need this?

hongyegong: We use this cache as a mean to persist a <storage, bucketName> so not to create random bucket name with helper class each time, plus we can use a trick to reference the storage object itself from cache in @Afterclass static method, but this name will not be accessible in @Beforeclass method if we want to reuse same bucket i.e. only create bucket once.

Refactored to just use original bucket prefix as common bucket name.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Yeah, turns out the test framework is not flexible enough for tasks that need to be done once before or after all tests.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Moved this into a static filed to make sure it's only created once. (constructor itself will be called as many times as number of tests.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Removed. Not needed.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+  // Prefix this name with the prefix used in other gcs io integrate tests once it's whitelisted by
+  // GCS to access gRPC API.
+  private static final String BUCKET_NAME = "gcs-grpc-team-dataproc-it";
+
+  private static final GoogleCloudStorage rawStorage = getRawStorage();
+
+  private static GoogleCloudStorage getRawStorage() {
+    try {
+      return new ThrottledGoogleCloudStorage(
```
medb, line 44: Why do we need to wrap in `ThrottledGoogleCloudStorage`?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,208 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class GoogleCloudStorageGrpcIntegrationTest {
+
+  // Prefix this name with the prefix used in other gcs io integrate tests once it's whitelisted by
+  // GCS to access gRPC API.
+  private static final String BUCKET_NAME = "gcs-grpc-team-dataproc-it";
+
+  private static final GoogleCloudStorage rawStorage = getRawStorage();
```
medb, line 40: I think it will make tests more flexible and independent if each test will have its own instance of the `GoogleCloudStorage` - overhead of creating instance per test should be negligible.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,208 @@
...
+      GoogleLogger logger = GoogleLogger.forEnclosingClass();
+      logger.atWarning().withCause(e).log(
+          "Caught exception during GCS (%s) buckets creation", BUCKET_NAME);
+      return null;
+    }
+  }
+
+  @BeforeClass
+  public static void createBuckets() throws IOException {
+    rawStorage.createBucket(BUCKET_NAME);
```
medb, line 62: I think that you need to use integration test helper to create and cleanup shared test bucket.

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: I think this is to rate-limit the request so that GCS doesn't throttle - might now be necessary here but I think we use this as a good practice in some of other integration tests?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: hmm i'm a bit confused. So to create bucket only once as you suggested we need a storage object and it needs to be static due to the @Beforeclass limitation. Are you suggesting we create another per-test rawstorage object in a e.g. @Before method?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: I used to do that, but it doesn't let me create bucket only once, i.e. have a rawstorage instance object and used in @Beforeclass method. If we use helper it will basically have to create bucket per test which contradicts with your comment above..

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
medb: I think that if we will have one bucket per test we do not need this.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
medb: Yes, or we can just call static `createGoogleCloudStorage()` method in each test method directly - there little value in assigning this object to class field.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
medb: It has `getSharedBucket` method that returns shared bucket per helper instance.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: if you are talking about [getShareBucketName](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/09a4e9050c9acacf5acd497d38849a33fd21efe2/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTest.java#L161), this uses cache mapping from gcs object to a unique bucket name and key is a single rawstorage object shared across all tests. In this test if we have multiple gcs objects then we cannot resuse this method because each time a new unique bucket name will be generated for a new storage object.

All the cleanup method of this helper class does is to delete objects + bucket, which is what this following `@Afterclass` method does, so I think this should be equivalent and most suitable way for this test.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+  // GCS to access gRPC API.
+  private static final String BUCKET_NAME = "gcs-grpc-team-dataproc-it";
+
+  private static GoogleCloudStorage createGoogleCloudStorage() {
+    try {
+      return new GoogleCloudStorageImpl(
+          GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true).build(),
+          GoogleCloudStorageTestHelper.getCredential());
+    } catch (IOException e) {
+      GoogleLogger logger = GoogleLogger.forEnclosingClass();
```
medb, line 39: I think it's better to add `throws IOException` to the method signature and delete this try-catch block.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+      logger.atWarning().withCause(e).log(
+          "Caught exception during GCS (%s) buckets creation", BUCKET_NAME);
+      return null;
+    }
+  }
+
+  @BeforeClass
+  public static void createBuckets() throws IOException {
+    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    rawStorage.createBucket(BUCKET_NAME);
```
medb, line 49: Using a static bucket name will not work if 2 tests are running at the same time in different projects - you need to randomize bucket name.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+  public static void cleanupBuckets() throws IOException {
+    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    // randomize buckets order in case concurrent clean ups are running
+    List<GoogleCloudStorageItemInfo> objectsToDelete =
+        rawStorage.listObjectInfo(
+            BUCKET_NAME, /* objectNamePrefix= */ null, ListObjectOptions.DEFAULT_FLAT_LIST);
+
+    rawStorage.deleteObjects(
+        Lists.transform(objectsToDelete, GoogleCloudStorageItemInfo::getResourceId));
+    rawStorage.deleteBuckets(Arrays.asList(BUCKET_NAME));
```
medb, line 62: Cleanup in integration test helper is more complicated because it deletes leaked buckets as well, not only current test bucket, ideally we should reuse it here.

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,193 @@
...
+    try (SeekableByteChannel readChannel = rawStorage.open(resourceId)) {
+      for (int i = 0; i < segmentCount; i++) {
+        ByteBuffer segmentBuffer = ByteBuffer.wrap(readSegments[i]);
+        int bytesRead = readChannel.read(segmentBuffer);
+        assertThat(bytesRead).isEqualTo(segmentSize);
+        byte[] expectedSegment =
+            Arrays.copyOfRange(
+                data,
+                i * segmentSize, /* from index */
+                (i * segmentSize) + segmentSize /* to index */);
```
medb, line 164: Fix variable comments:

```java

            Arrays.copyOfRange(
                data,
                /* from= */ i * segmentSize,
                /* to= */ i * segmentSize + segmentSize);
```

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,193 @@
...
+        writeObject(rawStorage, resourceId, /* partitionSize= */ 10 * 1024 * 1024, partitionsCount);
+
+    assertObjectContent(rawStorage, resourceId, partition, partitionsCount);
+  }
+
+  @Test
+  public void testPartialRead() throws IOException {
+    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    final int segmentSize = 553;
+    final int segmentCount = 5;
```
medb, line 148: `final` modifiers for local variables are redundant.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -0,0 +1,193 @@
...
+            .that(readSegments[i])
+            .isEqualTo(expectedSegment);
+      }
+    }
+  }
+
+  @Test
+  public void testChannelClosedException() throws IOException {
+    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    final int totalBytes = 1200;
```
medb, line 175: Remove `final`

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
hongyegong: Done.

medb (approved):

PR #492, opened 2020-11-18
Remove redundant constructors

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -138,40 +140,31 @@
...
    * GoogleCloudStorage {@code gcs}. Any options pertaining to GCS creation will be ignored.
    */
   @VisibleForTesting
   public GoogleCloudStorageFileSystem(
-      GoogleCloudStorage gcs, GoogleCloudStorageFileSystemOptions options) {
-    this.gcs = gcs;
+      CheckedFunction<GoogleCloudStorageOptions, GoogleCloudStorage, IOException> gcsFn,
+      GoogleCloudStorageFileSystemOptions options)
+      throws IOException {
+    this.gcs = gcsFn.apply(options.getCloudStorageOptions());
```
hongyegong, line 167: Would it be better to construct GoogleCloudStorageFileSystemOptions and pass in a GCSOption here so that you dont have to do `GoogleCloudStorageFileSystemOptions.builder()
              .setCloudStorageOptions(getInMemoryGoogleCloudStorageOptions())
              .build()` for every caller?

util/src/main/java/com/google/cloud/hadoop/util/CheckedFunction.java
```diff
@@ -0,0 +1,86 @@
...
+   * it is relayed to the caller of the composed function.
+   *
+   * @param <V> the type of input to the {@code before} function, and to the composed function
+   * @param before the function to apply before this function is applied
+   * @return a composed function that first applies the {@code before} function and then applies
+   *     this function
+   * @throws NullPointerException if before is null
+   * @see #andThen(Function)
+   */
+  default <V> CheckedFunction<V, R, E> compose(Function<? super V, ? extends T> before) throws E {
```
hongyegong, line 55: Are these methods except `apply` being used anywhere?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
medb: Different tests want to create different instances of `CloudStorageInterface`, that's why I do not think that just passing options class will work.

util/src/main/java/com/google/cloud/hadoop/util/CheckedFunction.java
medb: No, it's just copy-paste from Java's `Function` class so I took everything that it had.

hongyegong (approved):

PR #496, opened 2020-11-23
Cleanup and delete unused code

medb:
#cleanup

Note that thus PR removes `ThrottledGoogleCloudStorage` because integration tests should use shared bucket instead of creating many test buckets with rate limiting, also HTTP client should handle low-level retries.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -278,15 +274,15 @@ public int read(ByteBuffer byteBuffer) throws IOException {
     return bytesRead;
   }
 
-  private void requestObjectMedia(OptionalLong bytesToRead) throws IOException {
+  private void requestObjectMedia(@Nullable Integer bytesToRead) throws IOException {
```
bsidhom, line 277: Do we definitely want to narrow this to an integer?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageIntegrationTest.java
```diff
@@ -61,13 +56,8 @@ private static GoogleCloudStorage getGoogleCloudStorage() throws IOException {
...
       GoogleCloudStorageOptions.Builder optionsBuilder) throws IOException {
-    Credential credential = GoogleCloudStorageTestHelper.getCredential();
-    return new ThrottledGoogleCloudStorage(
-        // Allow 2 create or delete bucket operation every second. This will hit rate limits,
-        // but GCS now has back-offs implemented for bucket operations.
-        RateLimiter.create(2),
-        new GoogleCloudStorageImpl(optionsBuilder.build(), credential),
-        EnumSet.of(StorageOperation.DELETE_BUCKETS, StorageOperation.CREATE_BUCKET));
+    return new GoogleCloudStorageImpl(
+        optionsBuilder.build(), GoogleCloudStorageTestHelper.getCredential());
```
bsidhom, line 60: It would be helpful to include a commit/review comment about why we're removing the rate limiting.

bsidhom (approved):
LGTM. Just want two small questions.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageIntegrationTest.java
medb: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
medb: Yes, because `ByteBuffer.remaining` that passed as a parameter here uses `int`, so there no sense to use `long`

bsidhom (approved):
LGTM

PR #497, opened 2020-11-23
List +1 object only when prefix object is not included in the list result

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -1227,13 +1227,20 @@ private void listStorageObjectsAndPrefixes(
         listedPrefixes != null && listedPrefixes.isEmpty(),
         "Must provide a non-null empty container for listedPrefixes.");
 
+    // List +1 object if prefix should not be included in the result,
```
functicons, line 1230: Add an example to better explain why.

functicons (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: Done.

functicons (approved):

PR #499, opened 2020-11-25
Fix flaky tests

medb:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -196,8 +196,7 @@ public void conflictingWrites_noOverwrite_lastFails() throws IOException {
                 resourceId.getBucketName(),
                 resourceId.getObjectName(),
                 /* generationId= */ 4,
-                /* replaceGenerationId= */ 2))
-        .inOrder();
```
cyxxy, line 200: Can you explain why the order cannot be guaranteed?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/CsekEncryptionIntegrationTest.java
```diff
@@ -72,10 +70,7 @@ public void rewriteObject() throws IOException {
     gcs.createBucket(srcBucketName);
 
     String dstBucketName = BUCKET_HELPER.getUniqueBucketName("rewrite-dst");
-    // Create destination bucket with different location and storage class,
-    // because this is supported by rewrite but not copy requests
-    gcs.createBucket(
-        dstBucketName, CreateBucketOptions.builder().setStorageClass("coldline").build());
+    gcs.createBucket(dstBucketName);
```
cyxxy, line 73: Can you explain the storage class change?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
medb: Because test creates 2 files in parallel that leads to flakiness in some cases when underlaying requests are sent in different order.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/CsekEncryptionIntegrationTest.java
medb: This was copy paste from rewrite requests test where this is necessary, but for CMEK test bucket storage class is irrelevant.

cyxxy (approved):

PR #506, opened 2020-12-03
Add json credentials support for grpc

hongyegong:


util/src/main/java/com/google/cloud/hadoop/util/CredentialFactory.java
```diff
@@ -69,6 +72,24 @@
 
   static final String CREDENTIAL_ENV_VAR = "GOOGLE_APPLICATION_CREDENTIALS";
 
+  public static class CredentialWrapper {
```
medb, line 75: This seems like very invasive solution. Can we implement a `CredentialCredentials` wrapper that extends [Credentials](https://github.com/googleapis/google-auth-library-java/blob/master/credentials/java/com/google/auth/Credentials.java) class and delegates to [Credential](https://github.com/googleapis/google-oauth-java-client/blob/master/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/Credential.java) instance that we already have, instead?

medb (changes requested):

PR #507, opened 2020-12-03
Add Hadoop Service lifecycle support for delegation token binding integrations

pzampino:
This change adds Hadoop Service lifecycle support for delegation token bindings integrations, such that these integrations can respond to lifecycle events from the FileSystem implementations.

This is especially helpful for allowing delegation token binding implementations the opportunity to clean up any resources they've allocated when the associated FileSystem instance is closed.

This change also aligns the GoogleHadoopFileSystem more closely with the corresponding Hadoop FileSystem implementations for S3 and ABFS, which already support this Service lifecycle facility.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
@@ -121,6 +121,7 @@ import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
+import org.apache.hadoop.service.ServiceStateException;
 import org.apache.hadoop.util.Progressable;
 
 /**
@@ -482,6 +483,7 @@ public abstract class GoogleHadoopFileSystemBase extends FileSystem
     dts.bindToFileSystem(this, service);
     try {
       dts.init(config);
+      dts.start();
       delegationTokens = dts;
       if (delegationTokens.isBoundToDT()) {
         logger.atFine().log(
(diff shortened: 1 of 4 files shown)


medb:
@pzampino Thanks for the PR! Please sign a CLA.

pzampino:
I've signed the CLA.

medb:
@googlebot I signed it!

pzampino:
What needs to be done wrt the codecov check?

medb:
I think it's just a flakiness - I will take a closer look at PR and if everything OK will merge it.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -482,18 +483,29 @@ private void initializeDelegationTokenSupport(Configuration config, URI path) th
...
       logger.atFiner().withCause(e).log("Failed to initialize delegation token support");
     }
   }
 
+  private void stopDelegationTokens() {
+    if (delegationTokens != null) {
+      try {
+        delegationTokens.close();
+      } catch (IOException e) {
+        logger.atSevere().withCause(e).log("Failed to stop delegation tokens support");
```
cyxxy, line 504: Can you explain why it is OK to swallow the exception?

pzampino: This code is only invoked when the FileSystem close() method is called (i.e., the FS is no longer needed). Since this ultimately triggers the AbstractService#stop() invocation, the only real possibility of encountering an Exception here is if the underlying binding implements serviceStop() and throws an exception therefrom. In which case, I don't think the Exception should be propagated. It is logged here, but what can be done by the FileSystem or its closer in terms of handling the error? Probably nothing. And while it's possible that some resource(s) might be leaked as a consequence, I believe the log message is sufficient (note that it's level SEVERE).

cyxxy (approved):

pzampino:
Is there anything else needed from my side for this to be merged?

medb (approved):

medb:
Thank you for contribution!

PR #509, opened 2020-12-04
Add fs.gs.create.items.conflict.check.enable property

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -231,8 +231,9 @@ public WritableByteChannel create(URI path, CreateFileOptions createOptions) thr
               "Cannot create a file whose name looks like a directory: '%s'", resourceId));
     }
 
-    // Before creating a leaf directory we need to check if there are no conflicting files
-    // with the same name as any subdirectory
+    // Before creating a file we need to check:
+    // - if there are no conflicting files with the same name as any parent subdirectory
+    // - if there are no conflicting directory with the name as a file
```
hongyegong, line 236: These comments are a bit confusing, what is "conflicting directory with the name as a file", is it file or directory? Should it be - check if there are no conflicting directories with the same name as the file being created?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -115,7 +116,9 @@
    * authority corresponding to that root path in the process.
    */
   public static final HadoopConfigurationProperty<Boolean> GCE_BUCKET_DELETE_ENABLE =
-      new HadoopConfigurationProperty<>("fs.gs.bucket.delete.enable", false);
+      new HadoopConfigurationProperty<>(
```
hongyegong, line 119: Do we really wanna do this? This looks to me less obvious by using a default. I feel we better unify the way to define these default values, not mixed.

hongyegong (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
medb: Clarified and added example.

medb (commented):
Addressed comments.

hongyegong (approved):

PR #514, opened 2020-12-16
Upgrade protobuf to 3.14.0

itay:
This is reported as VULNDB-243350 and VULNDB-243351 by Sysdig/Anchore with 3.13.0.


medb:
Thank you for contribution!

medb (approved):

itay:
Thanks for the quick merge @medb 

Do you have any sense when it will be available in a published release? Just looking to get the vulnerability scanner to not complain 😀

PR #515, opened 2020-12-17
Add gcsio version as usre agent for gRPC

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -53,6 +54,13 @@
   private final ExecutorService backgroundTasksThreadPool;
   private final List<ChannelAndRequestCounter> mediaChannelPool;
 
+  private static final String PROPERTIES_FILE = "gcsio.properties";
+  private static final String VERSION_PROPERTY = "gcsio.version";
+  private static final String UNKNOWN_VERSION = "0.0.0";
+
+  private static final String VERSION = PropertyUtil.getPropertyOrDefault(
+      StorageStubProvider.class, PROPERTIES_FILE, VERSION_PROPERTY, UNKNOWN_VERSION);
+
```
medb, line 63: Can we reuse the same user agent as we use for HTTP requests?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
hongyegong: In theory we could, but that will need to add irrelevant properties in GCSReadOptions in order to plumb that piece of data thru. This is just following the same way we do in other modules.

ALso depends on what we want to include here, userAgent for http will include other infomation I think which might not apply here.

medb: I think that we need to include the same user agent in gRPC requests for consistency with GCS requests.

Probably issue here is that we initialize stub in the gRPC channels whereas HTTP channels accept pre-configured `storage` object?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -53,6 +53,8 @@
   private final ExecutorService backgroundTasksThreadPool;
   private final List<ChannelAndRequestCounter> mediaChannelPool;
 
+  private String userAgent;
```
medb, line 56: Make it `final`

medb (approved):

PR #529, opened 2020-12-22
Add more directory delete and rename integration tests

medb:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
```diff
@@ -666,6 +668,34 @@ public void delete_file_parallel() throws Exception {
     assertThat(gcsFs.exists(bucketUri.resolve(dirObject + "/f1"))).isFalse();
   }
 
+  @Test
+  public void delete_directory_parallel() throws Exception {
+    GoogleCloudStorageFileSystemOptions gcsFsOptions =
+        newGcsFsOptions().setStatusParallelEnabled(true).build();
```
functicons, line 674: How is parallel=true reflected in the verification? Or if we set parallel=false, how would the verification be different?

medb: Sometimes it affects - the difference is in assert of sent GCS requests. You can see it in `delete_file_sequential` and `delete_file_parallel` tests. For directory deletes there are no difference in GCS requests, but it's still better to have tests that ensure that this is the case.

functicons (approved):

PR #532, opened 2020-12-30
Support google credentials for gRPC with existing credential

hongyegong:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -241,14 +241,18 @@ public Boolean load(String bucketName) throws Exception {
    */
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Credential credential)
       throws IOException {
-    this(options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions()));
+    this(
+        options,
+        createStorage(
+            options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions())),
+        credential);
```
medb, line 248: Instead of adding this constructor and duplicating `credential` that already passed in the `RetryHttpInitializer` may you add a package-scoped getter in the `RetryHttpInitializer` class for `credential` field?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -153,12 +165,51 @@ private ChannelAndRequestCounter buildManagedChannel() {
     return new ChannelAndRequestCounter(channel, counter);
   }
 
-  public StorageBlockingStub getBlockingStub() {
-    return StorageGrpc.newBlockingStub(getManagedChannel());
+  public StorageBlockingStub getBlockingStub() throws IOException {
+    StorageBlockingStub blockingStub = StorageGrpc.newBlockingStub(getManagedChannel());
+    try {
+      if (credential instanceof GoogleCredential) {
```
medb, line 171: In what cases it will be not `GoogleCredential`?

hongyegong: When authenticating with default GCE vm service account, in that case we can assume it's running on GCE env and credential will be auto-loaded without the need to explicitly pass it in.

medb: In this case I think that it's better to be explicit and check that metadata server authentication is configured instead of checking credential class.

hongyegong: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -153,12 +168,51 @@ private ChannelAndRequestCounter buildManagedChannel() {
...
+    } catch (URISyntaxException e) {
+      throw new IOException("Token server URI could not be parsed.", e);
+    }
+    return blockingStub;
   }
 
-  public StorageStub getAsyncStub() {
-    return StorageGrpc.newStub(getManagedChannel()).withExecutor(backgroundTasksThreadPool);
+  public StorageStub getAsyncStub() throws IOException {
+    StorageStub asyncStub = StorageGrpc.newStub(getManagedChannel());
```
medb, line 194: I think it will be more readable and efficient if we wrap this in `if` and immediately return from it:

```java
if (credentialFactory.useMetadataService()) {
  return StorageGrpc.newBlockingStub(getManagedChannel());
}
```

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
medb: Actually, can we avoid this, and initialize credential channel from passed credential object always?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
hongyegong: I think branch is inevitable unless we can make the Credential used in default matadata service case a GoogleCredential. So the current logic would essentially be use GCE default SA (this will be done by grpc lib if we don't provide a credential on the stub) by default otherwise use the provided credential. I feel like this way is the most feasible and straightforward way to solve this problem without overengineering it.

medb:
This PR was superseded by #534

PR #542, opened 2021-03-11
Make one more GoogleCloudStorageImpl constructor public

Fokko:
This constructor is being used by Apache Beam:
https://github.com/apache/beam/blob/master/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/util/GcsUtil.java#L201


medb (approved):

Fokko:
Thanks for picking this up so fast @medb Any idea when a new release will be done?

medb:
> Thanks for picking this up so fast @medb Any idea when a new release will be done?

np, probably in mid-April - we are waiting on new gRPC Java release that have changes that we want to include in next GCS connector release.

suztomo:
(Coming from https://github.com/apache/beam/pull/14200)
It's mid-April and I see gRPC Java 1.37.0 has been released 13 days ago. How is this going?

PR #544, opened 2021-03-15
Update Hadoop 3 version

mprashanthsagar:


gcs/src/test/resources/contract/gs.xml
```diff
@@ -93,4 +93,14 @@
     <name>fs.contract.rename-overwrites-dest</name>
     <value>false</value>
   </property>
+
+  <property>
```
medb, line 97: Do they have any new HCFS Contract tests that we need to add/run too?

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,23 @@ protected int getDefaultPort() {
     return result;
   }
 
+  public boolean hasPathCapability(Path path, String capability) throws IOException {
+    switch (validatePathCapabilityArgs(path, capability)) {
+      case "fs.capability.paths.append":
+      case "fs.capability.paths.concat":
```
medb, line 528: Maybe add TODO to switch to constants when we will drop Hadoop 2 support?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,23 @@ protected int getDefaultPort() {
...
+        return true;
+      default:
+        return false;
+    }
+  }
+
+  public static String validatePathCapabilityArgs(final Path path, final String capability) {
```
medb, line 535: Remove `final` from parameters - we usually do not add them in GCS connector.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,23 @@ protected int getDefaultPort() {
...
+    checkArgument(path != null, "null path");
```
medb, line 536: You can use `checkNotNull` instead.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,23 @@ protected int getDefaultPort() {
...
+    checkArgument(capability != null, "capability parameter is null");
+    checkArgument(!capability.isEmpty(), "capability parameter is empty string");
```
medb, line 538: You can combine this check into one w/ Guava's `isNullOrEmpty`:

```
checkArgument(!isNullOrEmpty(capability), ...);
```

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,23 @@ protected int getDefaultPort() {
...
+    return capability.toLowerCase(Locale.ENGLISH);
```
medb, line 539: Usually we use `UTF-8` `in such cases.

medb (changes requested):

PR #546, opened 2021-03-16
Fix proxy configuration for Apache HTTP transport

medb:


util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -140,10 +140,13 @@ public static ApacheHttpTransport createApacheHttpTransport(
...
-    HttpClientBuilder httpClientBuilder =
-        ApacheHttpTransport.newDefaultHttpClientBuilder()
-            .setProxy(
-                proxyUri == null ? null : new HttpHost(proxyUri.getHost(), proxyUri.getPort()));
+    HttpClientBuilder httpClientBuilder = ApacheHttpTransport.newDefaultHttpClientBuilder();
+
+    if (proxyUri != null) {
+      httpClientBuilder
+          .setRoutePlanner(null)
+          .setProxy(new HttpHost(proxyUri.getHost(), proxyUri.getPort()));
```
mprashanthsagar, line 148: Can we add a test here to ensure `setProxy` overrides `setRoutePlanner` for posterity

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -140,10 +140,13 @@ public static ApacheHttpTransport createApacheHttpTransport(
...
 
-    HttpClientBuilder httpClientBuilder =
-        ApacheHttpTransport.newDefaultHttpClientBuilder()
-            .setProxy(
-                proxyUri == null ? null : new HttpHost(proxyUri.getHost(), proxyUri.getPort()));
+    HttpClientBuilder httpClientBuilder = ApacheHttpTransport.newDefaultHttpClientBuilder();
+
+    if (proxyUri != null) {
+      httpClientBuilder
+          .setRoutePlanner(null)
```
mprashanthsagar, line 147: we might want to add a comment here to indicating why we need to set RoutePlanner to `null`

mprashanthsagar (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
medb: Done.

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
medb: I tried to add unit test but this is not trivial because proxy settings applied at very late stage of request execution. Probably we will need to do something like [this](https://stackoverflow.com/a/50590840/3227693), but I think that it will be better to track this as a separate AI.

I have tested this manually and it works, not sure about credentials though.

mprashanthsagar (approved):

PR #548, opened 2021-03-16
Improve exception message for Hadoop CLI

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1105,7 +1105,11 @@ String trimToPrefixWithoutGlob(String path) {
       if (e instanceof InterruptedException) {
         Thread.currentThread().interrupt();
       }
-      throw new IOException("Concurrent glob execution failed", e);
+      throw new IOException(
```
mprashanthsagar, line 1108: How about we handle `ExecutionException` and `InterruptedException` separately ?

medb: Done.

mprashanthsagar (approved):

PR #549, opened 2021-03-16
Update Hadoop 3 version to 3.2.2

mprashanthsagar:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -521,6 +522,24 @@ protected int getDefaultPort() {
...
+      // from Hadoop 3 when Hadoop 2 is no longer supported
+      case "fs.capability.paths.append":
+      case "fs.capability.paths.concat":
+        return true;
+      default:
+        return false;
+    }
+  }
+
+  public static String validatePathCapabilityArgs(Path path, String capability) {
```
medb, line 537: Make it private, if it's used only internally

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
mprashanthsagar: we cannot make it public, since the Interface exposes it as `public`

medb: But this is a static method that is not part of the interface? So probably we can make it private?

mprashanthsagar: This is part of the [interface](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FileSystem.html#hasPathCapability-org.apache.hadoop.fs.Path-java.lang.String-)

The build breaks if we make it private

mprashanthsagar: my bad, was checking the other method. making the method private

medb (approved):

PR #552, opened 2021-03-19
Add credentials support to StorageStubProvider

veblush:
This PR does
 - Changed the signature of StorageStubProvider's constructor to have 
   - `Credentials` instead of `Credential` to be aligned with the gRPC-Java supporting only `Credentials` and the Beam SDK using `Credentials` 
   - Explicit `useDirectpath` because it's incomplete to infer this from the given `Credential`.
 -  Added `ComputeGoogleCredential` to the check-code to decide `useDirectpath` 

Note that it needs a future PR to support Beam SDK with directpath because it passes `HttpRequestInitializer` rather than `Credential` when creating `GoogleCloudStorageImpl`. Beam references:

 - How to create `GoogleCloudStorageImpl`: [code](https://github.com/apache/beam/blob/fe1116d1d8285084955d41f61d8039ebf5d9cf55/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/util/GcsUtil.java#L201)
 - How to create `GcsUtil`: [code](https://github.com/apache/beam/blob/fe1116d1d8285084955d41f61d8039ebf5d9cf55/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/util/GcsUtil.java#L105)
 - How to create `HttpRequestInitializer`: [code](https://github.com/apache/beam/blob/fe1116d1d8285084955d41f61d8039ebf5d9cf55/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/util/Transport.java#L97)

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -35,15 +35,12 @@ import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.InputStreamContent;
-import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.json.JsonFactory;
 import com.google.api.client.util.BackOff;
 import com.google.api.client.util.Data;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.Sleeper;
-import com.google.api.services.storage.Storage;
-import com.google.api.services.storage.Storage.Objects.Insert;
-import com.google.api.services.storage.StorageRequest;
 import com.google.api.services.storage.model.Bucket;
 import com.google.api.services.storage.model.Bucket.Lifecycle;
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule;
@@ -54,6 +51,9 @@ import com.google.api.services.storage.model.ComposeRequest;
 import com.google.api.services.storage.model.Objects;
 import com.google.api.services.storage.model.RewriteResponse;
 import com.google.api.services.storage.model.StorageObject;
+import com.google.api.services.storage.Storage;
+import com.google.api.services.storage.Storage.Objects.Insert;
+import com.google.api.services.storage.StorageRequest;
 import com.google.cloud.hadoop.gcsio.authorization.StorageRequestAuthorizer;
 import com.google.cloud.hadoop.util.ApiErrorExtractor;
 import com.google.cloud.hadoop.util.BaseAbstractGoogleAsyncWriteChannel;
@@ -85,19 +85,19 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentHashMap.KeySetView;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentHashMap.KeySetView;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
@@ -276,11 +276,8 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
           httpRequestInitializer instanceof RetryHttpInitializer,
           "request initializer must be an instance of the RetryHttpInitializer class"
               + " when gRPC API enabled");
-      this.storageStubProvider =
-          new StorageStubProvider(
-              this.storageOptions,
-              this.backgroundTasksThreadPool,
-              ((RetryHttpInitializer) httpRequestInitializer).getCredential());
+      Credential credential = ((RetryHttpInitializer) httpRequestInitializer).getCredential();
+      this.storageStubProvider = StorageStubProvider.newInstance(this.storageOptions, this.backgroundTasksThreadPool, credential);
     }
 
     this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(storageOptions);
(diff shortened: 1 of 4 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -158,15 +158,23 @@ private ChannelAndRequestCounter buildManagedChannel() {
...
-        (isComputeCredential()
-                ? GoogleDefaultChannelBuilder.forTarget(target)
-                : ManagedChannelBuilder.forTarget(target))
-            .enableRetry()
-            .defaultServiceConfig(getGrpcServiceConfig())
-            .intercept(counter)
-            .userAgent(userAgent)
-            .build();
+    ManagedChannel channel;
+    if (useDirectpath) {
```
jaceklaskowski, line 162: nit: is it possible to extract the common calls outside `if` so it's clear(er) what's shared between the implementations?

medb:
I think that https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/553 is a simpler and easier to maintain fix, PTAL

veblush:
@medb I updated this CL to have your change. Although your change fixes the benchmark's credential problem, this still needs to support the case of dataflow passing `Credentials`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -145,12 +142,14 @@ public int activeRequests() {
   StorageStubProvider(
       GoogleCloudStorageOptions options,
       ExecutorService backgroundTasksThreadPool,
-      Credential credential) {
+      Credentials credentials,
+      boolean useDirectpath) {
```
medb, line 146: Is an intention to use this constructor directly outside of GCS connector?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
veblush: No this is not about making `StorageStubProvider` public but this is required to incorporate this with Beam. (FYI, I updated the description of PR to have a few pointers to beam code) It passes `HttpRequestInitializer` when creating `GoogleCloudStorageImpl`. Thing is it uses `Credentials` for `HttpRequestInitializer`. At worse, it actually wrapped it with its own private class.

veblush: So next thing we need to figure out is how beam let gcsio know if it uses a service default account. This can be done by either passing its `Credentials` object or explicit intention to use directpath. Both are possible but I like the former because it can be changed later when directpath supports other credentials in future.

veblush: Therefore Credentials needs to be added to the public api of `GoogleCloudStorageImpl`. I guess `GoogleCloudStorageOptions` eventually needs it.

veblush:
@mprashanthsagar PTAL

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -145,12 +142,14 @@ public int activeRequests() {
...
     this.readOptions = options.getReadChannelOptions();
     this.userAgent = options.getAppName();
     this.backgroundTasksThreadPool = backgroundTasksThreadPool;
     this.mediaChannelPool = new ArrayList<>();
-    this.credential = checkNotNull(credential, "credential can not be null");
+    this.credentials = credentials;
+    this.useDirectpath = useDirectpath;
```
mprashanthsagar, line 152: Instead of using flags and branching logic, can we override the relevant methods in a child class ?

ex. An `abstract` class with common logic, override `newBlockingStub` and `newAsycStub` in child classes

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -276,11 +278,15 @@ public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage
           httpRequestInitializer instanceof RetryHttpInitializer,
           "request initializer must be an instance of the RetryHttpInitializer class"
               + " when gRPC API enabled");
+      Credential credential = ((RetryHttpInitializer) httpRequestInitializer).getCredential();
+      boolean useDirectpath = credential != null && java.util.Objects.equals(credential.getTokenServerEncodedUrl(),
+          ComputeCredential.TOKEN_SERVER_ENCODED_URL);
```
mprashanthsagar, line 283: For future reference, I dont think this is the right place for the `useDirectPath` logic , we might be better off with a Factory impl here, so that this logic does not get duplicated now that it is extracted out of the `StorageStubProvider` class

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
veblush: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
veblush: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -145,12 +146,12 @@ public int activeRequests() {
...
       GoogleCloudStorageOptions options,
       ExecutorService backgroundTasksThreadPool,
-      Credential credential) {
+      GrpcDecorator grpcDecorator) {
     this.readOptions = options.getReadChannelOptions();
     this.userAgent = options.getAppName();
     this.backgroundTasksThreadPool = backgroundTasksThreadPool;
     this.mediaChannelPool = new ArrayList<>();
-    this.credential = checkNotNull(credential, "credential can not be null");
+    this.grpcDecorator = grpcDecorator;
```
mprashanthsagar, line 154: should we  add a non null check here ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -209,45 +198,92 @@ private synchronized ManagedChannel getManagedChannel() {
...
+      Map<String, Object> childPolicy = ImmutableMap.of("childPolicy", ImmutableList.of(pickFirstStrategy));
+
+      Map<String, Object> grpcLbPolicy = ImmutableMap.of("grpclb", childPolicy);
+
+      return ImmutableMap.of("methodConfig", ImmutableList.of(methodConfig), "loadBalancingConfig",
+          ImmutableList.of(grpcLbPolicy));
+    }
+  }
+
+  public static StorageStubProvider create(GoogleCloudStorageOptions options, ExecutorService backgroundTasksThreadPool,
```
mprashanthsagar, line 273: how about using `newInstance` instead of `create` ?

mprashanthsagar (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
veblush: Good catch. It should have.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
veblush: yep, it sounds better.

veblush:
@mprashanthsagar Test run appeared to fail to build but I cannot see the actual error. Can you show me the error?

veblush:
Rebased on the HEAD & squashed.

PR #553, opened 2021-03-22
Fix compute credentials identification in gRPC API

medb:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -190,7 +191,9 @@ public StorageStub newAsyncStub() {
   }
 
   private boolean isComputeCredential() {
-    return credential instanceof ComputeCredential;
+    return credential != null
+        && Objects.equals(
+            credential.getTokenServerEncodedUrl(), ComputeCredential.TOKEN_SERVER_ENCODED_URL);
```
mprashanthsagar, line 196: Does this change the intent of the method ? If so, should we rename the method ?

medb: I do not think so, intent is the same, it just depends on actual token URL instead of class.

mprashanthsagar (approved):

PR #554, opened 2021-04-01
Move gRPC call timeout values to configurations

mprashanthsagar:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -355,6 +355,15 @@
   public static final HadoopConfigurationProperty<String> GCS_GRPC_SERVER_ADDRESS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.server.address");
 
+  public static final HadoopConfigurationProperty<Long> GCS_GRPC_READ_STREAM_TIMEOUT_MS =
```
functicons, line 358: Why add `_STREAM_` in the variable name? It feels inconsistent.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
```diff
@@ -46,6 +47,8 @@
   public static final Fadvise DEFAULT_FADVISE = Fadvise.SEQUENTIAL;
   public static final int DEFAULT_MIN_RANGE_REQUEST_SIZE = 2 * 1024 * 1024;
   public static final boolean GRPC_CHECKSUMS_ENABLED_DEFAULT = false;
+  public static final long DEFAULT_GRPC_READ_STREAM_TIMEOUT_MILLIS = 20 * 60 * 1000;
```
functicons, line 50: Consider extracting a global constant class which we could share across the codebase, instead of hardcoding it everywhere.

functicons (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
mprashanthsagar: The constants are spread across two modules `gcsio` and `util`, we do not have a dependency between them, to avoid duplication here.

mprashanthsagar: I am not particularly a fan of global constant class [[ref](https://www.baeldung.com/java-constants-good-practices)]. The constants in this case for default option values seem to be scattered in three packages `gcs` `gcsio` and `util` .

Will sync up offline, to figure out if we can eliminate duplication here and take up separately.

mprashanthsagar: Other option : Import the constants from `gcsio` and `util` packages to `gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java` and replace the hardcoded defaults

functicons: We can address this issue in another PR.

functicons (approved):

PR #562, opened 2021-05-06
Added credentials to the ctor of GoogleCloudStorageImpl 

veblush:
Added credentials to ctor of `GoogleCloudStorageImpl` so that `com.google.auth.Credentials` can be brought to StorageStub which is going to determine whether it will use directpath depending on the type of `Credentials`. Currently `Credentials` object is buried deep in `httpRequestInitializer` so it's not feasible to get the object from it. To address this, explicit field is added to the option. This PR is created with the Beam SDK's [usage](https://github.com/apache/beam/blob/e37bedea7955d2fc13ae44f72a96cc3b763b63e3/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/util/GcsUtil.java#L201) in mind.

In addition to this, `isDirectPathPreffered` is aded to `GoogleCloudStorageOptions` to disable Directpath particularly for the test.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -41,9 +41,6 @@ import com.google.api.client.util.BackOff;
 import com.google.api.client.util.Data;
 import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.Sleeper;
-import com.google.api.services.storage.Storage;
-import com.google.api.services.storage.Storage.Objects.Insert;
-import com.google.api.services.storage.StorageRequest;
 import com.google.api.services.storage.model.Bucket;
 import com.google.api.services.storage.model.Bucket.Lifecycle;
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule;
@@ -54,6 +51,10 @@ import com.google.api.services.storage.model.ComposeRequest;
 import com.google.api.services.storage.model.Objects;
 import com.google.api.services.storage.model.RewriteResponse;
 import com.google.api.services.storage.model.StorageObject;
+import com.google.api.services.storage.Storage;
+import com.google.api.services.storage.Storage.Objects.Insert;
+import com.google.api.services.storage.StorageRequest;
+import com.google.auth.Credentials;
 import com.google.cloud.hadoop.gcsio.authorization.StorageRequestAuthorizer;
 import com.google.cloud.hadoop.util.ApiErrorExtractor;
 import com.google.cloud.hadoop.util.BaseAbstractGoogleAsyncWriteChannel;
@@ -244,7 +245,6 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
     this(options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions()));
   }
 
-  @VisibleForTesting
   public GoogleCloudStorageImpl(
       GoogleCloudStorageOptions options, HttpRequestInitializer httpRequestInitializer)
       throws IOException {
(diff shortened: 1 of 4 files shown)


veblush:
@mprashanthsagar It's now ready to review. PTAL.

veblush:
@mprashanthsagar Can you review this? I like to get this included for the next gcsio release so that I can work on the Beam integration.

mprashanthsagar (changes requested):

mprashanthsagar:
Apologies. I had the comments in couple of weeks ago, missed to publish the review

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -272,12 +273,17 @@ public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage
 
     // Create the gRPC stub if necessary;
     if (this.storageOptions.isGrpcEnabled()) {
-      checkArgument(
+      Credentials credentials = this.storageOptions.getCredentials();
+      if (credentials != null) {
+        this.storageStubProvider = StorageStubProvider.newInstance(this.storageOptions, this.backgroundTasksThreadPool, credentials);
```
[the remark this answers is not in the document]
veblush: Yeah, I'll add the test to cover this code path. For the ctor, this particular cctor is used by Beam which relies on `Credentials` so it needs both `options` and `storage` anyway but Dataproc uses `Credential` so it doesn't need to have it.

veblush:
`setDirectPathPreffered` is added to `GoogleCloudStorageOptions` to disable directpath. Cloudpath benchmark needs a way to disable directpath and it's not easy without this option because directpath over ipv4 is now already enabled by default.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -66,7 +67,7 @@
   private final String userAgent;
   private final ExecutorService backgroundTasksThreadPool;
   private final List<ChannelAndRequestCounter> mediaChannelPool;
-  private final GrpcDecorator grpcDecorator;
+  @VisibleForTesting final GrpcDecorator grpcDecorator;
```
mprashanthsagar, line 70: How about we keep the instance variable `private` and add a `package-private` getter method ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -192,7 +193,7 @@ public Boolean load(String bucketName) throws Exception {
   @VisibleForTesting Storage storage;
 
   // Utility for building and caching storage channels and stubs.
-  private StorageStubProvider storageStubProvider;
+  @VisibleForTesting StorageStubProvider storageStubProvider;
```
mprashanthsagar, line 196: How about we keep the instance variable `private` and add a `package-private` getter method ?

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
veblush: Done.

veblush:
@mprashanthsagar Addressed your comments. PTAL.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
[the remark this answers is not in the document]
veblush: Done.

mprashanthsagar:
LGTM, please re-phrase the commit before merging

mprashanthsagar (approved):

veblush:
@mprashanthsagar Done for a commit and the description of the PR!

PR #567, opened 2021-05-31
Add support for footer prefetch in gRPC read channel

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +127,99 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+      BackOffFactory backOffFactory) throws IOException {
+    StorageBlockingStub stub = stubProvider.newBlockingStub();
+    // TODO(b/135138893): We can avoid this call by adding metadata to a read request.
+    //      That will save about 40ms per read.
+    com.google.google.storage.v1.Object storageObject = getObjectMetadata(resourceId, readOptions,
+        stub);
+
+    Preconditions.checkArgument(storageObject != null, "object metadata cannot be null");
+    long footerOffset = Math
+        .max(0, (storageObject.getSize() - (readOptions.getMinRangeRequestSize() / 2)));
```
davidrabinowitz, line 141: nit: Please create a local variable `long prefatchSizeInBytes = readOptions.getMinRangeRequestSize() / 2;`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +127,99 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+      StorageResourceId resourceId, GoogleCloudStorageReadOptions readOptions,
+      BackOffFactory backOffFactory) throws IOException {
+    StorageBlockingStub stub = stubProvider.newBlockingStub();
+    // TODO(b/135138893): We can avoid this call by adding metadata to a read request.
+    //      That will save about 40ms per read.
+    com.google.google.storage.v1.Object storageObject = getObjectMetadata(resourceId, readOptions,
+        stub);
+
+    Preconditions.checkArgument(storageObject != null, "object metadata cannot be null");
+    long footerOffset = Math
```
davidrabinowitz, line 140: please rename to `footerOffsetInBytes`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +127,99 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+        .max(0, (storageObject.getSize() - (readOptions.getMinRangeRequestSize() / 2)));
+    long footerSize = Math
```
davidrabinowitz, line 142: please rename to `footerSizeInBytes`

davidrabinowitz (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -85,10 +86,18 @@
...
+  CancellableContext requestContext;
 
   Fadvise readStrategy;
 
+  @Nullable
+  private final ByteString footerContent;
+
+  private final long footerStartOffset;
+
+  private final long footerSize;
```
mayanks, line 99: Do we need to store this value? Given that you can always to footerContent.size()?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +127,99 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+    //      That will save about 40ms per read.
+    com.google.google.storage.v1.Object storageObject = getObjectMetadata(resourceId, readOptions,
+        stub);
+
+    Preconditions.checkArgument(storageObject != null, "object metadata cannot be null");
+    long footerOffset = Math
+        .max(0, (storageObject.getSize() - (readOptions.getMinRangeRequestSize() / 2)));
+    long footerSize = Math
+        .min(storageObject.getSize(), (readOptions.getMinRangeRequestSize() / 2));
+    ByteString footerContent = getFooterContent(resourceId, readOptions, stub, footerOffset);
```
mayanks, line 144: Should we fetch the footer if the file is so small that the entire file can be read in this footer read?

mprashanthsagar: The control is upto the user, by configuring `minRangeRequestSize`, I don't see the value bringing in-consistent behaviour based on objectSize.

We could however allow to skip footer caching altogether based on config ( may be determined via file formats or user-config)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -291,9 +364,47 @@ public int read(ByteBuffer byteBuffer) throws IOException {
         bufferedContentReadOffset = bytesToWrite;
       }
     }
+
+    if (footerContent != null && position >= footerStartOffset && byteBuffer.hasRemaining()) {
+      int bytesToWrite = Math.toIntExact(min(byteBuffer.remaining(), footerSize));
```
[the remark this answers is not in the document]
mprashanthsagar: I tend to agree with idea, however i prefer to keep the behaviour consistent with existing logic with `BufferedContent` and JSON impl to avoid in-consistencies

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +124,96 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+    } catch (StatusRuntimeException e) {
+      throw convertError(e, resourceId);
+    }
+  }
+
+  private static ByteString getFooterContent(StorageResourceId resourceId,
+      GoogleCloudStorageReadOptions readOptions, StorageBlockingStub stub, long footerOffset)
+      throws IOException {
+    try {
+      Iterator<GetObjectMediaResponse> footerContentResponse = stub
```
mayanks, line 182: We don't do much logging in this file. I think we should add logs at important decision area, such as (a) reading footer, size etc.., (b) footer cache hit (and miss) etc..

mprashanthsagar: Done

mayanks:
LGTM

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +124,105 @@ static GoogleCloudStorageGrpcReadChannel open(
     }
   }
 
+  private static GoogleCloudStorageGrpcReadChannel openChannel(StorageStubProvider stubProvider,
```
jaceklaskowski, line 127: nit: Would that be considered an improvement to have the input arguments one at a line?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -153,12 +124,105 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+        if (objectMediaResponse.hasChecksummedData()) {
+          ByteString content = objectMediaResponse.getChecksummedData().getContent();
+          if (footerContent == null) {
+            footerContent = content;
+          } else {
+            footerContent = footerContent.concat(content);
+          }
+        }
+      }
+      if (footerContent == null) {
```
jaceklaskowski, line 205: I wonder when `footerContent` could be `null`? Only when `content` is, isn't it? If so, why is this log not printed out earlier?

mayanks (approved):
LGTM

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -245,13 +311,22 @@ public int read(ByteBuffer byteBuffer) throws IOException {
     if (position == objectSize) {
       return bytesRead > 0 ? bytesRead : -1;
     }
+
+    // read request content overlaps with cached footer data
+    if ((footerContent != null) && ((position + bytesToSkipBeforeReading)
```
jaceklaskowski, line 316: nit: Since this condition is fairly complex and `position + bytesToSkipBeforeReading` is used twice, why not to use a local variable for this sum and the whole condition? (It'd make reading this part lot easier IMHO)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -291,9 +366,49 @@ public int read(ByteBuffer byteBuffer) throws IOException {
         bufferedContentReadOffset = bytesToWrite;
       }
     }
+
+    if (footerContent != null && position >= footerStartOffsetInBytes && byteBuffer
```
jaceklaskowski, line 370: I'd be happy to see what the purpose of this condition is by using a local variable (`hasMoreFooterToRead`-like?) that would make this long-ish `if` easier.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -291,9 +366,49 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+    Integer bytesToRead =
+        readStrategy == Fadvise.RANDOM
+            ? max(byteBuffer.remaining(), readOptions.getMinRangeRequestSize())
+            : null;
+    if (footerContent == null) {
+      return bytesToRead;
+    }
+
+    long bytesToFooterOffset = footerStartOffsetInBytes - position;
+    if (bytesToRead == null) {
```
jaceklaskowski, line 392: Isn't this `if` + `return` a mere:

```java
return bytesToRead == null ? Math.toIntExact(bytesToFooterOffset) : Math.toIntExact(min(bytesToRead, bytesToFooterOffset))
```

that seems to lead me to believe there could be even more clever way to handle it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Agreed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: `footerContent` shoudn't be null if `open` succeeded, In case of any modifications in the future, it would be hard to determine and verify this usage. Added this just condition as a defensive approach

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Agreed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Agreed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Personally, not a fan of ternary operator. However, in this case, it does not hamper readability. Will modify the conditional

davidrabinowitz (approved):

PR #569, opened 2021-06-07
Support Hadoop 3.3.1

steveloughran:
To build and test against Hadoop 3.3.1 

* the staging repo needs to be supported as an optional profile
 "snapshots-and-staging". This is only needed when qualifying a hadoop RC.
* AssertJ needs to be added as a dependency.
  (Nobody knows why hadoop-common test jar's declaration isn't picked up)
* the gs.xml contract needs to declare that gcs returns false when
  renaming a file over an existing file, so a new test knows the
  expected behavior.

### DIFF
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -163,7 +163,6 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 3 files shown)


medb:
Will the test dependency be fixed in the next release of Hadoop 3.3.2 / 3.4.0?

steveloughran:
> Will the test dependency be fixed in the next release of Hadoop 3.3.2 / 3.4.0?

afraid not. We don't know what is causing the problem. Hadoop-common builds declare the dependency, it's just somehow it doesn't get through.

medb (approved):
Thank you for the PR!

steveloughran:
thanks!

PR #570, opened 2021-06-08
Switch to json client for metadata operation

mprashanthsagar:


jaceklaskowski (approved):
LGTM

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -126,58 +139,88 @@ static GoogleCloudStorageGrpcReadChannel open(
...
-  private static com.google.google.storage.v1.Object getObjectMetadata(StorageResourceId resourceId,
-      GoogleCloudStorageReadOptions readOptions, StorageBlockingStub stub) throws IOException {
+  private static GoogleCloudStorageItemInfo getObjectMetadata(
+      StorageResourceId resourceId,
+      ApiErrorExtractor errorExtractor,
+      BackOffFactory backOffFactory,
+      Storage gcs) throws IOException {
+    StorageObject object;
     try {
-      // TODO(b/151184800): Implement per-message timeout, in addition to stream timeout.
```
mayanks, line 159: Once read meta data is supported on gRPC, we are going to revert back to this logic right? So should we delete this piece of code or just hide it behind a hard coded flag?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -624,7 +624,8 @@ public SeekableByteChannel open(
         resourceId.isStorageObject(), "Expected full StorageObject id, got %s", resourceId);
 
     if (storageOptions.isGrpcEnabled()) {
-      return GoogleCloudStorageGrpcReadChannel.open(storageStubProvider, resourceId, readOptions);
+      return GoogleCloudStorageGrpcReadChannel
```
mayanks, line 627: I think it will be a good idea to have a separate constructor for GrpcReadChannel, one that exists as of today which internally reads metadata via gRPC, and other which we will introduce in this PR (passing the additional storage and errorExtractor) which will read metaData via JSON API.

Later when getMetaData via gRPC is supported, we only will need to switch the constructor/initializer.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: I prefer reverting the commit to flags, since we don't know the exact time lines of the switch back to gRPC for metadata APIs

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Since the constructors are public, we would not want to expose an implementation which isn't functional

mprashanthsagar: It does bring up a good point of backward compatibility, Since we do not support gRPC as GA, I would assume, it is ok to break the constructor.

@davidrabinowitz  Please advice

davidrabinowitz: As the grpc only implementation is broken at this time, let's remove it and force the new constructor

davidrabinowitz (changes requested):

davidrabinowitz (approved):

mayanks (approved):
LGTM

PR #577, opened 2021-06-24
Fix resumable upload from offset of last chunk

mprashanthsagar:
* In case of failure in last chunk, the resumable upload offset is equal to `bufferedChunkDataMap.key`, since the size of data is `0`. Adding additional condition to account for this case
   * `entry.getKey() == writeOffset`   ensures that only the chunk with data size `0` is selected but not the chunk before as condition `
entry.getKey() + entry.getValue().size() == writeOffset` is satisfied by both the chunks in the buffer
* Also Increase number of requests to be buffered for resumable upload to 20

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -370,7 +370,7 @@ public class GoogleHadoopFileSystemConfiguration {
       new HadoopConfigurationProperty<>("fs.gs.grpc.write.timeout.ms", 10 * 60 * 1000L);
 
   /** Configuration key for the number of requests to be buffered for uploads to GCS. */
-  public static final HadoopConfigurationProperty<Long> GCS_GRPC_UPLOAD_REQUESTS_BUFFER =
+  public static final HadoopConfigurationProperty<Long> GCS_GRPC_UPLOAD_BUFFERED_REQUESTS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.write.buffered.requests", 20L);
 
   /**
@@ -528,7 +528,7 @@ public class GoogleHadoopFileSystemConfiguration {
             GCS_OUTPUT_STREAM_DIRECT_UPLOAD_ENABLE.get(config, config::getBoolean))
         .setGrpcChecksumsEnabled(GCS_GRPC_CHECKSUMS_ENABLE.get(config, config::getBoolean))
         .setGrpcWriteTimeout(GCS_GRPC_WRITE_TIMEOUT_MS.get(config, config::getLong))
-        .setNumberOfRequestsBuffered(GCS_GRPC_UPLOAD_REQUESTS_BUFFER.get(config, config::getLong))
+        .setNumberOfRequestsBuffered(GCS_GRPC_UPLOAD_BUFFERED_REQUESTS.get(config, config::getLong))
         .build();
   }
 
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -347,39 +347,27 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.grpc.enable", false);
 
-  /**
-   * Configuration key for enabling checksum validation for the gRPC API.
-   */
+  /** Configuration key for enabling checksum validation for the gRPC API. */
   public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_CHECKSUMS_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.grpc.checksums.enable", false);
 
-  /**
-   * Configuration key for the Cloud Storage gRPC server address.
-   */
+  /** Configuration key for the Cloud Storage gRPC server address. */
   public static final HadoopConfigurationProperty<String> GCS_GRPC_SERVER_ADDRESS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.server.address");
 
-  /**
-   * Configuration key for the connect timeout (in millisecond) for gRPC read requests to GCS.
-   */
+  /** Configuration key for the connect timeout (in millisecond) for gRPC read requests to GCS. */
   public static final HadoopConfigurationProperty<Long> GCS_GRPC_READ_TIMEOUT_MS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.read.timeout.ms", 20 * 60 * 1000L);
 
-  /**
-   * Configuration key for the connect timeout (in millisecond) for gRPC metadata requests to GCS.
-   */
+  /** Configuration key for the connect timeout (in millisecond) for gRPC metadata requests to GCS. */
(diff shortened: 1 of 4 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -73,7 +73,7 @@
   // more complex implementation that periodically queries the service to find out the last
   // committed offset, to determine what's safe to discard, but that would also impose a performance
   // penalty.
-  private static final int NUMBER_OF_REQUESTS_TO_RETAIN = 5;
+  private static final int NUMBER_OF_REQUESTS_TO_RETAIN = 20;
```
medb, line 76: This should be configurable via property

davidrabinowitz (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mprashanthsagar: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -364,14 +364,19 @@
   public static final HadoopConfigurationProperty<Long> GCS_GRPC_WRITE_TIMEOUT_MS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.write.timeout.ms", 10 * 60 * 1000L);
 
+  public static final HadoopConfigurationProperty<Long> GCS_GRPC_UPLOAD_REQUESTS_BUFFER =
```
medb, line 367: We need to add a java doc for this and rest gRPC properties.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -364,14 +364,19 @@
...
+      new HadoopConfigurationProperty<>("fs.gs.grpc.write.requests.buffered", 20L);
```
medb, line 368: Maybe `fs.gs.outputstream.grpc.buffered.requests` will be a more consistent name w/ HTTP properties?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: Ack

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: Done

davidrabinowitz (approved):

util/src/main/java/com/google/cloud/hadoop/util/AsyncWriteChannelOptions.java
```diff
@@ -20,45 +20,70 @@
 import com.google.auto.value.AutoValue;
 import com.google.common.flogger.GoogleLogger;
 
-/** Options for the {@link AbstractGoogleAsyncWriteChannel}. */
+/**
```
medb, line 23: Please revert this formatting - it was already formatted w/ https://github.com/google/google-java-format

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -347,20 +347,39 @@
   public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.grpc.enable", false);
 
-  /** Configuration key for enabling checksum validation for the gRPC API. */
+  /**
```
medb, line 350: Please re-format w/ https://github.com/google/google-java-format

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: Code was formatted with `google-java-format` intellij plugin. 

The style guide indicates that the preferred style is multi line block [comments](https://google.github.io/styleguide/javaguide.html#s4.8.6-comments) and in case of [javadoc](https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line), both are acceptable.

Reverting manually, we can verify the preferred format and re-format the files separately

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -76,9 +76,10 @@
           put("fs.gs.glob.algorithm", GlobAlgorithm.CONCURRENT);
           put("fs.gs.grpc.checksums.enable", false);
           put("fs.gs.grpc.enable", false);
-          put("fs.gs.grpc.server.address", null);
           put("fs.gs.grpc.read.timeout.ms", 20 * 60 * 1000L);
           put("fs.gs.grpc.read.metadata.timeout.ms", 60 * 1000L);
+          put("fs.gs.grpc.server.address", null);
```
jaceklaskowski, line 81: Why not use `GCS_GRPC_SERVER_ADDRESS`?

jaceklaskowski (changes requested):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
mprashanthsagar: The `DEFAULT_GCS_GRPC_SERVER_ADDRESS` is retrieved from the client, Is the idea to use `GCS_GRPC_SERVER_ADDRESS` string as the value ?

Since the default [configuration](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L355) is null, it makes sense to maintain the same

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -29,6 +33,7 @@
 import static com.google.cloud.hadoop.util.HadoopCredentialConfiguration.USER_IMPERSONATION_SERVICE_ACCOUNT_SUFFIX;
 import static com.google.cloud.hadoop.util.testing.HadoopConfigurationUtils.getDefaultProperties;
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
```
medb, line 36: We use Truth asserts instead of JUnit asserts in all our tests, please use them for new tests too.

medb (changes requested):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
mprashanthsagar: Done

medb (approved):

PR #587, opened 2021-07-16
Authentication service integration

maen-allaga:
1. Update the AccessTokenProvider interface to allow context aware access token generation.
2. Introduce new flags to enable per-request authentication and disable low level http retry.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -293,7 +301,18 @@ public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage
       }
     }
 
-    this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(storageOptions);
+    this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(this.storageOptions);
+    this.shouldAuthenticationPerRequest = this.storageOptions.getShouldAuthenticatePerRequest();
+
+    if (this.storageOptions.getAccessTokenProvider() != null) {
+      try {
+        this.accessTokenProvider = this.storageOptions.getAccessTokenProvider().getDeclaredConstructor().newInstance();
```
cyxxy, line 309: In Delegation token support, the access token provider comes from the delegation token instead of directly through configuration, so the implementation here won't work for delegation tokens. Could you just inject the access token provider directly into GoogleCloudStorageImpl, if some configuration is set to "true"?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2096,9 +2115,13 @@ public GoogleCloudStorageItemInfo composeObjects(
...
   }
 
+  @VisibleForTesting
   <RequestT extends StorageRequest<?>> RequestT initializeRequest(
       RequestT request, String bucketName) throws IOException {
-    if (storageRequestAuthorizer != null) {
+    if (shouldAuthenticationPerRequest && this.accessTokenProvider != null) {
+      AccessToken accessToken = this.accessTokenProvider.getAccessToken(request);
+      request.setOauthToken(accessToken.getToken());
+    } else if (storageRequestAuthorizer != null) {
```
cyxxy, line 2124: so I think it is also worth mentioning at least in the CONFIGURATION.md file that "shouldAuthenticationPerRequest" will precede storage request authorizer.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -3457,3 +3544,32 @@ public int read() throws IOException {
     };
   }
 }
+
+
+class FaKeAccessTokenProviderWithCounter implements AccessTokenProvider {
```
cyxxy, line 3549: "Fake" instead of "FaKe"

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -3457,3 +3544,32 @@ public int read() throws IOException {
...
+  private static int callsCount = 0;
```
cyxxy, line 3550: This will result in order-dependent  tests.

gcsio/pom.xml
```diff
@@ -54,6 +54,24 @@
...
+    </profile>
+    <profile>
+      <id>hadoop3</id>
+      <dependencies>
+        <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-client-api</artifactId>
+        </dependency>
+      </dependencies>
+    </profile>
```
medb, line 74: GCSIO library can not depend on Hadoop classes, please remove.

gcsio/pom.xml
```diff
@@ -142,27 +160,38 @@
...
+      <artifactId>util-hadoop</artifactId>
+      <version>${hadoop.identifier}-2.2.3-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
```
medb, line 174: It's already defined in parent pom, please remove here and below.

gcsio/pom.xml
```diff
@@ -142,27 +160,38 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigdataoss</groupId>
+      <artifactId>util-hadoop</artifactId>
+      <version>${hadoop.identifier}-2.2.3-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
medb, line 168: Remove this dependency it brings in Hadoop classes.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
maen-allaga: This is fixed now so that `GoogleCloudStorageImpl` keeps track of the `AccessTokenProvider` to provide the AccessToken.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -236,6 +237,10 @@
   public static final HadoopConfigurationProperty<Integer> GCS_HTTP_READ_TIMEOUT =
       new HadoopConfigurationProperty<>("fs.gs.http.read-timeout", 20 * 1000);
 
+  /** Whether to enable low level HTTP retry                                                                                                                                                                                                                                                                                                                                                               */
```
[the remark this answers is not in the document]
maen-allaga: it was closed a column 240 :) fixed it.

gcs/CONFIGURATION.md
```diff
@@ -152,6 +152,19 @@ provider.
...
+*   `fs.gs.should.authenticate.per.request` (default: `false`)
+
+    Whether to generate a new access token for each GCS request.
+    Note: When setting this to true, you have to have a AccessTokenProvider 
+    implementation provided.
+    Note2: Using this flag should accompany disabling `fs.gs.enable.http.retry.initializer`.
+    
+*   `fs.gs.enable.http.retry.initializer` (default: `true`)
+
+    Whether to enable HTTP Retry mechanism.
```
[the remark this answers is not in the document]
maen-allaga: Yeah, this is a gray area, they're technically related. Semantically, they're independent. I will combine them here since this should evolved into "auth.provider.method"

gcs/CONFIGURATION.md
```diff
@@ -152,6 +152,19 @@ provider.
...
     Connector.
+    
+*   `fs.gs.should.authenticate.per.request` (default: `false`)
+
+    Whether to generate a new access token for each GCS request.
+    Note: When setting this to true, you have to have a AccessTokenProvider 
+    implementation provided.
+    Note2: Using this flag should accompany disabling `fs.gs.enable.http.retry.initializer`.
+    
+*   `fs.gs.enable.http.retry.initializer` (default: `true`)
```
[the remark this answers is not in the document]
maen-allaga: makes sense. Anyway according to the below comment, this is going to be merged with the other flag.

gcs/CONFIGURATION.md
```diff
@@ -152,6 +152,19 @@ provider.
 
     The implementation of the AccessTokenProvider interface used for GCS
     Connector.
+    
+*   `fs.gs.should.authenticate.per.request` (default: `false`)
```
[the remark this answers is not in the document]
maen-allaga: Totally agree, the name can use more thought. According to the below comments, this will be merged with the other flag to get something like "auth.token.method"

gcsio/pom.xml
maen-allaga: will do. Thought to make it clear after the `<!-- Test dependencies -->` comment :)

gcsio/pom.xml
maen-allaga: This is interesting. The module definition says `An implementation of org.apache.hadoop.fs.FileSystem targeting Google Cloud Storage`. I understand the separation of concerns, but It's unclear to me why we want to avoid depending on Hadoop classes when we're implementing `org.apache.hadoop.fs.FileSystem`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
maen-allaga: Will do. I will introduce an `AuthenticationMethod` to make it clear what is being used.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
maen-allaga: This class will be removed in the next iteration.

gcsio/pom.xml
medb: GCSIO does not implement `org.apache.hadoop.fs.FileSystem` it implements its own FS-like abstraction. GCSIO is Hadoop-free because it used by frameworks (Beam, etc) that do not depend on the Hadoop - this is a whole point of having `gcs/` and `gcsio/` modules - if both will depend on Hadoop we as well may just merge them together.

medb (changes requested):

gcsio/pom.xml
maen-allaga: Ok, this makes sense now. I will update the module definition (the statement above is copied from the pom.xml definition).

gcsio/pom.xml
maen-allaga: I removed it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
maen-allaga: Actually, on a second thought, I don't think we should make these two mutually exclusive. So will keep both running if they're set. I will make it clear through the CONFIGURATION.md that the access token provider is the preferred way now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -293,7 +321,9 @@ public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage
       }
     }
 
-    this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(storageOptions);
+    this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(this.storageOptions);
+    this.refreshAccessTokenPerRequestEnabled = this.storageOptions.getRefreshAccessTokenPerRequestEnabled();
```
cyxxy, line 325: Don't you just have this boolean value in the Storage Options?

gcs/CONFIGURATION.md
```diff
@@ -153,18 +153,13 @@ provider.
     The implementation of the AccessTokenProvider interface used for GCS
     Connector.
     
-*   `fs.gs.should.authenticate.per.request` (default: `false`)
+*   `fs.gs.refresh.access-token.per.request.enable` (default: `false`)
```
cyxxy, line 156: Please also mention that "when this property is set to true, low level HTTP retries will be disabled".

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
maen-allaga: yeah, I can pull it from the options at anytime. Let me do that change

cyxxy (approved):
LGTM

util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialConfiguration.java
```diff
@@ -60,11 +60,15 @@
       new HadoopConfigurationProperty<>(
           ".auth.service.account.email", /* defaultValue= */ null, ".service.account.auth.email");
 
-  /** Key suffix used to specify private key id for the service account. */
+  /**
```
medb, line 63: Use Google Java Format to format changed lines in this PR: https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
```diff
@@ -103,6 +104,37 @@ public void testConstructorNullCredential() {
     new RetryHttpInitializer(/* credential= */ null, "foo-user-agent");
   }
 
+  @Test
+  public void testInitializeWithInitializerDisabled() throws IOException {
```
medb, line 108: What does this test check?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializerOptions.java
```diff
@@ -42,6 +44,9 @@ public static Builder builder() {
 
   public abstract Builder toBuilder();
 
+  /** Whether the RetryHttpInitializer is enabled or not. */
+  public abstract boolean getIsEnabled();
```
medb, line 48: rename to `isEnabled`

gcs/CONFIGURATION.md
```diff
@@ -248,6 +257,8 @@ When configured, a specified authorization handler will be used to authorize
 Cloud Storage API requests before executing them. The handler will throw
 `AccessDeniedException` for rejected requests if user does not have enough
 permissions (not authorized) to execute these requests.
+Note: Using a `AccessTokenProvider` is the preferred way to perform authentication
```
medb, line 260: Why this is the case?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1287,6 +1290,35 @@ public GoogleCloudStorageFileSystem getGcsFs() {
...
+  /**
+   * Load's an {@link AccessTokenProvider} implementation. If the user provided an AbstractDelegationTokenBinding
+   * we get the AccessTokenProvider, otherwise if a class name is provided
+   * (See {@link HadoopCredentialConfiguration#ACCESS_TOKEN_PROVIDER_IMPL_SUFFIX} then we use it,
+   * otherwise it's null.
+   */
+  private AccessTokenProvider getAccessTokenProvider(
+      Configuration config, GoogleCloudStorageFileSystemOptions gcsFsOptions)
+      throws IOException {
+    AccessTokenProvider atp = null;
```
medb, line 1302: Use full variable name: `accessTokenProvider`

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1287,6 +1290,35 @@ public GoogleCloudStorageFileSystem getGcsFs() {
...
+
+    // Check if delegation token support is configured
+    if (delegationTokens != null) {
```
medb, line 1305: Replace with ternary operator.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -475,11 +480,12 @@
         .setEncryptionKey(RedactedString.create(GCS_ENCRYPTION_KEY.getPassword(config)))
         .setEncryptionKeyHash(RedactedString.create(GCS_ENCRYPTION_KEY_HASH.getPassword(config)))
         .setGrpcEnabled(GCS_GRPC_ENABLE.get(config, config::getBoolean))
-        .setAuthorizationHandlerImplClass(
-            GCS_AUTHORIZATION_HANDLER_IMPL.get(
+        .setAuthorizationHandlerImplClass(GCS_AUTHORIZATION_HANDLER_IMPL.get(
```
medb, line 483: Fix formatting

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -242,13 +247,31 @@ public Boolean load(String bucketName) throws Exception {
    */
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Credential credential)
       throws IOException {
-    this(options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions()));
+    this(options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions()), /*accessTokenProvider*/ null);
```
medb, line 250: Change `/*accessTokenProvider*/` to `/* accessTokenProvider= */`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -257,16 +280,18 @@ public GoogleCloudStorageImpl(
    * @param storage {@link Storage} to use for I/O.
    */
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage) {
-    this(options, storage, null);
+    this(options, storage, /*credentials*/ null, /*accessTokenProvider*/ null);
```
medb, line 283: Add `=` after field names

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2096,8 +2122,13 @@ public GoogleCloudStorageItemInfo composeObjects(
     return compositeInfo;
   }
 
+  @VisibleForTesting
   <RequestT extends StorageRequest<?>> RequestT initializeRequest(
       RequestT request, String bucketName) throws IOException {
+    if (this.storageOptions.getRefreshAccessTokenPerRequestEnabled() && this.accessTokenProvider != null) {
```
medb, line 2128: We usually do not use `this` in methods.

gcs/CONFIGURATION.md
```diff
@@ -152,6 +152,15 @@ provider.
...
     Connector.
+    
+*   `fs.gs.refresh.access-token.per.request.enable` (default: `false`)
+
+    Whether to generate a new access token for each GCS request.
+    Note: When setting this to true, you have to have an AccessTokenProvider 
+    implementation provided. When using `refresh.access-token.per.request.enable`
+    The low level HTTP retry will be turned off.
+    Note: When this flag is set to true, it takes precedency over 
+    `fs.gs.authorization.handler.impl`
```
medb, line 163: Why do we need this option?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -64,6 +64,7 @@
           put("fs.gs.application.name.suffix", "");
           put("fs.gs.authorization.handler.impl", null);
           put("fs.gs.authorization.handler.properties.", ImmutableMap.of());
+          put("fs.gs.refresh.access-token.per.request.enable", false);
```
medb, line 67: Sort alphabetically and rename this property to `fs.gs.refresh.access.token.per.request.enable` for consistency

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -45,10 +45,16 @@
 import static com.google.cloud.hadoop.util.testing.MockHttpTransportHelper.mockTransport;
 import static com.google.common.net.HttpHeaders.CONTENT_LENGTH;
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
```
medb, line 49: We do not use JUnit assets, use Truth instead

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImplCreateTest.java
```diff
@@ -50,7 +50,7 @@ public void create_grpcAndVmComputeEngineCredentials_useDirectpath() throws IOEx
   public void create_grpcAndDisableDirectPathAndVmComputeEngineCredentials_useCloudpath() throws IOException {
     GoogleCloudStorageImpl gcs = new GoogleCloudStorageImpl(GoogleCloudStorageOptions.builder().setAppName("app")
         .setGrpcEnabled(true).setDirectPathPreffered(false).build(), createStorage(),
-        ComputeEngineCredentials.newBuilder().build());
+        ComputeEngineCredentials.newBuilder().build(), null);
```
medb, line 53: We can not change constructor for public classes in subminor release, we can only add new ones and deprecate old ones.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -3387,6 +3397,48 @@ public void testInstantiateAuthorizationHandler() {
...
+  @Test
+  public void initializeRequest_withAccessTokenProviderNotUsingNewTokenPerRequest()
+      throws IOException {
+    MockHttpTransport transport = mockTransport(jsonDataResponse(newBucket(BUCKET_NAME)));
+
+    GoogleCloudStorageOptions gcsOptions = GCS_OPTIONS.toBuilder()
+        .setRefreshAccessTokenPerRequestEnabled(false)
+        .build();
+
+    StorageAccessTokenProvider mockAtp = Mockito.mock(StorageAccessTokenProvider.class);
```
medb, line 3409: Use stub/fake instead, we try to not use Mockito mocks if possible

util-hadoop/pom.xml
```diff
@@ -97,6 +92,16 @@
...
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigdataoss</groupId>
+      <artifactId>util</artifactId>
+      <version>${bigdataoss.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigdataoss</groupId>
+      <artifactId>gcsio</artifactId>
```
medb, line 102: Hadoop utils can not depend on GCSIO.
Why is this necessary?

util-hadoop/src/main/java/com/google/cloud/hadoop/util/AccessTokenProvider.java
```diff
@@ -14,41 +14,39 @@
...
+public interface AccessTokenProvider extends Configurable, StorageAccessTokenProvider {
 
   /** @return an access token. */
   AccessToken getAccessToken();
 
+  /**
+   * @param storageRequest a storage request holding the context.
+   * @return an access token.
+   */
+  default<RequestT extends StorageRequest<?>> AccessToken getAccessToken(RequestT storageRequest) {
```
medb, line 32: This makes interface GCS JSON-API specific, I think that we need to make it GCS API-agnostic.

medb (changes requested):

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
maen-allaga: It's testing if the `initialize` method would exit when the `RetryHTTPInitializer.isEnabled()` turned off.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializerOptions.java
maen-allaga: the `<AutoValue>` requires that all methods either use `set/get` convention or none use. Please let me know if there's a workaround.

gcs/CONFIGURATION.md
maen-allaga: Because using the new interface of `StorageAccessTokenProvider` gives both authentication and authorization (i.e. provides the token and checks permissions).

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
maen-allaga: sure.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
maen-allaga: done.

gcs/CONFIGURATION.md
maen-allaga: Probably this is a naming issue, this option is really about "Should we obtain a token that is related to the request in hand".

maen-allaga: Converted this to an enum type and renamed to `authorization.mode`

util-hadoop/pom.xml
maen-allaga: Because AccessTokenProvider inherits from StorageAccessTokenProvider which exists in GCSIO

util-hadoop/src/main/java/com/google/cloud/hadoop/util/AccessTokenProvider.java
maen-allaga: Introduced a new type and an adapter to make this generic.

gcs/CONFIGURATION.md
```diff
@@ -153,14 +153,13 @@ provider.
...
-    Whether to generate a new access token for each GCS request.
-    Note: When setting this to true, you have to have an AccessTokenProvider 
-    implementation provided. When using `refresh.access-token.per.request.enable`
-    The low level HTTP retry will be turned off.
-    Note: When this flag is set to true, it takes precedency over 
-    `fs.gs.authorization.handler.impl`
+*   `fs.gs.authorization.mode` (default: `GENERIC`)
+
+    The mode to define how an Access Token is Generated. Possible values are:
+    `Generic`: Generates a generic token, i.e. not aware of request object.
```
cyxxy, line 159: Upper case, also, how about the word 'BASIC' - which is already used by the 'fs.gs.outputstream.type' property.

gcs/CONFIGURATION.md
```diff
@@ -153,14 +153,13 @@ provider.
...
+    `REQUEST_CONTEXT_RELATED`: Which consider the request's objects and actions
```
cyxxy, line 160: Can you put these into a sub-list? See the explanation of 'fs.gs.outputstream.type'.

gcs/CONFIGURATION.md
```diff
@@ -153,14 +153,13 @@ provider.
...
+    while generating the token. When using this mode, an AccessTokenProvider 
+    Should be provided using `fs.gs.auth.access.token.provider.impl` config.
```
cyxxy, line 162: Not necessarily true.  Access token provider can also come from a delegation token binding, so maybe explain like "when using this mode, an access token provider should be provided either through the delegation token binding or by directly specifying the `fs.gs.auth.access.token.provider.impl` configuration."

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/authorization/GcsResourceAndActions.java
```diff
@@ -0,0 +1,51 @@
+package com.google.cloud.hadoop.gcsio.authorization;
+
+import java.util.Objects;
+
+public class GcsResourceAndActions {
```
cyxxy, line 5: Looks like the name of class should be "GcsResourceAndAction"? with "action" being singular

gcs/CONFIGURATION.md
maen-allaga: hmm, the way I'm thinking about it is that it's a token that works for any requests... so basically Generic. Basic could work, I'm not sure if that's more descriptive. What do you think?

cyxxy: I was thinking more about the use case, like "basic use case". I don't feel too strong about this though.

gcs/CONFIGURATION.md
maen-allaga: Sure.

gcs/CONFIGURATION.md
maen-allaga: True! updated the description.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/authorization/GcsResourceAndActions.java
maen-allaga: Yep, I missed the class name.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
maen-allaga: understood, I will keep this one since it's very straight forward and I'm looking at `verify` functionality.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1454,14 +1487,22 @@ private GoogleCloudStorageFileSystem createGcsFs(Configuration config) throws IO
     GoogleCloudStorageFileSystemOptions gcsFsOptions =
         GoogleHadoopFileSystemConfiguration.getGcsFsOptionsBuilder(config).build();
 
+    checkArgument(gcsFsOptions.getCloudStorageOptions().getAuthorizationMode()
```
medb, line 1490: Does not look like this formatted with Google Java Format, may you reformat?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -242,13 +249,51 @@ public Boolean load(String bucketName) throws Exception {
    */
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Credential credential)
       throws IOException {
-    this(options, new RetryHttpInitializer(credential, options.toRetryHttpInitializerOptions()));
+    this(
+        options,
+        new RetryHttpInitializer(
+            credential, options.toRetryHttpInitializerOptions()), /*accessTokenProvider*/
```
medb, line 255: nit add `=` to comment:
```
/*accessTokenProvider=*/
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/authorization/AuthorizationMode.java
```diff
@@ -0,0 +1,18 @@
...
+ * The Authorization mode defines how to obtain the Access Token.
+ * GENERIC means the generated Access Token works for any GCS request.
+ * REQUEST_CONTEXT_RELATED means the generated Access Token works for the objects/actions
+ * specified in the request
+ */
+public enum AuthorizationMode {
+  GENERIC("GENERIC"),
+  REQUEST_CONTEXT_RELATED("REQUEST_CONTEXT_RELATED");
+
+  private final String authorizationTypeName;
```
medb, line 13: Why do we need a name?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -3400,8 +3451,17 @@ private GoogleCloudStorage mockedGcs(
...
+
+  private GoogleCloudStorage mockedGcs(
+      GoogleCloudStorageOptions gcsOptions,
+      HttpTransport transport,
+      HttpRequestInitializer requestInitializer,
+      StorageAccessTokenProvider accessTokenProvider) {
     Storage storage = new Storage(transport, JSON_FACTORY, requestInitializer);
-    return new GoogleCloudStorageImpl(gcsOptions, storage);
+    return new GoogleCloudStorageImpl(
+        gcsOptions, storage, /*credentials*/ null, accessTokenProvider);
```
medb, line 3464: nit add `=` to the parameter comment: `credentials=`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
medb: While it's straightforward now, if we do not have stub for `StorageAccessTokenProvider` it means that when we start to work on new functionality we will need to implement it and will end up with mocks again, that's why I think that better to do it right away and use it here.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1298,31 +1328,34 @@ private Credential getCredential(
...
 
-    // Check if delegation token support is configured
-    if (delegationTokens != null) {
-      // If so, use the delegation token to acquire the Google credentials
-      AccessTokenProvider atp = delegationTokens.getAccessTokenProvider();
-      if (atp != null) {
-        atp.setConf(config);
+    // If the AuthorizationMode is set to `Request_Context_Relate`, Credential will be generated
+    // When GCS requests are created.
+    if (gcsFsOptions.getCloudStorageOptions().getAuthorizationMode()
```
medb, line 1333: May we flip this condition and return `null` if it's default.

Also, we need to validate that all enums values are handled

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -1454,14 +1487,22 @@ private GoogleCloudStorageFileSystem createGcsFs(Configuration config) throws IO
     GoogleCloudStorageFileSystemOptions gcsFsOptions =
         GoogleHadoopFileSystemConfiguration.getGcsFsOptionsBuilder(config).build();
 
+    checkArgument(gcsFsOptions.getCloudStorageOptions().getAuthorizationMode()
+            != AuthorizationMode.REQUEST_CONTEXT_RELATED
+            || (HadoopCredentialConfiguration.getNullCredentialEnabled(config, GCS_CONFIG_PREFIX)
+            && HadoopCredentialConfiguration.getServiceAccountEnabled(config, GCS_CONFIG_PREFIX)),
+        "When using AuthorizationMode==REQUEST_CONTEXT_RELATED, `fs.gs.auth.null.enabled` should be true and `fs.gs.auth.service.account.enable` should be false");
```
medb, line 1494: Can we move this condition in the `getAccessTokenProvider` provider, so all `AuthorizationMode` requests are in the same place?

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
maen-allaga: I will add Fake/stub. But I'm even more confused now. If we add more functionality, we will have to update the stub/fake. no? what am I missing?

gcs/CONFIGURATION.md
maen-allaga: fair enough. I will rename to basic as it's more abstract and make it easier to change default implementation if we decided to.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
maen-allaga: yeah, seems like I missed running the script one last time.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
maen-allaga: Of course.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/authorization/AuthorizationMode.java
maen-allaga: Just keeping the door open to evolve this enum when we refactor/consolidate all Authorization codes (i.e. each mode will have certain flags set/turned off. Instead of having flags all over the place). Let me know if you feel strongly I can keep it simple for now and let the refactoring decide what should happen.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
maen-allaga: Sure, I will do both.

maen-allaga: FWIW, I didn't add a condition to throw for `else`. Let me know that should be there too.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -0,0 +1,159 @@
...
+    resourceAndActions.add(
+        AccessBoundary.create(request.getSourceBucket(), request.getSourceObject(), "read"));
+    resourceAndActions.add(
+        AccessBoundary.create(
+            request.getDestinationBucket(), request.getDestinationObject(), "write"));
+    return resourceAndActions;
+  }
+
+  private static List<AccessBoundary> translateObjectCopyRequest(Storage.Objects.Copy request) {
+    List<AccessBoundary> resourceAndActions = new ArrayList<>();
```
medb, line 125: Should all of these be `ImmutableList`?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverterTest.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class StorageRequestToAccessBoundaryConverterTest {
```
medb, line 39: Add java doc

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverterTest.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  @Test
+  public void translateNotRecognizedRequest() throws IOException {
+    StorageObject object = new StorageObject().setName(OBJECT_NAME);
+    Storage.Objects.Update request = storage.objects().update(BUCKET_NAME, OBJECT_NAME, object);
+
+    try {
+      StorageRequestToAccessBoundaryConverter.fromStorageObjectRequest(request);
+      fail();
```
medb, line 266: Use `assertThrows` instead.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverterTest.java
```diff
@@ -0,0 +1,274 @@
...
+ *  Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.gcsio;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
```
medb, line 18: Migrate to Truth asserts.

util-hadoop/src/main/java/com/google/cloud/hadoop/util/CredentialFromAccessTokenProviderClassFactory.java
```diff
@@ -80,22 +79,9 @@ protected TokenResponse executeRefreshToken() throws IOException {
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
 
-  /** Generate the credential from the {@link AccessTokenProvider}. */
-  public static Credential credential(
```
medb, line 84: What these changes are about?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -0,0 +1,159 @@
...
+import com.google.api.services.storage.model.ComposeRequest;
+import com.google.api.services.storage.model.StorageObject;
+import com.google.cloud.hadoop.util.AccessBoundary;
+import com.google.common.base.Strings;
+import java.security.InvalidParameterException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+class StorageRequestToAccessBoundaryConverter {
```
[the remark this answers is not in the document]
maen-allaga: Added

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
maen-allaga: Sure!

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -0,0 +1,159 @@
...
+    String bucketName = ((Bucket) getData(request)).getName();
+    return Collections.singletonList(AccessBoundary.create(bucketName, "/", "write"));
+  }
+
+  private static List<AccessBoundary> translateBucketDeleteRequest(Storage.Buckets.Delete request) {
+    return Collections.singletonList(AccessBoundary.create(request.getBucket(), "/", "write"));
+  }
+
+  private static List<AccessBoundary> translateBucketListRequest(Storage.Buckets.List request) {
+    return Collections.singletonList(AccessBoundary.create("", "", ""));
```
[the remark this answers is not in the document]
maen-allaga: This boundary is used for bucket list, it means "No bucket is specified". I believe the action should be read. @cyxxy to confirm

util-hadoop/src/main/java/com/google/cloud/hadoop/util/CredentialFromAccessTokenProviderClassFactory.java
maen-allaga: Extracting the logic to pull AccessTokenProvider out of this class since it's used in a larger scope.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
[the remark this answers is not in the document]
cyxxy: Right the action should be read.

However, is the buckets.list operation ever used by the connector at all? @medb

medb (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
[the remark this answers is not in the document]
medb: It's not sued in Hadoop GCS connector, but GCSIO has listBuckets method

PR #588, opened 2021-07-22
[DO NOT MERGE] Removed the service account restriction to use directpath

veblush:
Once the GCS ESF will get the new update to lift the default service account on Directpath, this needs to get merged to be aligned with it.


davidrabinowitz:
Directpath is on GCE only. We should check (either here or when the option is set) the connector is run on GCE and not on another setting (on-prem/another cloud)

veblush:
@davidrabinowitz Yeah that's a good point. I added the same check routine of [gax-java](https://github.com/googleapis/gax-java/blob/209b4944feba1c62be2c9de4545e3b01a806b738/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L288).

PR #589, opened 2021-07-25
Evict gRPC channel from pool, for transient errors

mprashanthsagar:
On transient errors, we re-create the Storage stub, which internally uses the same channel pool. We end up with re-using the same channel (as number of active requests may not have altered the priority order in the pool) and error for the next attempt as well.

Evicting the channel from pool, will ensure that the same channel is not re-used in new stubs.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
@@ -37,6 +37,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -148,6 +149,23 @@ class StorageStubProvider {
     public int activeRequests() {
       return counter.ongoingRequestCount.get();
     }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (!(o instanceof ChannelAndRequestCounter)) {
+        return false;
+      }
+      ChannelAndRequestCounter that = (ChannelAndRequestCounter) o;
+      return com.google.common.base.Objects.equal(channel, that.channel);
+    }
+
+    @Override
+    public int hashCode() {
+      return com.google.common.base.Objects.hashCode(channel);
+    }
   }
 
   StorageStubProvider(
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -148,6 +149,23 @@ public ChannelAndRequestCounter(ManagedChannel channel, ActiveRequestCounter cou
     public int activeRequests() {
       return counter.ongoingRequestCount.get();
     }
+
+    @Override
+    public boolean equals(Object o) {
```
medb, line 154: Can we just use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object- ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -148,6 +149,23 @@ public ChannelAndRequestCounter(ManagedChannel channel, ActiveRequestCounter cou
...
+      if (!(o instanceof ChannelAndRequestCounter)) {
+        return false;
+      }
+      ChannelAndRequestCounter that = (ChannelAndRequestCounter) o;
+      return com.google.common.base.Objects.equal(channel, that.channel);
+    }
+
+    @Override
+    public int hashCode() {
+      return com.google.common.base.Objects.hashCode(channel);
```
medb, line 167: Is it better than Java 8's `hashCode`?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
mprashanthsagar: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -159,12 +159,12 @@ public boolean equals(Object o) {
         return false;
       }
       ChannelAndRequestCounter that = (ChannelAndRequestCounter) o;
-      return com.google.common.base.Objects.equal(channel, that.channel);
+      return Objects.equals(channel, that.channel);
```
medb, line 162: Why do we need above checks? Aren't they are covered by `Objects#equals` already?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
mprashanthsagar: `Objects.equals` checks for channel object, but not `ChannelAndRequestCounter`,  we would want to delete the counter if the channel is same, but not necessarily the count, we would have access to channel alone to evict.

medb (approved):

PR #591, opened 2021-07-27
Prevent on using dot in a bucket name when running integration tests

majdyz:
This small PR is removing the occurrence of . ("dot") character in the system username for generating the bucket name when running the integration test. 

To reproduce the error:
Set the `user.name` property with any username that has dot character in it, and run the integration tests.


majdyz:
@googlebot I signed it!

jaceklaskowski:
LGTM

medb (approved):

medb:
Thank you for the fix!

PR #592, opened 2021-07-27
Unshade the "AccessTokenType" class

cyxxy:


medb (approved):

jaceklaskowski:
Why is this change required? I know what the PR does, but not sure I understood why?

medb:
@jaceklaskowski `AccessTokenProvider` is a part of the published interface that third party implementations depend on to extend authentication functionality in GCS connector. That's why `AccessTokenProvider` and all its nested classes should not be shaded, so third party implementation can continue to work with GCS connector shaded jar.

PR #599, opened 2021-08-10
Adding `com.google` debug logging configuration flag 

majdyz:
There could be a common use cases where the user needs to enable the debug log for the connector but has no access to the JVM/cluster that is running it , where attaching the logging configuration can't be done.

This PR is adding a new Hadoop configuration `fs.gs.debug.logging.enable` for enabling the global debug logging (`com.google.*`) which might include:
* `com.google.api.client.http` for HTTP requests logging
* `com.google.cloud.hadoop.gcsio`
* `com.google.cloud.hadoop.util`

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -213,6 +213,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Long> GCS_REWRITE_MAX_BYTES_PER_CALL =
       new HadoopConfigurationProperty<>("fs.gs.rewrite.max.bytes.per.call", 512 * 1024 * 1024L);
 
+  /** Configuration key for enabling gcs debug logging. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_ENABLE_DEBUG_LOGGING =
+      new HadoopConfigurationProperty<>("fs.gs.debug.logging.enable", false);
+
   /** Configuration key for number of items to return per call to the list* GCS RPCs. */
   public static final HadoopConfigurationProperty<Long> GCS_MAX_LIST_ITEMS_PER_CALL =
       new HadoopConfigurationProperty<>("fs.gs.list.max.items.per.call", 1024L);
@@ -445,6 +449,7 @@ public class GoogleHadoopFileSystemConfiguration {
         .setAutoRepairImplicitDirectoriesEnabled(
             GCS_REPAIR_IMPLICIT_DIRECTORIES_ENABLE.get(config, config::getBoolean))
         .setCopyWithRewriteEnabled(GCS_COPY_WITH_REWRITE_ENABLE.get(config, config::getBoolean))
+        .setEnableDebugLogging(GCS_ENABLE_DEBUG_LOGGING.get(config, config::getBoolean))
         .setMaxBytesRewrittenPerCall(GCS_REWRITE_MAX_BYTES_PER_CALL.get(config, config::getLong))
         .setTransportType(
             HTTP_TRANSPORT_SUFFIX.withPrefixes(CONFIG_KEY_PREFIXES).get(config, config::getEnum))
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -101,6 +101,8 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 /**
@@ -239,6 +241,10 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
   // Function that generates downscoped access token.
   private final Function<List<AccessBoundary>, String> downscopedAccessTokenFn;
 
+  // Google root logger, logger that spans over the <package_shading_prefix>.com.google.* coverage.
+  private static final Logger googleLogger = Logger.getLogger(
+      GoogleLogger.class.getName().replaceAll("(com\\.google).*", "$1"));
+
   /**
    * Constructs an instance of GoogleCloudStorageImpl.
    *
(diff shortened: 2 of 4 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -213,6 +213,10 @@
   public static final HadoopConfigurationProperty<Long> GCS_REWRITE_MAX_BYTES_PER_CALL =
       new HadoopConfigurationProperty<>("fs.gs.rewrite.max.bytes.per.call", 512 * 1024 * 1024L);
 
+  /** Configuration key for enabling gcs debug logging. */
```
jaceklaskowski, line 216: nit: Does the description explain anything that the value's name itself does?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -355,6 +361,11 @@ public GoogleCloudStorageImpl(
 
     this.storageRequestAuthorizer = initializeStorageRequestAuthorizer(storageOptions);
     this.downscopedAccessTokenFn = downscopedAccessTokenFn;
+
+    // Set GSC connector logging-level based on the passed debug logging option.
```
jaceklaskowski, line 365: nit: Remove the comment (that serves no purpose really)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -355,6 +361,11 @@ public GoogleCloudStorageImpl(
...
+    if (googleLogger.getLevel() == null) {
```
jaceklaskowski, line 366: What if the level is non-`null`? I thought the property should override it

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -91,6 +91,8 @@
   public static final Map<String, String> AUTHORIZATION_HANDLER_PROPERTIES_DEFAULT =
       ImmutableMap.of();
 
+  public static final boolean ENABLE_DEBUG_LOGGING_DEFAULT = false;
```
jaceklaskowski, line 94: Looks weird as it seems a duplicate of the default value of the `GCS_ENABLE_DEBUG_LOGGING` above?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -114,6 +116,7 @@ public static Builder builder() {
         .setWriteChannelOptions(AsyncWriteChannelOptions.DEFAULT)
         .setRequesterPaysOptions(RequesterPaysOptions.DEFAULT)
         .setCooperativeLockingOptions(CooperativeLockingOptions.DEFAULT)
+        .setEnableDebugLogging(ENABLE_DEBUG_LOGGING_DEFAULT)
```
jaceklaskowski, line 119: Why is this required? Not set could be simply "disabled".

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
majdyz: Added the logged package info in the comment.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
majdyz: Removed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
majdyz: The intention is to only configure this value on initialization, If the value is not null then it's likely the user is manually overriding the level which it's better to not force it back.

This is also a check to avoid repeatedly calling a synchronized `setLevel` method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
majdyz: Removed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
majdyz: Removed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
majdyz: Edit: I re-wrote this to only override when the level value doen't match with the config value

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
jaceklaskowski: OK. I'm not very familiar with the code style and could be wrong. Just a general observation.

jaceklaskowski (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImplCreateTest.java
```diff
@@ -15,13 +15,17 @@
 package com.google.cloud.hadoop.gcsio;
 
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
medb, line 19: We use Truth asserts, please switch to them from JUnit asserts.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -241,6 +243,11 @@ public Boolean load(String bucketName) throws Exception {
   // Function that generates downscoped access token.
   private final Function<List<AccessBoundary>, String> downscopedAccessTokenFn;
 
+  // Google root logger, logger that spans over the <package_shading_prefix>.com.google.* coverage.
+  @VisibleForTesting
+  protected static final Logger googleLogger =
+      Logger.getLogger(GoogleLogger.class.getName().replaceAll("(com\\.google).*", "$1"));
+
```
medb, line 250: I wonder if we actually need to do this, I think it will be cleaner to implement utility class that has a method similar to this: https://github.com/google/flogger/blob/flogger-0.5.1/api/src/main/java/com/google/common/flogger/LoggerConfig.java#L129-L138

medb (changes requested):

PR #603, opened 2021-08-16
Add new maven profile for vpc tests

mprashanthsagar:
Adding a new maven profile for VPC-SC tests temporarily, ignores gRPC integration tests till VPC-SC tests are supported over gRPC

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,7 +251,7 @@
                 <include>**/*IntegrationTest.java</include>
               </includes>
               <excludes>
-                <exclude>**/GoogleCloudStorageGrpcIntegrationTest.java</exclude>
+                <exclude>**/*GrpcIntegrationTest.java</exclude>
               </excludes>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -226,6 +226,43 @@
...
+                  <goal>verify</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <includes>
+                <include>**/*IntegrationTest.java</include>
+              </includes>
+              <excludes>
+                <exclude>**/*GrpcIntegrationTest.java</exclude>
```
medb, line 254: Can we just exclude these tests in Maven CLI instead, because this is a temporary change?

Something like this:
```shell
mvn ... -Dit.test="!**/*GrpcIntegrationTest.java"
```

medb (changes requested):

PR #606, opened 2021-08-20
Short-circuiting in readObjectContentFromGCS

veblush:
`readObjectContentFromGCS` has a following while-block;

```
  while (moreServerContent() &&  byteBuffer.hasRemaining()) {
  ...
  }
```

Meaning it will receive data while there are more data to read and a target buffer is not filled yet. This condition needs to be adjusted a bit like

```
  while (byteBuffer.hasRemaining() && moreServerContent()) {
  ...
  }
```

because the original condition will evaluate `moreServerContent()` even when the target buffer is already full. `moreServerContent` can block until the actuall data is transferred and this would introduce unnecessary waiting-time when byteBuffer is already full. (This can be a problem for random-access case)

So let's check byteBuffer.hasRemaining() first (which is super fast) and evaluate `moreServerContent` which isn't that super fast.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -438,7 +438,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
 
   private int readObjectContentFromGCS(ByteBuffer byteBuffer) throws IOException {
     int bytesRead = 0;
-    while (moreServerContent() && byteBuffer.hasRemaining()) {
+    while (byteBuffer.hasRemaining() && moreServerContent()) {
       ReadObjectResponse res = resIterator.next();
 
       // When zero-copy mashaller is used, the stream that backs GetObjectMediaResponse


veblush:
A brief benchmark shows this can improve the random-access read latency more than 10%.

#### Benchmark

Using [gRPC Java GCS benchmark](https://github.com/GoogleCloudPlatform/grpc-gcp-java/tree/master/end2end-test-examples/gcs) running on Debian 10 using OpenJDK Java 1.8.0_265-b01

```
$ java -Xmx4G -jar $JAR --bkt=gcs-grpc-team-veblush1 --obj=1G.txt --client=gcsio-grpc \
  --method=random --dp=true --call=1000 --size 1048576 --buffSize 10
```

#### Without this PR

```
****** Test Results [client: gcsio-grpc, method: random, size: 1048576, threads: 1, dp: true, calls: 1000]: 
                Min     p5      p10     p25     p50     p75     p90     p99     Max     Total
  Time(ms)      17      53      55      59      64      70      76      103     204     68678
```

#### With this PR

```
***** Test Results [client: gcsio-grpc, method: random, size: 1048576, threads: 1, dp: true, calls: 1000]: 
                Min     p5      p10     p25     p50     p75     p90     p99     Max     Total
  Time(ms)      0       43      45      48      53      59      68      143     218     59259
```

mprashanthsagar (approved):

mprashanthsagar:
@veblush Please raise the pull request to master, will cherry pick it to branch-2.2.x separately

veblush:
@mprashanthsagar Thanks for the review and rebase is done.

PR #620, opened 2021-09-20
Format all java files, remove incremental format check

mprashanthsagar:


pom.xml
```diff
@@ -637,7 +637,6 @@
         <artifactId>spotless-maven-plugin</artifactId>
         <version>2.12.2</version>
         <configuration>
-          <ratchetFrom>branch-2.2.x</ratchetFrom>
```
medb, line 640: Why to remove this?

mprashanthsagar: Now that we format all packages, we do not need to find any diffs to check format

This simplifies our presubmits across branches

medb (approved):

PR #622, opened 2021-09-22
Add precondition to ensure that rename operation does not override destination objects

mprashanthsagar:
`rename` is achieved in two steps 
* cp src dst
* rm src

To ensure atomicity for `rename` op, adding pre-condition on `dst` obj for `cp` operation.

Failure scenarios :

1. Precondition/Any failure for `cp`, we throw exception maintaining atomicity
2. Any failure for `rm`, we attempt to delete `dst` to clean-up.


medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
```diff
@@ -120,6 +120,19 @@ public void copy(
     delegate.copy(srcBucketName, srcObjectNames, dstBucketName, dstObjectNames);
   }
 
+  @Override
+  public void copy(
+      String srcBucketName,
+      String dstBucketName,
```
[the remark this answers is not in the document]
mprashanthsagar: makes sense, will extract bucketName from `StorageResourceId`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -193,11 +195,44 @@ SeekableByteChannel open(StorageResourceId resourceId, GoogleCloudStorageReadOpt
...
    */
-  void copy(
+  default void copy(
       String srcBucketName,
       List<String> srcObjectNames,
       String dstBucketName,
       List<String> dstObjectNames)
+      throws IOException {
+    Preconditions.checkArgument(srcObjectNames != null, "srcObjectNames must not be null");
+    Preconditions.checkArgument(dstObjectNames != null, "dstObjectNames must not be null");
```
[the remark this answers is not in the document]
mprashanthsagar: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -193,11 +196,38 @@ SeekableByteChannel open(StorageResourceId resourceId, GoogleCloudStorageReadOpt
...
+   * is reachable by two different names. Copying between two different locations or between two
+   * different storage classes is not allowed.
+   *
+   * @param srcObjects list of the objects to copy
+   * @param dstObjects list of the objects after copy
+   * @throws java.io.FileNotFoundException if the source object or the destination bucket does not
+   *     exist
+   * @throws IOException in all other error cases
+   */
+  void copy(List<StorageResourceId> srcObjects, List<StorageResourceId> dstObjects)
```
medb, line 230: We need default implementation for this method to preserve backward compatibility.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -630,21 +630,38 @@ public void rename(URI src, URI dst) throws IOException {
...
-
-      // TODO(b/110833109): populate generation ID in StorageResourceId when getting info
-      gcs.deleteObjects(
+          ImmutableList.of(srcResourceId),
           ImmutableList.of(
               new StorageResourceId(
-                  srcInfo.getItemInfo().getBucketName(),
-                  srcInfo.getItemInfo().getObjectName(),
-                  srcInfo.getItemInfo().getContentGeneration())));
+                  dstResourceId.getBucketName(),
```
medb, line 636: What `generationId` has `generationId` object? Can we set it to desired value when we create this object upstream instead?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -630,21 +630,38 @@ public void rename(URI src, URI dst) throws IOException {
...
+
+      try {
+        gcs.deleteObjects(
+            ImmutableList.of(
+                new StorageResourceId(
+                    srcInfo.getItemInfo().getBucketName(),
+                    srcInfo.getItemInfo().getObjectName(),
+                    srcInfo.getItemInfo().getContentGeneration())));
+      } catch (IOException e) {
+        deleteSilently(dstResourceId);
```
medb, line 648: Why do we need this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -630,21 +630,38 @@ public void rename(URI src, URI dst) throws IOException {
...
     }
 
     repairImplicitDirectory(srcParentInfoFuture);
   }
 
+  private void deleteSilently(StorageResourceId destinationObject) throws IOException {
+    try {
+      gcs.deleteObjects(ImmutableList.of(destinationObject));
+    } catch (IOException ioException) {
+      logger.atFiner().log(
```
medb, line 660: Log exception using `.withCause` method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,21 +966,32 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
+    if (srcObjects.isEmpty()) {
+      return;
+    }
+
+    String srcBucketName = srcObjects.get(0).getBucketName();
+    String dstBucketName = dstObjects.get(0).getBucketName();
+
+    for (int i = 0; i < srcObjects.size(); i++) {
+      if (!srcBucketName.equals(srcObjects.get(i).getBucketName()))
+        throw new UnsupportedOperationException(
```
medb, line 989: Probably better to throw `IllegalArgumentException`?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
mprashanthsagar: In case of failure in deleting the `src` file, we would have `src` and `dst` both with same data. This attempts to delete the `dst` so that a retry would be possible

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
mprashanthsagar: We only receive `URI` from upstream, setting `generationId=0` would ensure that the dst file does not exist.

Could you elaborate? may be i don't follow the comment

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
Deependra-Patel: Here we are deleting `dst` without being sure if it was indeed created by us as part of rename.

Eg. Client 1 : Rename (`src`, `dst`), Client 2 (running in parallel in different application) upload `dst`, then get
Now let's consider this order of execution
Client 1: copies `src` to `dst`
Client 2: uploads new `dst`, get's success in response
Client 1: Tries deleting `src` but fails, then deletes `dst`
Client 2: gets `dst`, gets 404 (this was unexpected)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,21 +966,32 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
+            "This operation is not supported across multiple source locations");
```
Deependra-Patel, line 990: Nit: Should we say buckets just to avoid confusion with region?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
```diff
@@ -1027,8 +1027,13 @@ public void rename_file_sequential() throws Exception {
...
             copyRequestString(
-                bucketName, dirObject + "/f1", bucketName, dirObject + "/f2", "copyTo"),
-            deleteRequestString(bucketName, dirObject + "/f1", 1));
+                bucketName,
+                dirObject + "/f1",
+                bucketName,
+                dirObject + "/f2",
+                "copyTo",
+                /* generationId= */ 1),
+            deleteRequestString(bucketName, dirObject + "/f1", 2));
```
Deependra-Patel, line 1036: Why is the generation id changed to 2 for source?
We didn't update /f1, is it because we ran a copy operation on it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -630,21 +630,38 @@ public void rename(URI src, URI dst) throws IOException {
...
+                  /* generationId= */ 0)));
+
+      try {
+        gcs.deleteObjects(
+            ImmutableList.of(
+                new StorageResourceId(
+                    srcInfo.getItemInfo().getBucketName(),
+                    srcInfo.getItemInfo().getObjectName(),
+                    srcInfo.getItemInfo().getContentGeneration())));
+      } catch (IOException e) {
```
Deependra-Patel, line 647: We should add unit test for this case when src deletion fails.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
Deependra-Patel: If Client 2 doesn't use conditional write (with generation id = 0) which is usually the case, then it should overwrite the `dst` and succeed.

Is the first content generation number random? If it is 1, then we could just use that to make sure that this is indeed the file we created.

In line #646 above, you delete `src` only if it is not updated, I am giving the same argument for `dst`.

mprashanthsagar: I concur if `dst` is overwritten, rather than deleting the wrong version of file, it would be better to just fail
i.e, removing `deleteSilently` 

@medb @Deependra-Patel thoughts ?

Deependra-Patel: I am of opinion we shouldn't delete something we aren't sure we created.

Correct me if I understand correctly, earlier behavior before this commit was that the rename would just rewrite the `dst` if it existed and then delete the `src`
We implement `renameInternal` method of `AbstractFileSystem` which has documentation ["and NO OVERWRITE is performed"](https://hadoop.apache.org/docs/r2.7.0/api/org/apache/hadoop/fs/AbstractFileSystem.html#renameInternal(org.apache.hadoop.fs.Path,%20org.apache.hadoop.fs.Path)) , so we are enforcing that condition with this commit.

This will keep `rename` still to be non-atomic which you are trying to get. The solution for that would be to also store generationId of the `dst` objects that you get in copy response (just checked the api). And then conditionally delete (using these ids) `dst`  in case deletion of `src` fails.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
```diff
@@ -1027,8 +1027,13 @@ public void rename_file_sequential() throws Exception {
                 bucketName, dirObject + "/", /* maxResults= */ 1, /* pageToken= */ null),
```
Deependra-Patel, line 1027: Should we have an integration test where `dst` already exists and this precondition would fail, making rename to now fail?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: I think `UnsupportedOperationException` suites better and is consistent with rest of the validations

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
mprashanthsagar: We mock increment genId, so any new check on genId, we would need to modify (increment) the rest in the assertion

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ForwardingGoogleCloudStorage.java
```diff
@@ -120,6 +120,13 @@ public void copy(
     delegate.copy(srcBucketName, srcObjectNames, dstBucketName, dstObjectNames);
   }
 
+  @Override
+  public void copy(List<StorageResourceId> srcObjects, List<StorageResourceId> dstObjects)
```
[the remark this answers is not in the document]
mprashanthsagar: I prefer `Objects` notation, for easier usage

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -200,6 +203,43 @@ void copy(
...
+    List<String> srcObjectNames =
+        srcObjects.stream().map(StorageResourceId::getObjectName).collect(Collectors.toList());
+    List<String> dstObjectNames =
+        dstObjects.stream().map(StorageResourceId::getObjectName).collect(Collectors.toList());
+    String srcBucketName = srcObjects.get(0).getBucketName();
+    String dstBucketName = dstObjects.get(0).getBucketName();
+
+    for (int i = 0; i < srcObjects.size(); i++) {
+      if (!srcBucketName.equals(srcObjects.get(i).getBucketName()))
+        throw new UnsupportedOperationException(
```
medb, line 231: Hm, after further thinking, if GCS does not have this limitation, why should we impose it artificially?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -200,6 +203,43 @@ void copy(
...
+   * is reachable by two different names. Copying between two different locations or between two
+   * different storage classes is not allowed.
+   *
+   * @param srcObjects list of the objects to copy
+   * @param dstObjects list of the objects after copy
+   * @throws java.io.FileNotFoundException if the source object or the destination bucket does not
+   *     exist
+   * @throws IOException in all other error cases
+   */
+  default void copy(List<StorageResourceId> srcObjects, List<StorageResourceId> dstObjects)
```
medb, line 217: Instead of 2 disjoin lists as a parameters, should we just take a Map to be less error prone and reflect intent better? Or list of `Map.Entry` objects?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
medb: Is `generationId=0` a default value for `StorageResourceId` objects?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
mprashanthsagar: For the default Impl, I think we should impose this check, to ensure that compatibility with existing `copy` method.

If we are ok, breaking compatibility, we can remove default impl, and the check as well

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
mprashanthsagar: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -200,6 +206,46 @@ void copy(
...
+  default void copy(Map<StorageResourceId, StorageResourceId> sourceToDestinationObjectsMap)
+      throws IOException {
+    checkArgument(
+        sourceToDestinationObjectsMap != null, "sourceToDestinationObjectsMap must not be null");
+    if (sourceToDestinationObjectsMap.isEmpty()) {
+      return;
+    }
+
+    List<String> srcObjectNames = new ArrayList<>();
+    List<String> dstObjectNames = new ArrayList<>();
```
medb, line 228: nit: initialize list size base on map size to optimize memory allocation a little

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -629,11 +630,15 @@ public void rename(URI src, URI dst) throws IOException {
...
           StorageResourceId.fromUriPath(dst, /* allowEmptyObjectName= */ true);
 
-      gcs.copy(
-          srcResourceId.getBucketName(), ImmutableList.of(srcResourceId.getObjectName()),
-          dstResourceId.getBucketName(), ImmutableList.of(dstResourceId.getObjectName()));
+      StorageResourceId destinationObject =
+          new StorageResourceId(
+              dstResourceId.getBucketName(),
+              dstResourceId.getObjectName(),
+              /* generationId= */ 0); // set generationId = 0 to ensure no live version on gcs
```
medb, line 637: We create 2 `StorageResourceId`s (`dstResourceId` and `destinationObject`) for destination object here, can we merge it into one?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
```diff
@@ -629,11 +630,15 @@ public void rename(URI src, URI dst) throws IOException {
...
+      Map<StorageResourceId, StorageResourceId> srcToDstMap = new HashMap<>();
+      srcToDstMap.put(srcResourceId, destinationObject);
```
medb, line 639: Use `ImmutableMap.of(srcResourceId, destinationObject)` instead.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,21 +986,32 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
   // Please check that removing it is correct, and remove this comment along with it.
   // @VisibleForTesting
   public static void validateCopyArguments(
-      String srcBucketName,
-      List<String> srcObjectNames,
-      String dstBucketName,
-      List<String> dstObjectNames,
+      List<StorageResourceId> srcObjects,
+      List<StorageResourceId> dstObjects,
```
medb, line 990: Use map as an argument here too?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,21 +986,32 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
+        "Must supply same number of elements in srcObjects and dstObjects");
+
+    if (srcObjects.isEmpty()) {
+      return;
+    }
+
+    String srcBucketName = srcObjects.get(0).getBucketName();
+    String dstBucketName = dstObjects.get(0).getBucketName();
+
+    for (int i = 0; i < srcObjects.size(); i++) {
```
medb, line 1007: While this check makes sense in default implementation in the interface, it maybe too strict here?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -1032,9 +1067,34 @@ public void copy(
...
       List<String> dstObjectNames)
       throws IOException {
-    validateCopyArguments(srcBucketName, srcObjectNames, dstBucketName, dstObjectNames, this);
+    checkArgument(srcObjectNames != null, "srcObjectNames must not be null");
+    checkArgument(dstObjectNames != null, "dstObjectNames must not be null");
+    checkArgument(
+        srcObjectNames.size() == dstObjectNames.size(),
+        "Must supply same number of elements in srcObjects and dstObjects");
+
+    Map<StorageResourceId, StorageResourceId> sourceToDestinationObjectsMap = new HashMap<>();
```
medb, line 1076: Initialize Map with size, or even better - use streams.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystem.java
medb: bump

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,63 +990,82 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
 
-      if (!gcsImpl.getOptions().isCopyWithRewriteEnabled()) {
-        if (!srcBucketInfo.getLocation().equals(dstBucketInfo.getLocation())) {
-          throw new UnsupportedOperationException(
-              "This operation is not supported across two different storage locations.");
+    if (sourceToDestinationObjectsMap.isEmpty()) {
+      return;
+    }
+
+    Map<StorageResourceId, GoogleCloudStorageItemInfo> bucketInfoCache =
```
medb, line 1003: Why do we need this cache?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,63 +990,82 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
         throw new IllegalArgumentException(
             String.format(
                 "Copy destination must be different from source for %s.",
-                StringPaths.fromComponents(srcBucketName, srcObjectNames.get(i))));
+                StringPaths.fromComponents(srcBucketName, source.getObjectName())));
       }
     }
   }
 
+  private static GoogleCloudStorageItemInfo getGoogleCloudStorageItemInfo(
```
medb, line 1054: You can replace this whole method with `Map.computeIfAbsent` method.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: In case of copy of multiple objects within the same bucket, we would end up calling `getItemInfo` of the bucket multiple times, this cache avoids multiple network calls

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: `getItemInfo` throws exception, `Map.computeIfAbsent` accepts a lambda expression, which does not allow an unhandled exception, any suggestions ?

medb: I think you can catch an exception inside the lambda, ideally we would use `TunnelException`, but it is not open source yet: https://github.com/google/guava/issues/2828#issuecomment-304187823

medb: Discussed offline, `RuntimeException` will not work here, but we still can try to make this method a little more optima and concise:

```java
    GoogleCloudStorageItemInfo storageItemInfo = bucketInfoCache.get(resourceId);
    if (storageItemInfo != null) {
      return storageItemInfo;
    }
    storageItemInfo = gcsImpl.getItemInfo(resourceId);
    bucketInfoCache.put(resourceId, storageItemInfo);
    return storageItemInfo;
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -964,63 +990,82 @@ public void onFailure(GoogleJsonError jsonError, HttpHeaders responseHeaders) {
...
-      if (!gcsImpl.getOptions().isCopyWithRewriteEnabled()) {
-        if (!srcBucketInfo.getLocation().equals(dstBucketInfo.getLocation())) {
-          throw new UnsupportedOperationException(
-              "This operation is not supported across two different storage locations.");
+    if (sourceToDestinationObjectsMap.isEmpty()) {
+      return;
+    }
+
+    Map<StorageResourceId, GoogleCloudStorageItemInfo> bucketInfoCache =
+        new ConcurrentHashMap<>(sourceToDestinationObjectsMap.size());
```
medb, line 1004: Seems like this do not need to be a `ConcurrentHashMap` and `sourceToDestinationObjectsMap.size()` heuristic could be too aggressive if there are a lot of objects and few buckets, so we can skip it here.

medb (approved):

PR #628, opened 2021-09-30
Reduced Hadoop version to ensure 2.8 compatibility

davidrabinowitz:


pom.xml
```diff
@@ -101,7 +101,7 @@
     <google.oauth-client.version>1.31.5</google.oauth-client.version>
     <google.protobuf.version>3.17.3</google.protobuf.version>
     <grpc.version>1.39.0</grpc.version>
-    <hadoop.two.version>2.10.1</hadoop.two.version>
+    <hadoop.two.version>2.8.0</hadoop.two.version>
```
medb, line 104: May we use 2.7.7 instead - it still widely used?

medb (changes requested):

bigquery/pom.xml
```diff
@@ -286,6 +292,7 @@
                   <include>io.grpc</include>
                   <include>io.opencensus</include>
                   <include>io.perfmark</include>
+                  <include>org.apache.avro</include>
```
medb, line 295: Will this include avro in shaded jar? If so, we can not do this.

medb (changes requested):

mprashanthsagar:
@davidrabinowitz Is this PR still relevant ?

davidrabinowitz:
Yes, but not urgent. It is to keep us supporting old Hadoop versions.

medb:
Seems like it will be better to keep status quo than try to fix this, taking into account that this will be a maintenance branch anyway.

PR #643, opened 2021-10-28
Switch gRPC lb policy to round_robin

veblush:
Switched gRPC LoadBalance Policy to `round_robin` from `pick_first`. For this change, a channel pool should be replaced with a simple gRPC channel because a round-robin channel has a built-in channel management inside (having many subchannels) so just having a single gRPC channel should be enough. Due to no need to maintain a pool, channel eviction and active call counting on every channel can be removed altogether.


mprashanthsagar:
Could you please fix the tests ?

veblush:
@mprashanthsagar Could you put the output of the test? It doesn't seem to allow me to look at it.

mprashanthsagar (approved):

PR #644, opened 2021-10-29
Update dependencies to latest version

### LINKED ISSUE #647: GoogleCloudStorageReadChannel.cacheFooter causes infinite loop when used with google-http-client 1.40
The newer http client uses a BufferedInputStream for the response which will return 0 if called to read 0 bytes of data.
(Apparently the implementation used before returned -1 in that case).
This mean the loop will never terminate given the condition
while (bytesRead >= 0 && totalBytesRead <= footerSize)

mprashanthsagar:
Fixes #647

Deletes two testing, which use `MockHttpTransport` . The updated HttpClient uses `BufferedInputStream` instead of `InputStream`, which breaks the test as the mocked `read` api ends up filling the buffer in `BufferedInputStream` instead of the called buffer. 

`MockHttpTransport` currently does not support setting `rawInputStream` option, which would enforce using `InputStream`

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -811,7 +811,7 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
       do {
         totalBytesRead += bytesRead;
         bytesRead = footerStream.read(footerContent, totalBytesRead, footerSize - totalBytesRead);
-      } while (bytesRead >= 0 && totalBytesRead <= footerSize);
+      } while (bytesRead >= 0 && totalBytesRead < footerSize);
       checkState(
           footerStream.read() < 0,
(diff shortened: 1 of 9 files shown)


medb (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -811,7 +811,7 @@ private void cacheFooter(HttpResponse response) throws IOException {
       do {
         totalBytesRead += bytesRead;
         bytesRead = footerStream.read(footerContent, totalBytesRead, footerSize - totalBytesRead);
-      } while (bytesRead >= 0 && totalBytesRead <= footerSize);
+      } while (bytesRead >= 0 && totalBytesRead < footerSize);
```
mukund-thakur, line 814: Why was this change done? Could someone explain more @medb  @mprashanthsagar Thanks

mukund-thakur: We are seeing some infinite loop in our system in older versions while caching the footers.

PR #648, opened 2021-11-11
Making GoogleCloudStorage implement Closeable

davidrabinowitz:


mprashanthsagar:
Looks like presubmit is failing. could you please take a look 

Also, please add description as to why this change is needed

medb:
This looks like something that should be handled in `@After` method in the test.

PR #651, opened 2021-11-16
Move retries to application level

mprashanthsagar:
Adds retries at application level

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -705,9 +705,8 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
    * Waits until more data is available from the server, or returns false if read is done.
    *
    * @return true if more data is available with .next()
-   * @throws IOException of the appropriate type if there was an I/O error.
    */
-  private boolean moreServerContent() throws IOException {
+  private boolean moreServerContent() {
     if (resIterator == null || requestContext == null || requestContext.isCancelled()) {
       return false;
     }
@@ -719,7 +718,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
       return moreDataAvailable;
     } catch (StatusRuntimeException e) {
       cancelCurrentRequest();
-      throw convertError(e, resourceId);
+      throw e;
     }
   }
 
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -671,8 +671,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
 
     requestContext = Context.current().withCancellation();
     Context toReattach = requestContext.attach();
-    StorageBlockingStub blockingStub =
-        stub.withDeadlineAfter(readTimeout, MILLISECONDS);
+    StorageBlockingStub blockingStub = stub.withDeadlineAfter(readTimeout, MILLISECONDS);
     try {
       if (useZeroCopyMarshaller) {
         resIterator =
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -519,6 +519,14 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     }
   }
 
+  /**
+   * Reads data from GCS over network, with retries
+   *
+   * @param byteBuffer Buffer to be filled with data from GCS
+   * @return number of bytes read into the buffer
+   * @throws IOException In case of data errors or network errors
+   * @throws InterruptedException In case of thread interrupt while retrying
+   */
   private int readFromGCS(ByteBuffer byteBuffer) throws IOException, InterruptedException {
     int read = 0;
     StatusRuntimeException statusRuntimeException;
@@ -710,16 +718,11 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     if (resIterator == null || requestContext == null || requestContext.isCancelled()) {
       return false;
     }
-    try {
-      boolean moreDataAvailable = resIterator.hasNext();
-      if (!moreDataAvailable) {
-        cancelCurrentRequest();
-      }
-      return moreDataAvailable;
-    } catch (StatusRuntimeException e) {
+    boolean moreDataAvailable = resIterator.hasNext();
+    if (!moreDataAvailable) {
       cancelCurrentRequest();
-      throw e;
     }
+    return moreDataAvailable;
   }
 
   @Override
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -574,7 +574,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     try {
       ByteString content = res.getChecksummedData().getContent();
       if (bytesToSkipBeforeReading >= 0 && bytesToSkipBeforeReading < content.size()) {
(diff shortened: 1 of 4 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -462,61 +463,82 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
-      bytesRead += bytesToWrite;
+      return bytesRead;
+    } catch (InterruptedException e) {
+      cancelCurrentRequest();
+      throw new IOException(e);
     }
+  }
 
-    return bytesRead;
+  private int readFromGCS(ByteBuffer byteBuffer) throws IOException, InterruptedException {
```
mayanks, line 516: Let's please add comment here so the intent of making a separate function is clear
- the lower level exception that it internally handles
- exponential sleep in case of retries
- which exceptions and when are they raised
- return value of this function

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -462,61 +463,82 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+          if (bytesToRead.isPresent()) {
+            contentChannelEndOffset = positionInGrpcStream + bytesToRead.getAsLong();
+          }
+        }
+        while (byteBuffer.hasRemaining() && moreServerContent()) {
+          read += readObjectContentFromGCS(byteBuffer);
+        }
+        return read;
+      } catch (StatusRuntimeException e) {
+        cancelCurrentRequest();
```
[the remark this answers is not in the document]
mprashanthsagar: Yeah. I concur with multiple calls to `cancelCurrentRequest` . Will remove it from `moreServerContent

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -462,61 +463,82 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+        statusRuntimeException = e;
+      }
+    } while (ResilientOperation.nextSleep(backoff, sleeper, statusRuntimeException));
```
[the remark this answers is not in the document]
mprashanthsagar: The log is part of the nextSleep method. It is logged at info

util/src/main/java/com/google/cloud/hadoop/util/ResilientOperation.java
```diff
@@ -113,7 +113,7 @@
    * @param currentException exception that caused the retry and sleep. For logging.
    * @throws InterruptedException if sleep is interrupted
    */
-  private static boolean nextSleep(BackOff backoff, Sleeper sleeper, Exception currentException)
+  public static boolean nextSleep(BackOff backoff, Sleeper sleeper, Exception currentException)
```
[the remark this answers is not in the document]
mprashanthsagar: No. It makes sense to be private. Ideally we want to utilise the retry method, unfortunately that ties with a callable without the ability to continue some context (track number of bytes read before we retry in this case)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -693,32 +699,20 @@ private void cancelCurrentRequest() {
    * Waits until more data is available from the server, or returns false if read is done.
    *
    * @return true if more data is available with .next()
-   * @throws IOException of the appropriate type if there was an I/O error.
    */
-  private boolean moreServerContent() throws IOException {
+  private boolean moreServerContent() {
```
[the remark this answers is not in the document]
mprashanthsagar: Yeah, but there are no checked exceptions now

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -462,61 +463,82 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+      return bytesRead;
+    } catch (InterruptedException e) {
+      cancelCurrentRequest();
+      throw new IOException(e);
     }
+  }
 
-    return bytesRead;
+  private int readFromGCS(ByteBuffer byteBuffer) throws IOException, InterruptedException {
+    int read = 0;
```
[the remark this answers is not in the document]
mprashanthsagar: All ops happen to read in terms of Bytest. The current naming makes sense to me to call out a network call instead of in-memory op

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Yeah, we can add some javadoc. I don't think we want to document low level details even for a private method.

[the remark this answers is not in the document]
mprashanthsagar: Footer cache happens to be a ByteString. And retries are simpler there because of context carry over. 

Adding zero-copy deserializer and unifyication of logic *needs* to be done. I did not want to include it in the current PR to avoid complications unless there is a bug in the current impl

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -536,48 +565,46 @@ private boolean isByteBufferBeyondCurrentRequestRange(ByteBuffer byteBuffer) {
...
-          continue;
-        }
+    // When zero-copy marshaller is used, the stream that backs GetObjectMediaResponse
+    // should be closed when the message is no longed needed so that all buffers in the
+    // stream can be reclaimed. If zero-copy is not used, stream will be null.
+    InputStream stream = getObjectMediaResponseMarshaller.popStream(res);
+    try {
+      ByteString content = res.getChecksummedData().getContent();
+      if (bytesToSkipBeforeReading >= 0 && bytesToSkipBeforeReading < content.size()) {
+        content = res.getChecksummedData().getContent().substring((int) bytesToSkipBeforeReading);
```
mayanks, line 577: Is there a risk of calling `res.getChecksummedData()` twice above? Either with respect to idempotency or latency. Wouldn't this be better?
```
if ......
  content = content.substring((int) bytesToSkipBeforeReading);
...
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -536,48 +565,46 @@ private boolean isByteBufferBeyondCurrentRequestRange(ByteBuffer byteBuffer) {
...
-          stream = null;
-        }
-      } finally {
-        if (stream != null) {
-          stream.close();
-        }
+      if (responseSizeLargerThanRemainingBuffer) {
+        invalidateBufferedContent();
+        bufferedContent = content;
+        bufferedContentReadOffset = bytesToWrite;
```
mayanks, line 600: Just like we slice off the content when `bytesToSkipBeforeReading` is > 0, wouldn't it be consistent to do the same thing here and keep bufferedContentReadOffset 0.
```
  bufferredContent = content.substring(bytesToWrite)
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mayanks: can we check if substring is just pointer manipulation and not actual data copy.

mprashanthsagar: Done.

Yes, [Documentation](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString.html#substring-int-int-) suggests it not actual data copy

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: In case of exceptions later on, where we would have to re-populate the buffer, we would lose the data. Holding on to the gcs data one chunk at a time should not be that expensive

mayanks (approved):

PR #652, opened 2021-11-18
Add option to enable directpath for gRPC

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -37,7 +37,7 @@
   public static final boolean ENABLE_GRPC_DEFAULT = false;
 
   /** Default setting to prefer DirectPath for gRPC. */
-  public static final boolean DIRECT_PATH_PREFFERED_DEFAULT = true;
+  public static final boolean DIRECT_PATH_PREFERRED_DEFAULT = true;
```
Deependra-Patel, line 40: Do we want to make the false by default? So whoever uses our client doesn't have to override themselves?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -375,6 +375,10 @@
   public static final HadoopConfigurationProperty<Long> GCS_GRPC_WRITE_TIMEOUT_MS =
       new HadoopConfigurationProperty<>("fs.gs.grpc.write.timeout.ms", 10 * 60 * 1000L);
 
+  /** Configuration key for enabling use of directpath gRPC API for read/write. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_DIRECTPATH_ENABLE =
```
Deependra-Patel, line 379: Same comment as below

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
mprashanthsagar: Switching to Cloudpath should be Opt-in IMO. As most customers would benefit from Directpath.

Swithcing to cloudpath would be an option only in case of RVIP configurations. Customers who don't use RVIP with VPC-SC would be penalised with a default of cloudpath

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: Dup

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
Deependra-Patel: If we communicate to customer who will use with RVIP, then I guess it is okay. Not sure how communication to customer in preview is going to work. Otherwise the customer who use RVIP may spend lot of time trying to debug something which anyway wouldn't work.

Also I am assuming in VPC-SC tests, we should be able to override this to false in their config.

Deependra-Patel (approved):

PR #666, opened 2021-12-28
Switch CI to Java 11

medb:


mprashanthsagar:
Is this change meant for 3.x only OR do we plan to cherry pick this for 2.2.x ?

We may want to keep our release scripts in sync with CI

medb:
I do not think that we want to cherry-pick this to 2.x branch, but we may want to run CI for multiple Java versions (8, 11 and 17) as main branch - I think we can explore this as follow up.

mprashanthsagar (approved):

PR #668, opened 2021-12-28
Add missing change/realse notes

medb:


ranu010101 (approved):

gcs/CHANGES.md
```diff
@@ -22,6 +22,10 @@
     fs.gs.http.transport.type
     ```
 
+1.  Support GCS fine-grained action in AuthorizationHandlers.
+
+1.  Decrease log level for `hflush` rate limit log message.
+
```
mprashanthsagar, line 28: @medb 
This was a change in 2.2.4. Do we want to include it in under 3.0 as well ? 

If not, could we please revert this

medb: We already have this in 2.2.4 release notes: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/branch-2.2.x/gcs/CHANGES.md#224---2021-11-05

mprashanthsagar: Yes. Adding these two again in 3.0.0 doesn't make sense. 

Should we sync 2.2.x branch changes.md file ? The change is not specific to 3.0

medb: Discussed offline, seems like sync is not necessary.

PR #676, opened 2022-01-06
Shutdown background threadpool on GCSImpl close

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -246,6 +247,10 @@ public Boolean load(String bucketName) {
               .setDaemon(true)
               .build());
 
+  // Thread-pool used for background watchdog
+  private final ScheduledExecutorService scheduledExecutorService =
+      newSingleThreadScheduledExecutor();
```
medb, line 252: We already have `backgroundTasksThreadPool` that we use for different background tasks, can we reuse it?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: We need an instance of `ScheduledExecutorService` for this usecase

medb (approved):

PR #677, opened 2022-01-08
Mark watchdog executor service as daemon

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -249,7 +249,11 @@ public Boolean load(String bucketName) {
 
   // Thread-pool used for background watchdog
   private final ScheduledExecutorService scheduledExecutorService =
```
medb, line 251: If it's used only in watchdog, I think that maybe it will be better to actually move this thread pool management in the `Watchdog` class itself? It will allow to not create this thread pool if gRPC is not used, and I'm actually not sure what are implication if something else will start to use this thread pool in the future, maybe it will cause bugs/threads starvation?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Makes sense. We would need only one instance of watchdog anyways.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -370,9 +364,7 @@ public GoogleCloudStorageImpl(
             : this.storage.getRequestFactory().getInitializer();
 
     this.watchdog =
-        Watchdog.create(
-            Duration.ofMillis(storageOptions.getGrpcMessageTimeoutCheckInterval()),
-            scheduledExecutorService);
+        Watchdog.create(Duration.ofMillis(storageOptions.getGrpcMessageTimeoutCheckInterval()));
```
medb, line 367: Let's create it only if gRPC is enabled.

medb (changes requested):

medb (approved):

PR #681, opened 2022-01-13
Retry request on 408/timeout response from server

Deependra-Patel:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -416,6 +417,70 @@ public void upload_failure_runtimeException() throws Exception {
         .inOrder();
   }
 
+  @Test
+  public void upload_retry_requestTimeout() throws Exception {
+    byte[] testData = new byte[MediaHttpUploader.MINIMUM_CHUNK_SIZE];
```
mprashanthsagar, line 422: let's add a test for multiple chunks

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -154,7 +160,8 @@ public CredentialOrBackoffResponseHandler() {
       errorCodeHandler.setBackOffRequired(
           response ->
               BASE_HTTP_BACKOFF_REQUIRED.isRequired(response)
-                  || response.getStatusCode() == HTTP_SC_TOO_MANY_REQUESTS);
+                  || response.getStatusCode() == HTTP_SC_TOO_MANY_REQUESTS
+                  || response.getStatusCode() == HTTP_REQUEST_TIMEOUT);
```
mprashanthsagar, line 164: How about adding a list of retryable http codes ?

mprashanthsagar (changes requested):

mprashanthsagar:
`RetryHttpInitializer` has multiple usages, Is retry on 408 safe for all of the requests which use `RetryHttpInitializer` ?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
medb, line 164: I think that `google-cloud-storage` library added retries only for idempotent requests (`GET`, `HEAD`, etc), maybe we need to check with them and clarify if/why this additional check necessary?

cc @benwhitehead

medb (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
BenWhitehead: This is correct. Not all GCS operations neatly fit into the standard HTTP idempotency conventions, in some cases extra parameters need to be taken into account to determine if a request is idempotent.

These docs https://cloud.google.com/storage/docs/retry-strategy#idempotency give a good overview of some of the specifics.

Deependra-Patel: Good point Igor, ideally these should be handled by libraries themselves.
1. Let me know if I understand correctly. We use `google-api-services-storage` instead of `google-cloud-storage` from https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/pom.xml#L70.

2. Thanks for link @BenWhitehead, on same page, I can see under Client Libraries -> Java, 408 is actually retried by default. Under code snippet, I see `com.google.cloud.storage.Storage` as class while we use `com.google.api.services.storage.Storage` (code link below), is the documentation still valid?

3. @medb in the above documentation, it is mentioned that we can override default settings which we do but bit differently from what is mentioned in the sample code, we use HttpRequestInitializer instead, this may be the reason why we are not benefitting from default retry settings? https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java#L364 cc: @BenWhitehead

BenWhitehead: Correct `google-api-services-storage` is the Discovery doc based automatically generated JSON client (aka Apiary client) (`google-cloud-storage` also uses this for the majority of it's RPCs) which doesn't have any automatic retry AFAIK, something like a request interceptor or request initializer is what would need to be leveraged to have the Apiary client perform multiple attempts. In `google-cloud-storage` we handle retries at a higher level as some of our RPCs are out of band of the apiary client.

Any links to code from `google-cloud-storage` are intended to be purely informational to show how that client has approached retry handling for some cases. The https://cloud.google.com/storage/docs/retry-strategy#idempotency doc is the general public which serves as a guide for anyone doing client level work against gcs (including the `google-cloud-storage` team as well as other customers).

Deependra-Patel: Now have a Set

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -123,6 +123,7 @@
       new StorageResourceId(BUCKET_NAME, OBJECT_NAME);
 
   private static final int STATUS_CODE_RESUME_INCOMPLETE = 308;
+  private static final int STATUS_CODE_REQUEST_TIMEOUT = 408;
```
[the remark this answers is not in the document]
Deependra-Patel: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -416,6 +417,70 @@ public void upload_failure_runtimeException() throws Exception {
...
+            getRequestString(BUCKET_NAME, OBJECT_NAME),
+            resumableUploadRequestString(
+                BUCKET_NAME, OBJECT_NAME, /* generationId= */ 0, /* replaceGenerationId= */ false),
+            resumableUploadChunkRequestString(BUCKET_NAME, OBJECT_NAME, /* uploadId= */ 1),
+            resumableUploadChunkRequestString(BUCKET_NAME, OBJECT_NAME, /* uploadId= */ 2))
+        .inOrder();
+  }
+
+  @Test
+  public void upload_does_not_retry_forbidden() throws Exception {
```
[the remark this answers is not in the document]
Deependra-Patel: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
Deependra-Patel: RetryHttpInitializer is used for all the requests made for Json APIs. I think this test should be enough to display the retry behavior.

Deependra-Patel:
> `RetryHttpInitializer` has multiple usages, Is retry on 408 safe for all of the requests which use `RetryHttpInitializer` ?

@mprashanthsagar According this link https://cloud.google.com/storage/docs/retry-strategy#tools, under "Client Libraries" > "Java", we can see HTTP 408 is safe to retry for all requests as already being done by latest client.

medb:
> HTTP 408 is safe to retry for all requests as already being done by latest client.

@Deependra-Patel I think that `google-cloud-storage` lib recently made retries idempotency aware: https://github.com/googleapis/java-storage/pull/1132

We should do the same - seems like it's simple enough to do based on the available request data.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -422,6 +422,70 @@ public void upload_failure_runtimeException() throws Exception {
...
+            emptyResponse(HttpStatusCodes.STATUS_CODE_NOT_FOUND),
+            resumableUploadResponse(BUCKET_NAME, OBJECT_NAME),
+            emptyResponse(HttpStatusCodes.STATUS_CODE_FORBIDDEN),
+            jsonDataResponse(
+                newStorageObject(BUCKET_NAME, OBJECT_NAME)
+                    .setSize(BigInteger.valueOf(testData.length))));
+
+    GoogleCloudStorage gcs = mockedGcs(GCS_OPTIONS, transport);
+
+    WritableByteChannel writeChannel = gcs.create(RESOURCE_ID);
```
singhravidutt, line 471: nit: can we follow the same style as in above test case i.e. try{...}. I am fine with any as far as it's consistent.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -422,6 +422,70 @@ public void upload_failure_runtimeException() throws Exception {
...
+            resumableUploadResponse(BUCKET_NAME, OBJECT_NAME),
+            emptyResponse(408), // HTTP 408 Request Timeout
+            jsonDataResponse(
+                newStorageObject(BUCKET_NAME, OBJECT_NAME)
+                    .setSize(BigInteger.valueOf(testData.length))));
+
+    GoogleCloudStorage gcs = mockedGcs(GCS_OPTIONS, transport);
+
+    try (WritableByteChannel writeChannel = gcs.create(RESOURCE_ID)) {
+      writeChannel.write(ByteBuffer.wrap(testData));
```
singhravidutt, line 442: shouldn't we also call `close()`?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -422,6 +422,70 @@ public void upload_failure_runtimeException() throws Exception {
...
+  @Test
+  public void upload_retry_requestTimeout() throws Exception {
+    byte[] testData = new byte[MediaHttpUploader.MINIMUM_CHUNK_SIZE];
+    new Random().nextBytes(testData);
+
+    MockHttpTransport transport =
+        mockTransport(
+            emptyResponse(HttpStatusCodes.STATUS_CODE_NOT_FOUND),
+            resumableUploadResponse(BUCKET_NAME, OBJECT_NAME),
+            emptyResponse(408), // HTTP 408 Request Timeout
```
singhravidutt, line 434: What about creating 408 status code in `HttpStatusCodes`?

Deependra-Patel: This class is in `google-http-client` library and it only has popular status codes and 408 is not there.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
Deependra-Patel: This is in `try` with resource, so will be closed automatically.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
Deependra-Patel: On line 474, we expect an error on call to `close`,  hence can't use `try {}`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
singhravidutt: aah, Cool.

singhravidutt (approved):

PR #686, opened 2022-01-21
core-site.xml to always include auth-keys.xml

steveloughran:
* adds an empty <fallback/> element to ignore the absence of
  the file when it is not present.
* highlights you can do the same trick in auth-keys.xml to keep
  your secrets out the source tree entirely


steveloughran:
also needs auth keys to be explicitly added to .gitignore for safety.

medb (approved):

PR #653, opened 2021-11-18
Update read time based on object size

mprashanthsagar:

PR #691, opened 2022-01-24
Revert "Update read time based on object size (#653)"

mprashanthsagar:
This reverts commit 8d446a5a901d0354bdd38d8942761b3f830eef5e.


Deependra-Patel (commented):
IIUC we are reverting this because we have message level timeouts now. But from code I can see that is only for writes but this change is for reads. Could you please clarify why we are doing this change?

Deependra-Patel:
> > IIUC we are reverting this because we have message level timeouts now. But from code I can see that is only for writes but this change is for reads. Could you please clarify why we are doing this change?
> 
> Message level timeouts are applied for both reads and writes
> 
> Reference for [reads](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java#L744)

Didn't notice the other `watch` method.  LGTM.

Deependra-Patel (approved):

PR #692, opened 2022-01-24
Drain iterator for server-streaming grpc call cancellation 

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -759,6 +758,19 @@ private void cancelCurrentRequest() {
...
+  private void drainIterator() {
+    if (resIterator == null) {
+      return;
+    }
+    try {
+      while (resIterator.hasNext()) {
+        resIterator.next();
+      }
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log("Exception while draining the iteration on cancellation");
```
mayanks, line 770: Suggest we change this to Info as it is expected to raise exception

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -759,6 +758,19 @@ private void cancelCurrentRequest() {
     contentChannelEndOffset = -1;
   }
 
+  private void drainIterator() {
+    if (resIterator == null) {
```
mayanks, line 762: nit: maybe add the entire logic in 
```
if (resIterator != null) {
 ...
}
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -745,9 +745,8 @@ private void cancelCurrentRequest() {
       requestContext.close();
       requestContext = null;
     }
-    if (resIterator != null) {
-      resIterator = null;
-    }
+    drainIterator();
```
mayanks, line 748: nit: I know that `resIterator` is a global/object variable. It will be unambiguous if we can pass `resIterator` as parameter to `drainIterator`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: This is a common pattern `guard-clause`. I think it is clear to avoid nesting

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: I don't think it makes a difference for private method, rather it would be redundant. I would agree for any public api. 
`resIterator` is not global variable, it is an instance variable here.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Done. `expected exception` is  counter-intuitive. We should explore a clear solution, possibly embed this logic under `cancel` grpc api

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -759,6 +758,19 @@ private void cancelCurrentRequest() {
     contentChannelEndOffset = -1;
   }
 
+  private void drainIterator() {
```
Deependra-Patel, line 761: I think we should add a comment here why we are doing this which should ideally be handled by the client library.

Deependra-Patel (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Good call. Updating

mayanks (approved):

PR #693, opened 2022-01-24
Shared GCS delegation tokens

ntr:
This change introduces new configuration variable applicable to GCS connector `fs.gs.delegation.token.instantiation_strategy`. It defines how delegation tokens are instantiated. 

Its default value is 'GCS_TOKEN_INSTANTION_STRATEGY' and it does not change current functionality. Gcs tokens are instantiated per service and per renewer. 

When it is set to 'SHARED' - GCS delegation tokens are shared across multiple storage buckets. Tokens are still unique per renewer. 

This change applies to user authentication only and allows controlling user policies via Ranger plugin for jobs that use multiple GCS buckets at the same time.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -95,6 +96,17 @@
   public static final HadoopConfigurationProperty<String> DELEGATION_TOKEN_BINDING_CLASS =
       new HadoopConfigurationProperty<>("fs.gs.delegation.token.binding");
 
+  /**
+   * Configuration key for GCS delegation token instantiation strategy. SHARED means delegation
+   * tokens are shared across multiple instanced of Cloud storage file services . Default value:
+   * INSTANCE_PER_SERVICE
+   */
+  public static final HadoopConfigurationProperty<GCSTokenInstantiationStrategy>
+      GCS_TOKEN_INSTANTION_STRATEGY =
```
cyxxy, line 105: misspelled "INSTANTIATION".

Also, I suggest we still put "delegation" in the name. so:
"DELEGATION_TOKEN_INSTANTIATION_STRATEGY"

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GCSTokenInstantiationStrategy.java
```diff
@@ -0,0 +1,9 @@
+package com.google.cloud.hadoop.fs.gcs.auth;
+
+public enum GCSTokenInstantiationStrategy {
```
cyxxy, line 3: May I suggest "DelegationTokenInstantiationStrategy"? Or at least mention "delegation"

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,11 +146,99 @@ public void testDelegationTokenStatistics() throws IOException {
...
+    assertThat(dt1).isNotEqualTo(dt2);
+    assertThat(dt1).isNotEqualTo(dt3);
+    assertThat(dt3).isNotEqualTo(dt2);
+
+    fs.close();
+    fsOther.close();
+    fsNullRenewer.close();
+  }
+
+  private Configuration loadConfig(GCSTokenInstantiationStrategy gcsTokenInstantiationStrategy) {
```
cyxxy, line 234: I think you can overload this method definition so default instantiation strategy is one instance per service.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -50,7 +51,7 @@ public void testDelegationTokenBinding() throws IOException {
     Text expectedKind = TestTokenIdentifierImpl.KIND;
 
     GoogleHadoopFileSystem fs = new GoogleHadoopFileSystem();
-    fs.initialize(initUri, loadConfig());
```
cyxxy, line 53: overload the method "loadConfig" so you do not have to change the call here.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,11 +146,99 @@ public void testDelegationTokenStatistics() throws IOException {
...
+    Token<?> dt1 = fs.getDelegationToken("current_user");
+    Token<?> dt2 = fsOther.getDelegationToken("current_user");
+    assertThat(dt1).isEqualTo(dt2);
+
+    fs.close();
+    fsOther.close();
+  }
+
+  @Test
+  public void testDelegationAreSharedForSharedTokensConfigAndNullRenewers() throws IOException {
```
cyxxy, line 189: testDelegation*Tokens*AreShared....

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,11 +146,99 @@ public void testDelegationTokenStatistics() throws IOException {
...
+    Token<?> dt1 = fs.getDelegationToken(null);
+    Token<?> dt2 = fsOther.getDelegationToken(null);
+    assertThat(dt1).isEqualTo(dt2);
+
+    fs.close();
+    fsOther.close();
+  }
+
+  @Test
+  public void testDelegationAreNotSharedForSharedTokensConfigAndDifferentRenewers()
```
cyxxy, line 207: again, Delegation*Tokens*

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,11 +146,99 @@ public void testDelegationTokenStatistics() throws IOException {
...
+    Token<?> dt2 = fs.getDelegationToken(null);
+    assertThat(fs.getIOStatistics().counters().get(INVOCATION_GET_DELEGATION_TOKEN.getSymbol()))
+        .isEqualTo(2);
+    assertThat(fs.getIOStatistics().counters().get(DELEGATION_TOKENS_ISSUED.getSymbol()))
+        .isEqualTo(2);
+    fs.close();
+  }
+
+  @Test
+  public void testDelegationAreSharedForSharedTokensConfigAndSameRenewers() throws IOException {
```
cyxxy, line 171: Delegation*Tokens*

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -66,7 +75,7 @@ public GcsDelegationTokens() throws IOException {
   @Override
   public void serviceInit(Configuration conf) {
     String tokenBindingImpl = DELEGATION_TOKEN_BINDING_CLASS.get(conf, conf::get);
-
+    this.tokenInstantiationStrategy = GCS_TOKEN_INSTANTION_STRATEGY.get(conf, conf::getEnum);
```
medb: nit: `this.` is redundant

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -238,6 +247,24 @@ public boolean isBoundToDT() {
       return getBoundDT();
     }
 
+    // not bounded tokens, but using shared token instantiation strategy.
+    // in this case we create or reuse token per renewer but share across multiple services.
+    if (this.tokenInstantiationStrategy == GCSTokenInstantiationStrategy.SHARED) {
+      Optional<String> renewerKey = Optional.ofNullable(renewer);
```
medb, line 254: Not sure if it's good idea to use optional as a map key, can we use some constant for empty values instead?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -22,9 +22,10 @@
 import static com.google.cloud.hadoop.util.HadoopCredentialConfiguration.getConfigKeyPrefixes;
 import static com.google.common.base.Strings.nullToEmpty;
 
-import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.GcsFileChecksumType;
-import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.GlobAlgorithm;
-import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.OutputStreamType;
+import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase.GcsFileChecksumType;
+import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase.GlobAlgorithm;
+import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase.OutputStreamType;
+import com.google.cloud.hadoop.fs.gcs.auth.DelegationTokenInstantiationStrategy;
```
medb, line 28: Fix these imports

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: Any string value can come from the client - I wanted to make sure that those values are not mixed with each other and tokens are not reused in the scenarios they are not supposed to.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,6 +140,89 @@ public void testDelegationTokenStatistics() throws IOException {
...
+
+    GoogleHadoopFileSystem fs = new GoogleHadoopFileSystem();
+    fs.initialize(new Path("gs://test/").toUri(), loadSharedDelegationTokenConfig());
+
+    GoogleHadoopFileSystem fsOther = new GoogleHadoopFileSystem();
+    fsOther.initialize(new Path("gs://test2/").toUri(), loadSharedDelegationTokenConfig());
+
+    Token<?> dt1 = fs.getDelegationToken("current_user");
+    Token<?> dt2 = fsOther.getDelegationToken("current_user");
+    assertThat(dt1).isEqualTo(dt2);
```
cnauroth, line 176: `Token` implements a value `equals`, so in theory, there could be a bug introduced that produces 2 separate but equivalent instances. I suggest changing from `isEqualTo` to `isSameInstanceAs`.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,6 +140,89 @@ public void testDelegationTokenStatistics() throws IOException {
...
+
+    GoogleHadoopFileSystem fs = new GoogleHadoopFileSystem();
+    fs.initialize(new Path("gs://test/").toUri(), loadSharedDelegationTokenConfig());
+
+    GoogleHadoopFileSystem fsOther = new GoogleHadoopFileSystem();
+    fsOther.initialize(new Path("gs://test2/").toUri(), loadSharedDelegationTokenConfig());
+
+    Token<?> dt1 = fs.getDelegationToken(null);
+    Token<?> dt2 = fsOther.getDelegationToken(null);
+    assertThat(dt1).isEqualTo(dt2);
```
cnauroth, line 194: Same comment here about `isSameInstanceAs`.

cnauroth (commented):
@ntr , thanks for sending in this pull request. I've entered a few comments. Additionally, would you please document the new property in gcs/CONFIGURATION.md?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
medb: We can use an empty string if it's not a valid use input, but if we want to be more clear about uniqueness of shared token strategy we may want to use `AtomicReference` for shred token, instead of trying to put it in the map with non-shared tokens.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -238,6 +248,24 @@ public boolean isBoundToDT() {
...
+    // in this case we create or reuse token per renewer but share across multiple services.
+    if (this.tokenInstantiationStrategy == DelegationTokenInstantiationStrategy.SHARED) {
+      Optional<String> renewerKey = Optional.ofNullable(renewer);
+      synchronized (sharedTokens) {
+        if (sharedTokens.containsKey(renewerKey)) {
+          logger.atFine().log("Reusing the shared GCS delegation token");
+          return sharedTokens.get(renewerKey);
+        } else {
+          Token<DelegationTokenIdentifier> token =
+              tokenBinding.createDelegationToken(renewer, getStats());
```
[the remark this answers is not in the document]
medb: Why did we switch to HashMap instead of ConcurrentHashMap?

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
cnauroth: I'm not confident that we can use an empty string as a sentinel value, considering wider Hadoop usage where only `null` represents a token without a renewer.

Another option would be to define a custom map key class that carefully handles the semantics of null vs. non-null renewer, but that seemed like the heaviest option compared to alternatives.

I'm unclear about the `AtomicReference` suggestion. Is the idea that a token with `null` renewer goes into separate storage from the map? Can you please elaborate?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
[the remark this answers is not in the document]
cnauroth: @ntr ?

From my perspective:
* `HashMap<String, Token<DelegationTokenIdentifier>>`: Pro: supports null keys, so easy to represent the "no renewer" use case. Con: forces more coarse-grained locking.
* `ConcurrentHashMap<Optional<String>, Token<DelegationTokenIdentifier>>`: Pro: fine-grained locking. Con: it's maybe weird to use `Optional` as key? (I really don't think it's a problem though.)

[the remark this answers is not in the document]
ntr: With ConcurrentHashMap map I was not sure how to implement synchronization without putting lock on the whole block. In this case it did not seem to provide much value. However, now learned about [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-) method  - that would simplify the code a lot.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: @medb another option I was considering adding an additional field for storing token when renewer is `null`, but it seems to be excessive.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/DelegationTokenInstantiationStrategy.java
```diff
@@ -0,0 +1,9 @@
+package com.google.cloud.hadoop.fs.gcs.auth;
```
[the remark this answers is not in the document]
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/DelegationTokenInstantiationStrategy.java
```diff
@@ -0,0 +1,9 @@
...
+
+public enum DelegationTokenInstantiationStrategy {
```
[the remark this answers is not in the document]
ntr: done

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
ntr: done

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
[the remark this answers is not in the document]
ntr: updated to ConcurrentHashMap

gcs/CONFIGURATION.md
```diff
@@ -36,7 +36,14 @@
...
 
     Delegation Token binding class.
+*   `fs.gs.delegation.token.instantiation_strategy` (default `INSTANCE_PER_SERVICE`)
 
+    Defines GCS delegation token instantiation strategy.
+
+    Valid values:
+      * `INSTANCE_PER_SERVICE`  New token instance is created for each service instance;
+      * `SHARED` -  Token instance is shared across the instances of GCS services. 
+         Different renewers still get unique token instances;
```
medb, line 46: Add new lines before and after this config doc.

gcs/CONFIGURATION.md
```diff
@@ -36,7 +36,14 @@
 *   `fs.gs.delegation.token.binding` (not set by default)
 
     Delegation Token binding class.
+*   `fs.gs.delegation.token.instantiation_strategy` (default `INSTANCE_PER_SERVICE`)
```
medb, line 39: Looking at other properties - we do not use `_` in the property name, so probably better to be consistent and rename this property to `fs.gs.delegation.token.instantiation.strategy`

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/DelegationTokenInstantiationStrategy.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
medb, line 2: Fix year

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -58,6 +62,11 @@
   /** Active Delegation token. */
   private Token<DelegationTokenIdentifier> boundDT;
 
+  private DelegationTokenInstantiationStrategy tokenInstantiationStrategy =
+      DelegationTokenInstantiationStrategy.INSTANCE_PER_SERVICE;
+  private static ConcurrentHashMap<Optional<String>, Token<DelegationTokenIdentifier>>
```
medb, line 67: This should be final.

medb, line 67: Also do we need to make it static, so it's shared between instances of these class even though they can be used by different FS objects?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -66,7 +75,8 @@ public GcsDelegationTokens() throws IOException {
   @Override
   public void serviceInit(Configuration conf) {
     String tokenBindingImpl = DELEGATION_TOKEN_BINDING_CLASS.get(conf, conf::get);
-
+    this.tokenInstantiationStrategy =
```
medb, line 78: `this.` prefix usually used only in constructors/setters when there is a parameter with the same name.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
```diff
@@ -238,6 +248,27 @@ public boolean isBoundToDT() {
...
     }
 
+    // not bounded tokens, but using shared token instantiation strategy.
+    // in this case we create or reuse token per renewer but share across multiple services.
+    if (this.tokenInstantiationStrategy == DelegationTokenInstantiationStrategy.SHARED) {
+      try {
+        return sharedTokens.computeIfAbsent(
+            Optional.ofNullable(renewer),
+            k -> {
+              logger.atFine().log("Created a shared GCS delegation token");
```
medb, line 258: Log `renewer` here?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -126,6 +127,9 @@
           put("fs.gs.storage.root.url", "https://storage.googleapis.com/");
           put("fs.gs.storage.service.path", "storage/v1/");
           put("fs.gs.working.dir", "/");
+          put(
+              "fs.gs.delegation.token.instantiation_strategy",
```
medb, line 131: nit: sort alphabetically by keys

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
medb, line 67: Is there a reason that we use nullable String as a renewer/key? Seems like in Hadoop code base they use `Text` object for renewer: https://github.com/apache/hadoop/blob/02f6bad1fff0f9c8e8dc5899272a93ab64d15eb9/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/auth/delegation/S3ADelegationTokens.java#L418

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: I noticed that renewer is not logged in other places of this class. I was wondering if that might be a privacy or security concern?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: Yes, that's the intention of this cache - to be able to share tokens across multiple GCS filsystems.

ntr: Existing code related to GCS delegation tokens already uses string as a renewer. 

Later it is converted to `Text` when DelegationTokenIdentifier is created - https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/AbstractDelegationTokenBinding.java#L123.

I do not have context on that decision.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
cnauroth: The renewer identifies a token-renewing user (e.g. YARN) but not any private secrets owned by that user. It would be safe to log.

gcs/CONFIGURATION.md
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/DelegationTokenInstantiationStrategy.java
ntr: done

gcs/CONFIGURATION.md
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
ntr: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/auth/GcsDelegationTokens.java
medb: I think that it's better to switch code to use `Text` as soon as possible - in this case we will not need `Optional` wrapper for key values.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
ntr: done

gcs/CONFIGURATION.md
```diff
@@ -37,6 +37,15 @@
 
     Delegation Token binding class.
 
+*   `fs.gs.delegation.token.instantiation.strategy` (default `INSTANCE_PER_SERVICE`)
```
medb, line 40: nit: add colon after `default` for consistency with rest of the values: `default:`

medb, line 40: Add release note about this to CHANGES.md

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -856,8 +856,12 @@ public FileStatus getFileStatus(Path hadoopPath) throws IOException {
   public Token<?> getDelegationToken(String renewer) throws IOException {
     incrementStatistic(GhfsStatistic.INVOCATION_GET_DELEGATION_TOKEN);
     Token<?> result = null;
+    Text renewerText = new Text();
+    if (renewer != null) {
+      renewerText.set(renewer);
+    }
```
medb, line 862: Use ternary operator with constructor to simplify this code:

```java
    Text renewerText = renewer == null ? new Text() : new Text(renewer);
```

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,6 +140,89 @@ public void testDelegationTokenStatistics() throws IOException {
...
+  public void testDelegationTokenStatisticsForSharedTokens() throws IOException {
+
+    GoogleHadoopFileSystem fs = new GoogleHadoopFileSystem();
+    fs.initialize(new Path("gs://test/").toUri(), loadSharedDelegationTokenConfig());
+
+    Token<?> dt = fs.getDelegationToken("shared_token_test1");
+    assertThat(fs.getIOStatistics().counters().get(INVOCATION_GET_DELEGATION_TOKEN.getSymbol()))
+        .isEqualTo(1);
+    assertThat(fs.getIOStatistics().counters().get(DELEGATION_TOKENS_ISSUED.getSymbol()))
+        .isEqualTo(1);
```
medb, line 153: Use map asserts:

```java
    assertThat(fs.getIOStatistics().counters()).contatinsExactly(
        INVOCATION_GET_DELEGATION_TOKEN.getSymbol(), 1,
        DELEGATION_TOKENS_ISSUED.getSymbol(), 1);
```

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -139,6 +140,89 @@ public void testDelegationTokenStatistics() throws IOException {
...
+
+    // second test for renewer which is null
+    Token<?> dt2 = fs.getDelegationToken(null);
+    assertThat(fs.getIOStatistics().counters().get(INVOCATION_GET_DELEGATION_TOKEN.getSymbol()))
+        .isEqualTo(2);
+    assertThat(fs.getIOStatistics().counters().get(DELEGATION_TOKENS_ISSUED.getSymbol()))
+        .isEqualTo(2);
```
medb, line 160: ditto

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemDelegationTokensTest.java
```diff
@@ -154,4 +238,12 @@ private Configuration loadConfig() {
 
     return config;
   }
+
+  private Configuration loadSharedDelegationTokenConfig() {
+    Configuration config = loadConfig();
+    config.setEnum(
+        GoogleHadoopFileSystemConfiguration.DELEGATION_TOKEN_INSTANTIATION_STRATEGY.getKey(),
```
medb, line 245: nit: use static import for `DELEGATION_TOKEN_INSTANTIATION_STRATEGY`

medb (approved):

PR #687, opened 2022-01-21
Enabled TCP keep alive while using HttpTransport

Deependra-Patel:
@frankyn The suggestion you made was for ApacheHttpTransport but we use NetHttpTransport in dataproc so couldn't use SocketConfig. 
This is how it will be used https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java#L173, so in case the condition fails, we won't have these settings -`connection instanceof HttpsURLConnection`. Shouldn't we be using Https all the time, so this condition would be true?

Let me know if this approach is good, then would extract the `20_000` and `true` to a config and also add unit tests.

Fixes #695

PR #696, opened 2022-01-27
Enable TCP keep alive while using HttpTransport

Deependra-Patel:
Reviewed here #687 
Extra diff is doing the same for Apache transport

### DIFF
--- a/util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
@@ -24,19 +24,26 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.flogger.GoogleLogger;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.Authenticator;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.PasswordAuthentication;
 import java.net.Proxy;
+import java.net.Socket;
+import java.net.SocketException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.GeneralSecurityException;
 import javax.annotation.Nullable;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLSocketFactory;
 import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
+import org.apache.http.config.SocketConfig;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.HttpClientBuilder;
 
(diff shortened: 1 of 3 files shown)


util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -138,6 +145,9 @@ public static ApacheHttpTransport createApacheHttpTransport(
 
     HttpClientBuilder httpClientBuilder = ApacheHttpTransport.newDefaultHttpClientBuilder();
 
+    SocketConfig keepAliveSocketConf = SocketConfig.custom().setSoKeepAlive(true).build();
+    httpClientBuilder.setDefaultSocketConfig(keepAliveSocketConf);
```
mprashanthsagar, line 149: let's add a test to cover this path

mprashanthsagar:
Please fix the presubmit build failures

Deependra-Patel:
> Please fix the presubmit build failures

First time it failed due to maven dependency not able to download, on retry the next time it failed due to a transient test failing. It succeeds now.

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
Deependra-Patel: I tried to find a way to test this but couldn't. So this config goes into building an [HttpClient](https://javadoc.io/doc/org.apache.httpcomponents/httpclient/latest/org/apache/http/client/HttpClient.html) which doesn't expose any method to expose this config.

I think it is fine to not unit test this config as it is obvious and we anyway don't have this in future versions. Also, for this version it is not default.

Deependra-Patel: wdyt?

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -233,4 +246,68 @@ static URI parseProxyAddress(@Nullable String proxyAddress) {
...
+
+    public Socket createSocket(
+        InetAddress address, int port, InetAddress clientAddress, int clientPort)
+        throws IOException {
+      return setSocketKeepAlive(
+          wrappedSockedFactory.createSocket(address, port, clientAddress, clientPort));
+    }
+
+    private static Socket setSocketKeepAlive(Socket socket) throws SocketException {
+      socket.setKeepAlive(true);
```
medb, line 309: Please, add a release note for this in CHANGES.md

medb (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
Deependra-Patel: Done

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
mprashanthsagar: ack

mprashanthsagar (approved):

PR #697, opened 2022-01-28
Enable TCP keep alive while using HttpTransport

Deependra-Patel:
Had to upgrade google-http-client to v2

### DIFF
--- a/util/pom.xml
+++ b/util/pom.xml
@@ -35,6 +35,11 @@
       <artifactId>google-api-client-java6</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-apache-v2</artifactId>
+      <version>1.41.1</version>
+    </dependency>
+    <dependency>
       <groupId>com.google.api-client</groupId>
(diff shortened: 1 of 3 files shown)


util/pom.xml
```diff
@@ -34,6 +34,11 @@
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client-java6</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-apache-v2</artifactId>
+      <version>1.41.1</version>
+    </dependency>
```
medb, line 41: This is a breaking change, we should not do this. If we can not enable TCP keep alive for Apache transport without this change, then it's fine to enable it for Java Net transport only, I think.

medb (changes requested):

util/pom.xml
Deependra-Patel: "breaking change", you mean in terms of dependency conflicts or clients not expecting this version, so we need a major version change?

Btw I also raised another PR without upgrading #698

PR #700, opened 2022-01-30
Support publishing metrics to Google cloud monitoring via Opencensus API

mprashanthsagar:
In the first iteration, adding support for grpc read channel metrics


gcs/pom.xml
```diff
@@ -225,6 +225,11 @@
                   <include>io.grpc</include>
                   <include>io.opencensus</include>
```
medb, line 226: I thought that gRPC already integrated with OpenCensus, can we leverage this integration instead of writing our custom?

medb (changes requested):

gcs/pom.xml
mprashanthsagar: The intent is to help debug performance issues. We will be extending support on JSON as well, to understand the current state.

We do want to include multiple custom metrics (number of retry attempts, calls to gcsio apis)

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -126,6 +126,7 @@
           put("fs.gs.storage.root.url", "https://storage.googleapis.com/");
           put("fs.gs.storage.service.path", "storage/v1/");
           put("fs.gs.working.dir", "/");
+          put("fs.gs.metrics.enable", false);
```
medb, line 129: sort alphabetically

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -351,17 +395,43 @@ private static GoogleCloudStorageItemInfo getObjectMetadata(
...
+      publishMetric(
+          resourceId,
+          metricsPublisher,
+          GET_OBJECT_METADATA,
+          SUCCESS,
+          JSON,
+          null,
+          "",
+          LATENCY_MS,
+          (System.currentTimeMillis() - startTime));
```
medb, line 407: This seems like too verbose, can we have some helpers that do not need so many parameters?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -351,17 +395,43 @@ private static GoogleCloudStorageItemInfo getObjectMetadata(
...
     } catch (IOException e) {
+      publishMetric(
+          resourceId,
+          metricsPublisher,
+          GET_OBJECT_METADATA,
+          ERROR,
+          JSON,
+          e,
+          e.getMessage(),
```
medb, line 416: Why do we need to pass exception message and exception itself?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -351,17 +395,43 @@ private static GoogleCloudStorageItemInfo getObjectMetadata(
...
+          LATENCY_MS,
+          (System.currentTimeMillis() - startTime));
```
medb, line 418: Not sure if `System.currentTimeMillis()` is the best option to use for metrics, we may want to use `Stopwatch.elapsed()` instead.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -376,6 +446,24 @@ private static GoogleCloudStorageItemInfo getObjectMetadata(
...
+      MetricsPublisher metricsPublisher,
+      String method,
+      String status,
+      String protocol,
+      Exception cause,
+      String errorMessage,
+      MeasureLong measureLong,
+      long time) {
+    TagKey[] keys = new TagKey[] {KEY_METHOD, KEY_STATUS, KEY_ERROR, KEY_PROTOCOL};
+    String[] values = new String[] {method, status, errorMessage, protocol};
```
medb, line 460: Allocation arrays for each metric could be suboptimal

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -376,6 +446,24 @@ private static GoogleCloudStorageItemInfo getObjectMetadata(
...
+    metricsPublisher.recordLong(keys, values, measureLong, time);
+    logger.atFinest().withCause(cause).log(
+        "read channel, resource:%s, time:%d, method:%s, status:%s, protocol:%s",
+        resourceId, time, method, status, protocol);
```
medb, line 464: We may want to separate logging from metrics publishing

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -398,6 +399,11 @@ public GoogleCloudStorageImpl(
             ? null
             : this.storage.getRequestFactory().getInitializer();
 
+    if (this.storageOptions.isMetricsEnabled()) {
+      this.metricsPublisher =
+          StackDriverMetricsPublisher.create(options.getProjectId(), credentials);
```
medb, line 404: Use CloudMonitoring instead of StackDriver, because StackDriver is an obsolete name

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright 2022 Google Inc. All Rights Reserved.
```
medb, line 2: s/Inc/LLC

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+import io.opencensus.tags.Tagger;
+import io.opencensus.tags.Tags;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.concurrent.GuardedBy;
+
+/** Interface for exposing utility methods to publish metrics via open-census api-spec. */
+interface MetricsPublisher {
```
medb, line 51: Can we consolidate OpenCensus metrics publishing with Hadoop's Statistics infra, so we do not duplicate code that doing roughly the same, just exposes data differently?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+
+  @VisibleForTesting static final Object monitor = new Object();
+
+  @GuardedBy("monitor")
+  private static boolean initialized = false;
+
+  private static final TagMetadata TAG_METADATA_NO_PROPAGATION =
+      TagMetadata.create(TagTtl.NO_PROPAGATION);
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
```
medb, line 85: Move to the very top of the class

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+          "gcsio/requests", "The distribution of retry attempts for gcs calls", "By");
+
+  static final MeasureLong REQUEST_RETRIES =
+      MeasureLong.create("gcsio/retries", "The distribution of retry attempts for gcs calls", "By");
+
+  static final TagKey KEY_METHOD = TagKey.create("method");
+  static final TagKey KEY_STATUS = TagKey.create("status");
+  static final TagKey KEY_ERROR = TagKey.create("error");
+  static final TagKey KEY_THREAD = TagKey.create("thread");
+  static final TagKey KEY_PROTOCOL = TagKey.create("protocol");
```
medb, line 105: Probably `KEY_` is redundant prefix?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+
+  static final List<TagKey> TAG_KEYS =
+      Collections.unmodifiableList(
+          Arrays.asList(KEY_METHOD, KEY_STATUS, KEY_ERROR, KEY_THREAD, KEY_PROTOCOL));
```
medb, line 109: Use `ImmutableList.of()` instead

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+    return create(projectId, credentialAdapter);
+  }
+
+  public static MetricsPublisher create(String projectId, Credentials credentials) {
+    try {
+      registerAllViews();
+      setupStackDriverExporter(projectId, credentials);
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log("Exception while registering metrics publisher");
+      return new NoOpMetricsPublisher();
```
medb, line 125: Of you are throwing exception then you do not need to log, just attach cause to it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+
+  private static void setupStackDriverExporter(String projectId, Credentials credentials)
+      throws IOException {
+    synchronized (monitor) {
+      if (!initialized) {
+        StackdriverStatsExporter.createAndRegister(
+            StackdriverStatsConfiguration.builder()
+                .setCredentials(credentials)
+                .setProjectId(projectId)
+                .setExportInterval(fromMillis(60000))
```
medb, line 138: Seems like 60s is too high, maybe at least 30s?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log("Exception while registering metrics publisher");
+      return new NoOpMetricsPublisher();
+    }
+    return new StackDriverMetricsPublisher();
+  }
+
+  private static void setupStackDriverExporter(String projectId, Credentials credentials)
+      throws IOException {
+    synchronized (monitor) {
```
medb, line 132: Can we make it singleton instead via `INSTANCE` constant?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+      if (!initialized) {
+        StackdriverStatsExporter.createAndRegister(
+            StackdriverStatsConfiguration.builder()
+                .setCredentials(credentials)
+                .setProjectId(projectId)
+                .setExportInterval(fromMillis(60000))
+                .build());
+        Runtime.getRuntime().addShutdownHook(new Thread(StackdriverStatsExporter::unregister));
```
medb, line 140: Is this really necessary?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+  }
+
+  private static void registerAllViews() {
+    // Defining the distribution aggregations in milliseconds
+    Aggregation latencyDistribution =
+        Distribution.create(
+            BucketBoundaries.create(
+                Arrays.asList(
```
medb, line 151: Use ImmutableList.of

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,243 @@
...
+                TAG_METADATA_NO_PROPAGATION)
+            .put(key, TagValue.create(value), TAG_METADATA_NO_PROPAGATION)
+            .build();
+    try (Scope ignored = tagger.withTagContext(tagContext)) {
+      statsRecorder.newMeasureMap().put(ml, n).record();
+    }
+  }
+
+  @Override
+  public void recordLong(TagKey[] keys, String[] values, MeasureLong ml, Long n) {
```
medb, line 214: Can we use map instead of keys and values?

pom.xml
```diff
@@ -391,6 +392,27 @@
...
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-exporter-stats-stackdriver</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-contrib-grpc-metrics</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+
```
medb, line 415: nit: remove extra line

pom.xml
```diff
@@ -391,6 +392,27 @@
...
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-api</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-impl</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
```
medb, line 404: Do we really need explicitly depend on this? Aren't they pulled in by other 2 deps?

gcs/pom.xml
```diff
@@ -225,6 +225,11 @@
...
                   <include>io.perfmark</include>
+                  <include>com.lmax</include>
+                  <include>com.google.cloud</include>
+                  <include>com.google.api</include>
+                  <include>org.threeten:threetenbp</include>
+                  <include>org.apache.httpcomponents</include>
```
medb, line 232: Sort alphabetically

gcs/pom.xml
```diff
@@ -250,6 +255,7 @@
                     <include>com.google.thirdparty.**</include>
                     <include>com.google.type.**</include>
                     <include>com.google.storage.**</include>
+                    <include>com.google.api.gax.**</include>
```
medb, line 258: Sort alphabetically

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -82,6 +97,8 @@
   // The size of this object generation, in bytes.
   private final long objectSize;
 
+  private final MetricsPublisher metricsPublisher;
```
medb, line 100: What is the gap that we are trying to close with custom gRPC metrics? Did we try enable [default gRPC metrics](https://opencensus.io/guides/grpc/java/#1) and see if they are sufficient for our needs?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -62,6 +71,12 @@
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
   protected static final String METADATA_FIELDS = "contentEncoding,generation,size";
+  static final String ERROR = "ERROR";
+  static final String SUCCESS = "SUCCESS";
+  static final String GET_OBJECT_METADATA = "getObjectMetadata";
+  static final String GET_OBJECT_MEDIA = "getObjectMedia";
+  static final String GRPC = "grpc";
+  static final String JSON = "json";
```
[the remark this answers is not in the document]
mprashanthsagar: I don't think enum fits here, I was planning to move to a Constants file once we add metrics from other classes

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -62,6 +71,12 @@
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
   protected static final String METADATA_FIELDS = "contentEncoding,generation,size";
+  static final String ERROR = "ERROR";
+  static final String SUCCESS = "SUCCESS";
```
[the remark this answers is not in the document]
mprashanthsagar: Will be be moved to a separate class (not MetricPublisher) when we add for other classes

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -217,41 +238,51 @@ private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
...
-    long startTime = System.currentTimeMillis();
-    try {
-      // TODO(b/190617054) : Migrate to gRPC requests for metadata when available
-      // Request only fields that are used for metadata initialization
-      Get metadataRequest = getMetadataRequest(gcs, resourceId).setFields(METADATA_FIELDS);
-      object =
-          ResilientOperation.retry(
-              metadataRequest::execute,
-              backOffFactory.newBackOff(),
-              RetryDeterminer.SOCKET_ERRORS,
```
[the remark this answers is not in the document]
mprashanthsagar: Good catch, it was unintended. I re-wrote based on what the metadata API can throw. 
Considering that the api can through IOException, i think it could be useful to retry more than SOCKET_ERRORS, For now, i think we could revert.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -217,41 +238,51 @@ private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
...
+    long time = stopwatch.elapsed(MILLISECONDS);
+    TagKey[] keys = new TagKey[] {METHOD, STATUS, PROTOCOL};
+    String[] values = new String[] {method, SUCCESS, protocol};
+    metricsPublisher.recordLong(keys, values, measure, time);
+  }
+
+  private void publishErrorMetric(
+      MeasureLong measure, Stopwatch stopwatch, String method, String protocol, Exception e) {
+    long time = stopwatch.elapsed(MILLISECONDS);
+    TagKey[] keys = new TagKey[] {METHOD, STATUS, PROTOCOL, CloudMonitoringMetricsPublisher.ERROR};
```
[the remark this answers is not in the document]
mprashanthsagar: ack

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -607,14 +634,22 @@ private void drainIterator() {
...
-    }
-    boolean moreDataAvailable = resIterator.hasNext();
-    if (!moreDataAvailable) {
-      cancelCurrentRequest();
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    try {
+      if (resIterator == null || requestContext == null || requestContext.isCancelled()) {
+        return false;
+      }
+      boolean moreDataAvailable = resIterator.hasNext();
```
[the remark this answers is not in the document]
mprashanthsagar: Yes. This call initiates a network call to GCS

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -607,14 +634,22 @@ private void drainIterator() {
...
+      publishSuccessMetric(MESSAGE_LATENCY_MS, stopwatch, GET_OBJECT_MEDIA, GRPC);
+
+      if (!moreDataAvailable) {
+        cancelCurrentRequest();
+      }
+      return moreDataAvailable;
```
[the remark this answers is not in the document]
mprashanthsagar: I can see the possibility of `cancel` throwing an exception. I don't exceptions out of metrics publishing

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,240 @@
...
+  static final TagKey METHOD = TagKey.create("method");
+  static final TagKey STATUS = TagKey.create("status");
+  static final TagKey ERROR = TagKey.create("error");
+  static final TagKey THREAD = TagKey.create("thread");
+  static final TagKey PROTOCOL = TagKey.create("protocol");
+
+  static final List<TagKey> TAG_KEYS = ImmutableList.of(METHOD, STATUS, ERROR, THREAD, PROTOCOL);
+
+  static final List<Double> RPC_MILLIS_BUCKET_BOUNDARIES =
+      ImmutableList.of(
```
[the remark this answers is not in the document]
mprashanthsagar: It is based on OpenCensus lib impl for grpc, so that we align on buckets for both standard and custom metrics. Unfortunately it isn't public for us to use directly

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,240 @@
...
+
+  @VisibleForTesting static final Object monitor = new Object();
+
+  @GuardedBy("monitor")
+  private static boolean initialized = false;
+
+  private static final TagMetadata TAG_METADATA_NO_PROPAGATION =
+      TagMetadata.create(TagTtl.NO_PROPAGATION);
+
+  static final MeasureLong LATENCY_MS =
```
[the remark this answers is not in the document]
mprashanthsagar: Can you elaborate ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
mprashanthsagar: We are not throwing an exception here, We rely on No-Op

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,240 @@
...
+  private static final Tagger tagger = Tags.getTagger();
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
+
+  public static MetricsPublisher create(String projectId, Credentials credentials) {
+    try {
+      registerAllViews();
+      setupStackDriverExporter(projectId, credentials);
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log("Exception while registering metrics publisher");
+      return new NoOpMetricsPublisher();
```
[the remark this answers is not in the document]
mprashanthsagar: Right now, we don't have metrics coverage and support on the feature. I would prefer to fallback instead of failing

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
mprashanthsagar: We need projectId and creds fed in. Singleton with INSTANCE. would get loaded early in jvm.

SIngleton in general is hard to test. I would prefer constructor injected dependencies for a class instead

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
mprashanthsagar: `unregister` method flushes the metrics. It is necessary for the metrics to not get lost. Within spark and connector context, a file system can get closed on a single jvm multiple times. So, we don't have a clean handle to flush and close the metrics publisher

pom.xml
```diff
@@ -391,6 +392,32 @@
...
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-contrib-grpc-metrics</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client-jackson2</artifactId>
```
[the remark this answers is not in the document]
mprashanthsagar: gcsio/pom.xml dependencies are derived based on the uber pom

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,240 @@
...
+              Name.create("gcsio/requests"),
+              "The distribution of request counts for a method",
+              REQUESTS,
+              AGGREGATION_WITH_COUNT_HISTOGRAM,
+              TAG_KEYS),
+        };
+
+    ViewManager viewManager = Stats.getViewManager();
+
+    for (View view : views) {
```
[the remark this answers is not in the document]
mprashanthsagar: There is no state that we could verify against on this class, as it is relied on registering with Cloudmonitoring. Integration test should suffice IMO

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
mprashanthsagar: Unfortunately, the opencensus API accepts arrays of keys and values. Type casting twice would be redundant.

pom.xml
mprashanthsagar: I don't think so.[Ref](https://opencensus.io/quickstart/java/metrics/#import-metrics-packages)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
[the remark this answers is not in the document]
Deependra-Patel: I still see retries for all IOException

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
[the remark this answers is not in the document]
Deependra-Patel: As mentioned here https://opencensus.io/stats/measure/, “By”, “1”, “ms” are standard units and not free text.  I was suggesting to have these strings as constants in this class to be re-used multiple times.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -338,51 +374,44 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+    long effectivePosition = positionInGrpcStream + bytesToSkipBeforeReading;
+    if ((footerBuffer == null) || (effectivePosition < footerStartOffsetInBytes)) {
+      OptionalLong bytesToRead = getBytesToRead(byteBuffer);
+      bytesRead += readFromGCS(byteBuffer, bytesToRead);
+    }
 
-      return bytesRead;
-    } finally {
-      long endTime = System.currentTimeMillis();
-      logger.atFinest().log(
```
Deependra-Patel, line 382: Let's keep the log, they are super useful during debugging. metricsPublisher is disabled by default and it only has aggregate metrics published to cloudmonitring.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -217,41 +237,57 @@ private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
...
-    } catch (IOException e) {
-      throw errorExtractor.itemNotFound(e)
-          ? createFileNotFoundException(resourceId, e)
-          : new IOException("Error reading " + resourceId, e);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      throw new IOException("Thread interrupt received.", e);
-    } finally {
-      long endTime = System.currentTimeMillis();
-      logger.atFinest().log(
```
Deependra-Patel, line 241: Same as other comment about logging

Deependra-Patel (changes requested):

Deependra-Patel (approved):

gcs/pom.xml
```diff
@@ -208,10 +208,13 @@
               </filters>
               <artifactSet>
                 <includes>
+                  <include>com.fasterxml.jackson.core:jackson-core</include>
```
medb, line 211: Why do we need Faster XML? We should use Gson instead.

gcs/pom.xml
```diff
@@ -220,9 +223,12 @@
                   <include>com.google.oauth-client</include>
                   <include>com.google.protobuf</include>
                   <include>com.google.storage.v2</include>
+                  <include>com.lmax</include>
                   <include>io.grpc</include>
                   <include>io.opencensus</include>
                   <include>io.perfmark</include>
+                  <include>org.apache.httpcomponents</include>
```
medb, line 230: Why do we need Apache HTTP client?

gcs/pom.xml
```diff
@@ -220,9 +223,12 @@
...
+                  <include>org.threeten:threetenbp</include>
```
medb, line 231: Does not look like this is something that we need: https://github.com/ThreeTen/threetenbp

gcs/pom.xml
```diff
@@ -208,10 +208,13 @@
               </filters>
               <artifactSet>
                 <includes>
+                  <include>com.fasterxml.jackson.core:jackson-core</include>
+                  <include>com.google.api</include>
                   <include>com.google.api-client</include>
                   <include>com.google.api.grpc</include>
                   <include>com.google.apis</include>
                   <include>com.google.auth</include>
+                  <include>com.google.cloud</include>
```
medb, line 217: What will this pull in?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -414,6 +414,10 @@
   public static final HadoopConfigurationProperty<RedactedString> GCS_ENCRYPTION_KEY_HASH =
       new HadoopConfigurationProperty<>("fs.gs.encryption.key.hash");
 
+  /** Configuration key to enable publishing metrics to Google cloud monitoring. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_METRICS_ENABLE =
+      new HadoopConfigurationProperty<>("fs.gs.metrics.enable", false);
```
medb, line 419: Probably better to make it more specific if this is gRPC or Cloud Monitoring specific, even better is to make this an enum, so we can support multiple metrics sinks in the future, if necessary

gcsio/pom.xml
```diff
@@ -146,6 +150,26 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+    </dependency>
```
medb, line 156: This probably redundant as `-impl` dep will pull it in?

gcsio/pom.xml
```diff
@@ -146,6 +150,26 @@
...
+      <artifactId>opencensus-exporter-stats-stackdriver</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-contrib-grpc-metrics</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client-jackson2</artifactId>
+    </dependency>
```
medb, line 172: Why do we need this dependency?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -62,6 +72,11 @@
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
   protected static final String METADATA_FIELDS = "contentEncoding,generation,size";
+  static final String OK = "OK";
+  static final String GET_OBJECT_METADATA = "getObjectMetadata";
+  static final String GET_OBJECT_MEDIA = "getObjectMedia";
+  static final String GRPC = "grpc";
+  static final String JSON = "json";
```
medb, line 79: These constants should have more specific names to explain how they are used.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -217,41 +238,72 @@ private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
...
+
+  private void publishErrorMetric(
+      MeasureLong measure, Stopwatch stopwatch, String method, String protocol, Exception e) {
+    long time = stopwatch.elapsed(MILLISECONDS);
+    String error;
+    if (e instanceof StatusRuntimeException) {
+      error = ((StatusRuntimeException) e).getStatus().toString();
+    } else {
+      error = e.getClass().getSimpleName();
+    }
```
medb, line 300: Use ternary operator

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -217,41 +238,72 @@ private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
...
+    long time = stopwatch.elapsed(MILLISECONDS);
+    TagKey[] keys = new TagKey[] {METHOD, STATUS, PROTOCOL};
+    String[] values = new String[] {method, OK, protocol};
+    metricsPublisher.recordLong(keys, values, measure, time);
+    logger.atFinest().log(
+        "method : %s , status : %s, protocol : %s , measure : %s , time : %d",
+        method, OK, protocol, measure, time);
+  }
+
+  private void publishErrorMetric(
```
medb, line 292: These methods are called `publish...` but they call `metricsPublisher.record...` methods. Are they actually publish anything or just record metrics that are published asynchronously?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -144,6 +144,8 @@
 
   private static final String LIST_OBJECT_FIELDS_FORMAT = "items(%s),prefixes,nextPageToken";
 
+  private MetricsPublisher metricsPublisher = new NoOpMetricsPublisher();
```
medb, line 147: Probably better to make it final, and assign no-op implementation if metrics are disabled.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -92,6 +92,9 @@
   /** Default setting for grpc message timeout check interval (in milliseconds) */
   public static final long GRPC_MESSAGE_TIMEOUT_CHECK_INTERVAL = 1000;
 
+  /** Default setting for publishing open-census metrics */
+  public static final boolean METRICS_ENABLED = false;
```
medb, line 96: Probably this should be more specific about where/how metrics are published

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
medb: ping

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
```diff
@@ -0,0 +1,244 @@
...
+      registerAllViews();
+      setupStackDriverExporter(projectId, credentials);
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log("Exception while registering metrics publisher");
+      return new NoOpMetricsPublisher();
+    }
+    return new CloudMonitoringMetricsPublisher();
+  }
+
+  private static void setupStackDriverExporter(String projectId, Credentials credentials)
```
medb, line 141: Rename to `...CloudMonitoring`

pom.xml
medb: `-impl` actually depends on `-api`, so we can remove at least it: https://search.maven.org/artifact/io.opencensus/opencensus-impl/0.31.0/jar

pom.xml
```diff
@@ -391,6 +392,32 @@
...
         <type>pom</type>
       </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-api</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-impl</artifactId>
```
medb, line 402: Why we do not use `-impl-lite` or `-impl-core` dependency? Are there any functionality that is not available in either of those that we need?

pom.xml
```diff
@@ -391,6 +392,32 @@
...
+      </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-contrib-grpc-metrics</artifactId>
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client-jackson2</artifactId>
+        <version>1.20.0</version>
```
medb, line 418: This should not be hardcoded, but made a version var, also this is a very old version of this library.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
mprashanthsagar: There are a lot of fine grain metrics (message level latencies, status of each rpc, retries), some of which don't make sense to be aggregated. 

IMO, Hadoop stats and Opencensus metrics are operating at different levels of abstractions. 
If we consolidate Opencensus metrics and Hadoop Stats, it would be difficult to extend more metrics without co-ordinating between multiple layers, complicating the code-base. Spark metrics [architecture](https://db-blog.web.cern.ch/blog/luca-canali/2019-02-performance-dashboard-apache-spark) does encourage publishing directly to different sinks for metrics which are not viable to be aggregated at the master.

pom.xml
mprashanthsagar: `-impl-lite` [ref](https://www.javadoc.io/doc/io.opencensus/opencensus-impl-lite/latest/io/opencensus/impllite/metrics/MetricsComponentImplLite.html) is android compatible version. 

Official [documentation](https://opencensus.io/quickstart/java/metrics/#0) does recommend `impl` and `api`dependencies. 

Any particular reason to scope it down to -impl-core ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
mprashanthsagar: Along the lines of configuration, couldn't tie it to grpc/cloudmonitoring. Open to suggestions

gcsio/pom.xml
```diff
@@ -146,7 +150,23 @@
...
       <artifactId>gson</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-impl</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-exporter-stats-stackdriver</artifactId>
```
medb, line 159: Why we do not use https://search.maven.org/search?q=g:com.google.cloud.opentelemetry instead?

https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/exporters/metrics/README.md

gcs/pom.xml
medb: Yes, but we use google-api JSON client with Gson, not Jackson. Why do we use Jackson instead of Gson for metrics export?

gcs/pom.xml
medb: I can not find where this dependency declared. May you share dependency chain that brings this in?

medb (changes requested):

gcs/pom.xml
mprashanthsagar: Cloudmonitoring exporter relies on [cloud-core](https://mvnrepository.com/artifact/com.google.cloud/google-cloud-core/2.5.4) which relies on [http-client](https://mvnrepository.com/artifact/com.google.http-client/google-http-client/1.41.4) that has the dependency. I noticed runtime errors because of which i had to add this inclusion

gcs/pom.xml
mprashanthsagar: We need to include cloud-core lib as well, reverting to `com.google.cloud`

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
medb: I do not think that we need to have 2 flags, 1st to enable metrics and 2nd to chose sinks - sinks configuration is sufficient, I think. So we can already change this to `fs.gs.metrics.sink=NONE|CLOUD_MONITORING`. Not sure if we actually want to support multiple sinks at the same time, if no, we can live this as a single-value enum property

pom.xml
medb: Maybe `-impl-lite` or `-impl-core` have functionality that we need and will not pull in extra dependencies that we do not need?

gcsio/pom.xml
medb: I think that it's fine to use alpha library, this is just metrics, and we use alpha GCS gRPC client library anyway - if we know that this is a long term solution, I think that better to use it from get go, instead of migrating to it in the future - it will save us some work. Also, this library may not have dependency on Apache HTTP client that we want to remove from GCS connector 3.0 release

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -339,6 +341,13 @@ public GoogleCloudStorageImpl(
             ? null
             : this.storage.getRequestFactory().getInitializer();
 
+    if (this.storageOptions.isMetricsEnabled()) {
+      this.metricsPublisher =
+          CloudMonitoringMetricsPublisher.create(options.getProjectId(), credentials);
+    } else {
+      this.metricsPublisher = new NoOpMetricsPublisher();
+    }
```
medb, line 349: nit: use ternary operator

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -373,6 +382,18 @@ public GoogleCloudStorageImpl(
...
+      checkArgument(
+          httpRequestInitializer instanceof RetryHttpInitializer,
+          "request initializer must be an instance of the RetryHttpInitializer class"
+              + " when gRPC API enabled");
+      Credentials credentials = ((RetryHttpInitializer) httpRequestInitializer).getCredentials();
+      this.metricsPublisher =
+          CloudMonitoringMetricsPublisher.create(options.getProjectId(), credentials);
+    } else {
+      this.metricsPublisher = new NoOpMetricsPublisher();
+    }
```
medb, line 395: nit: consider using ternary operator, but not sure if it will make code more readable in this specific case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -373,6 +382,18 @@ public GoogleCloudStorageImpl(
             ? null
             : this.storage.getRequestFactory().getInitializer();
 
+    if (this.storageOptions.isMetricsEnabled()) {
+      checkArgument(
```
medb, line 386: Maybe this redundant check as case will fail anyway with quite descriptive exception?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -373,6 +382,18 @@ public GoogleCloudStorageImpl(
...
+          httpRequestInitializer instanceof RetryHttpInitializer,
+          "request initializer must be an instance of the RetryHttpInitializer class"
+              + " when gRPC API enabled");
+      Credentials credentials = ((RetryHttpInitializer) httpRequestInitializer).getCredentials();
```
medb, line 390: This could be inlined when ternary operator is used

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsPublisher.java
medb: Makes sense, thanks for explanation.

pom.xml
```diff
@@ -391,6 +394,36 @@
         <scope>provided</scope>
         <type>pom</type>
       </dependency>
+      <dependency>
+        <groupId>io.opencensus</groupId>
```
medb, line 398: Can we use bom to pin all opencensus dependency versions (see how we do this for grpc deps)?

pom.xml
```diff
@@ -391,6 +394,36 @@
...
+        <version>${opencensus.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-core</artifactId>
+        <version>${google.cloud-core.verion}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.api</groupId>
+        <artifactId>gax</artifactId>
```
medb, line 419: Can we use bom to pin all gax dependency versions and maybe other google api deps (see how we do this for grpc deps)?

medb (changes requested):

pom.xml
mprashanthsagar: I don't think bom is supported from opencensus [open issue](https://github.com/census-instrumentation/opencensus-java/issues/1117). I couldn't find it on mvn central either

pom.xml
mprashanthsagar: With only one dependency, bom is probably overkill. We can move to bom if/when we add more gax deps for consistency.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: It's a multi-line conditional code, ternary operator cannot be used.

gcsio/pom.xml
mprashanthsagar: On second look, grpc-java does support Opencensus metrics out of the box. I couldn't find opentelemetry integration yet. 

I think the migration would be worth the additional client metrics support as all the changes could be done behind metricsRecorder

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: I meant support multiple sinks (not at the same time). We could publish to Cloud-monitoring/ Prometheus/ ..

Ack the changes with single-value enum property

pom.xml
mprashanthsagar: The only extra difference in dependency is the [disruptor](https://mvnrepository.com/artifact/com.lmax/disruptor) framework which seems to improve performance. I am inclined to continue with `-impl` and leverage it

gcs/pom.xml
mprashanthsagar: It isn't a direct dependency. [Google-http-client](https://mvnrepository.com/artifact/com.google.http-client/google-http-client/1.41.4) depends on it

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -340,6 +343,11 @@ public GoogleCloudStorageImpl(
             ? null
             : this.storage.getRequestFactory().getInitializer();
 
+    this.metricsRecorder =
+        MetricsSink.CLOUD_MONITORING.equals(this.storageOptions.getMetricsSink())
```
medb, line 347: In Java usually we use `==` to check equality of enum values.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -371,6 +379,14 @@ public GoogleCloudStorageImpl(
 
     this.httpRequestInitializer = this.storage.getRequestFactory().getInitializer();
 
+    if (MetricsSink.CLOUD_MONITORING.equals(this.storageOptions.getMetricsSink())) {
```
medb, line 382: Same here

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -371,6 +379,14 @@ public GoogleCloudStorageImpl(
...
+      Credentials credentials = ((RetryHttpInitializer) httpRequestInitializer).getCredentials();
```
medb, line 383: Seems like it worth inlining this var to unblock usage of ternary operator instead of `if`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -95,6 +95,11 @@
   /** Default setting for GCS HTTP request headers. */
   public static final ImmutableMap<String, String> HTTP_REQUEST_HEADERS_DEFAULT = ImmutableMap.of();
 
+  public enum MetricsSink {
+    NONE,
+    CLOUD_MONITORING
```
medb, line 100: ni: add trailing comma to allow simpler diff when adding new value

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/MetricsRecorder.java
```diff
@@ -0,0 +1,239 @@
...
+   * @param keys - List of TagKeys attributes used for filtration
+   * @param values - corresponding values for list of TagKeys
+   * @param ml - Measurement to be recorded
+   * @param n - Value of the measurement recorded
+   */
+  void recordLong(TagKey[] keys, String[] values, MeasureLong ml, Long n);
+}
+
+/** Publishes open-census measurements to StackDriver */
+class CloudMonitoringMetricsRecorder implements MetricsRecorder {
```
medb, line 73: Is there a reason to not move it to its own class file?

medb (approved):

PR #713, opened 2022-02-04
[grpc] Refactor footer caching logic

mprashanthsagar:
Re-use zero-copy deserialiser and message level timeouts feature for footer caching as well.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -220,119 +183,39 @@ static GoogleCloudStorageGrpcReadChannel open(
...
-        itemInfo.getSize(),
-        footerOffsetInBytes,
-        footerContent,
-        watchdog,
-        readOptions,
-        backOffFactory);
+    this.gRPCReadMessageTimeout = readOptions.getGrpcReadMessageTimeoutMillis();
+    this.footerStartOffsetInBytes = max(0, (objectSize - prefetchSizeInBytes));
+    int footerSize = Math.toIntExact(min(prefetchSizeInBytes, objectSize));
+    this.footerBuffer = getFooterContent(footerStartOffsetInBytes, footerSize);
```
mayanks, line 215: I think reading the footer (2MB) on `open()` may impact connector performance. It maybe better to pre-fetch footer only when a read is happening in the footer range. 

For example of a full file is being read, the read will start from 0 offset and will continue till the end of the file. In that case pre-fetching footer does not give any advantage. It may infact hamper performance due to switching from a streaming read to fetching data from footerBuffer.

If a parquet file is being read, then the first read operation will happen with an offset in the footer range. In that case we prefetch the footer in that read operation and subsequent reads from footer range can be served from footer buffer.

If some other arbitrary range reads are happening, then again pre-fetching the footer is not going to give us any added advantage. The footer that we pre-fetch may be used only once to read.

mprashanthsagar: > For example of a full file is being read, the read will start from 0 offset and will continue till the end of the file. In that case pre-fetching footer does not give any advantage. It may infact hamper performance due to switching from a streaming read to fetching data from footerBuffer.

I don't think there is any perceivable impact here. Footer prefetch in sequential read would not add any advantage but it won't hinder performance either. Total network calls and bytes read would be same.


 > If a parquet file is being read, then the first read operation will happen with an offset in the footer range. In that case we prefetch the footer in that read operation and subsequent reads from footer range can be served from footer buffer.

This is the ideal case, which we typically hit with 1MB default for footer size, and support configuration to manipulate it 

> If some other arbitrary range reads are happening, then again pre-fetching the footer is not going to give us any added advantage. The footer that we pre-fetch may be used only once to read.

What would seem like Arbitrary range reads are offsets determined based on the footer metadata and query from Spark.

For clarification, the approach for footer pre-fetch is not in contention here, if we do want to explore alternatives, we should get back on the original design. The scope of this change is limited to adding support for current footer prefetch with message level timeouts and zero-copy deserializer

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -191,26 +138,42 @@ static GoogleCloudStorageGrpcReadChannel open(
...
+    validate(itemInfo);
+    this.resourceId = itemInfo.getResourceId();
+    this.objectGeneration = itemInfo.getContentGeneration();
+    this.objectSize = itemInfo.getSize();
+    this.watchdog = watchdog;
+    this.readOptions = readOptions;
+    this.backOffFactory = backOffFactory;
+    this.readStrategy = readOptions.getFadvise();
+    int prefetchSizeInBytes = readOptions.getMinRangeRequestSize() / 2;
+    this.gRPCReadMessageTimeout = readOptions.getGrpcReadMessageTimeoutMillis();
```
medb, line 156: nit: I think better to use `grpc` in var names

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -191,26 +138,42 @@ static GoogleCloudStorageGrpcReadChannel open(
...
-          RetryDeterminer.ALL_ERRORS,
-          IOException.class);
-    } catch (Exception e) {
-      throw new IOException(String.format("Error reading '%s'", resourceId), e);
+    checkArgument(storage != null, "GCS json client cannot be null");
+    this.useZeroCopyMarshaller =
+        ZeroCopyReadinessChecker.isReady() && readOptions.isGrpcReadZeroCopyEnabled();
+    this.stub = stubProvider.newBlockingStub();
+    GoogleCloudStorageItemInfo itemInfo =
+        getObjectMetadata(resourceId, errorExtractor, backOffFactory, storage);
```
medb, line 146: We should skip this, if fail fast option is not enabled

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -128,61 +128,8 @@
...
-      throws IOException {
-    return open(stubProvider, storage, itemInfo, watchdog, readOptions, BackOffFactory.DEFAULT);
-  }
-
   @VisibleForTesting
-  static GoogleCloudStorageGrpcReadChannel open(
+  GoogleCloudStorageGrpcReadChannel(
       StorageStubProvider stubProvider,
       Storage storage,
       ApiErrorExtractor errorExtractor,
```
medb, line 135: This is a static class - we do not need to inject it

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -128,61 +128,8 @@
...
-      GoogleCloudStorageItemInfo itemInfo,
-      Watchdog watchdog,
-      GoogleCloudStorageReadOptions readOptions)
-      throws IOException {
-    return open(stubProvider, storage, itemInfo, watchdog, readOptions, BackOffFactory.DEFAULT);
-  }
-
   @VisibleForTesting
-  static GoogleCloudStorageGrpcReadChannel open(
+  GoogleCloudStorageGrpcReadChannel(
```
medb, line 132: There are multiple constructors with very many parameters - I think it could be better to introduce builder for this class.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Builder with mandatory fields might be cumbersome here. Refactored, the constructer to 5 fields, (which will reduce to 4 once we migrate to grpc metadata apis) which I think is reasonable

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mprashanthsagar: Until we can use footer prefetch call for metadata, We would end up with an additional metadata call. fail fast would not provide an efficient read in this case. 

This is beyond the scope of this PR. will discuss & track this offline

mayanks (approved):

PR #716, opened 2022-02-08
Refactor and cleanup RetryHttpInitializer

medb:


util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -60,119 +58,124 @@
...
-      logger.atFine().log("Encountered an IOException when accessing URL %s", httpRequest.getUrl());
-      return delegateIOExceptionHandler.handleIOException(httpRequest, supportsRetry);
+    request
+        // Request will be retried if server errors (5XX) or I/O errors are encountered.
+        .setNumberOfRetries(options.getMaxRequestRetries())
+        // Set the timeout configurations.
+        .setConnectTimeout(toIntExact(options.getConnectTimeout().toMillis()))
+        .setReadTimeout(toIntExact(options.getReadTimeout().toMillis()))
+        // Supply a new handler for unsuccessful return codes. 401 Unauthorized will be handled by
+        // the Credentials, 410 Gone will be logged, and 5XX will be handled by a backoff handler.
```
mprashanthsagar, line 86: nit: can we move this comment to javadoc of the handler class ?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -60,119 +58,124 @@
...
+        .setUnsuccessfulResponseHandler(new UnsuccessfulResponseHandler(credentials))
+        .setIOExceptionHandler(new IoExceptionHandler());
+
+    HttpHeaders headers = request.getHeaders();
+    if (isNullOrEmpty(headers.getUserAgent()) && !isNullOrEmpty(options.getDefaultUserAgent())) {
```
mprashanthsagar, line 91: Do we need a check on useragent as a header in `options.getHttpHeaders` ?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -60,119 +58,124 @@
...
         // If credentials decides it can handle it, the return code or message indicated something
         // specific to authentication, and no backoff is desired.
         return true;
       }
 
-      if (delegateHandler.handleResponse(request, response, supportsRetry)) {
+      if (delegate.handleResponse(request, response, supportsRetry)) {
         // Otherwise, we defer to the judgement of our internal backoff handler.
         return true;
       }
```
mprashanthsagar, line 155: Should we move this check to top of method to catch frequent cases early ?

medb: I do not think so, we want to log responses regardless of handling and initialize with credentials if necessary first.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
medb: Done.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
medb: Done.

mprashanthsagar (approved):

PR #724, opened 2022-02-09
Decouple get metadata with get data

aalexx-S:
update:
Decouple both use cases by using new request types `ObjectsGetMetadata` and `ObjectsGetData`. Calling `execute` and `executeMedia` on the wrong type will cause exception.

===========================

Make objects get requests set `alt=media` when initializing so that authorization function can correctly scope the access boundary. Unfortunately, the `alt=media` parameter is set in `executeMedia` function, and we have no control over the procedure. We have to explicitly set it before calling the function so that initializeRequest function has enough information to configure the request.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -784,7 +784,12 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() {
 
       @Override
       protected Storage.Objects.Get createRequest() throws IOException {
-        return initializeRequest(super.createRequest(), resourceId.getBucketName());
+        // Creates a storage request that will be used to request media data from GCS.
+        // Sets "alt=media" query parameter to indicate that this is a media request (instead of
+        // a metadata request) so that the request can be initialized properly.
+        Storage.Objects.Get request = super.createRequest();
+        request.set("alt", "media");
```
medb, line 791: I do not think that all requests created by this method are media requests. Also, why do we need to set this manually, I thought it will be auto-set by client library depending on the request?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -107,8 +107,14 @@
   }
 
   private static List<AccessBoundary> translateObjectGetRequest(Storage.Objects.Get request) {
-    return Collections.singletonList(
-        AccessBoundary.create(request.getBucket(), request.getObject(), Action.READ_OBJECTS));
+    if (request.getOrDefault("alt", "").equals("media")) {
+      return Collections.singletonList(
+          AccessBoundary.create(request.getBucket(), request.getObject(), Action.READ_OBJECTS));
+    } else {
```
medb, line 113: nit: this else is redundant

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -107,8 +107,14 @@
...
+      return Collections.singletonList(
```
medb, line 114: Use `ImmutableList.of()`

util/src/main/java/com/google/cloud/hadoop/util/AccessBoundary.java
```diff
@@ -37,6 +37,7 @@ public static AccessBoundary create(String bucketName, String objectName, Action
     // EDIT_OBJECTS implies READ_OBJECTS and WRITE_OBJECTS
     EDIT_OBJECTS,
     DELETE_OBJECTS,
+    GET_OBJECT_METADATA,
```
medb, line 40: nit: sort alphabetically

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: I'll read the code again to confirm the first question. To the second question, we need the `alt=media` parameter for `initializeRequest` to distinguish the meaning of the request. The parameter is only set in `executeMedia()` function, and we it's too late for access token initializing. Unfortunately this means that we can't guarantee that every requests that will eventually call `executeMedia` is initialized with the correct access token.

util/src/main/java/com/google/cloud/hadoop/util/AccessBoundary.java
aalexx-S: done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
aalexx-S: done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
aalexx-S: done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: The only call I'm not sure about is [here](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L234). The comment says it is supposed to get metadata but it calls `execute` instead of `executeMedia`.

medb: In this case I think that we may want something more robust/future proof then setting manually query parameter, that can be overwritten anyway?

One option would be to extend `Get` class with `GetData` and `GetMetdata` that override `execute`/`executeMedia` requests to fail in cases that they are not supported - this will allow to know how request will be executed as soon as its object created.

aalexx-S: Although this doesn't prevent future codes from using the base class directly, I think it's a good start to decouple the actions.

Also, regarding to the code linked above ([this](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L234). Do you know the intent of this call? I can change it and run a few beeline/spark jobs to see if it works, but it would be better if we can for sure know that this is meant to be a metadata call instead.

aalexx-S: Appended a new commit for this change.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -229,6 +230,9 @@ public Boolean load(String bucketName) {
   // GCS access instance.
   @VisibleForTesting Storage storage;
```
medb, line 231: Do we need `storage` when we have `storageRequestFactory` already?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -350,16 +356,33 @@ public GoogleCloudStorageImpl(
...
+   *
+   * @param options {@link GoogleCloudStorageOptions} to use to initialize the object.
+   * @param storage {@link Storage} to use for I/O.
+   * @param downscopedAccessTokenFn Function that generates downscoped access token.
+   */
+  public GoogleCloudStorageImpl(
+      GoogleCloudStorageOptions options,
+      Storage storage,
+      Function<List<AccessBoundary>, String> downscopedAccessTokenFn) {
+    this(options, storage, new StorageRequestFactory(storage), downscopedAccessTokenFn);
```
medb, line 371: Do we really need this constructor?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -350,16 +356,33 @@ public GoogleCloudStorageImpl(
...
    *
    * @param options {@link GoogleCloudStorageOptions} to use to initialize the object.
    * @param storage {@link Storage} to use for I/O.
+   * @param storageRequestFactory creates storage requests that override storage api.
    * @param downscopedAccessTokenFn Function that generates downscoped access token.
    */
   public GoogleCloudStorageImpl(
       GoogleCloudStorageOptions options,
       Storage storage,
+      StorageRequestFactory storageRequestFactory,
```
medb, line 385: Is this just added for testing? If not, we maybe do not need it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -774,7 +799,12 @@ private SeekableByteChannel open(
     }
 
     return new GoogleCloudStorageReadChannel(
-        storage, resourceId, errorExtractor, clientRequestHelper, readOptions) {
+        storage,
+        storageRequestFactory,
```
medb, line 803: If both are not necessary, then let's leave just `storageRequestFactory` and even if both are necessary, then better to have a getter for `storage` in the factory instead of passing both, I think.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
```diff
@@ -46,8 +48,10 @@
       return translateObjectInsertRequest((Storage.Objects.Insert) request);
     } else if (request instanceof Storage.Objects.Compose) {
       return translateObjectComposeRequest((Storage.Objects.Compose) request);
-    } else if (request instanceof Storage.Objects.Get) {
```
medb, line 49: Should we fast-fail on plain `Get` request? Or show it will be treated if we accidentally will start to use it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/storageapioverride/ObjectsGetMedia.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.api.services.storage.Storage;
+import com.google.api.services.storage.Storage.Objects.Get;
+import com.google.api.services.storage.model.StorageObject;
+import com.google.common.annotations.VisibleForTesting;
+
+/**
+ * Overrides Storage.Objects.Get and provides specific functions for getting object media to
+ * decouple getting object media from getting metadata.
+ */
+public class ObjectsGetMedia extends Get {
```
medb, line 12: I think that `Media` could be a quite specific term for GCS JSON API. May you check GCS documentation on how do they refer to data/metadata requests? If they use `data` then I think it will be more clear to use `Data` instead of `Media` here too.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/storageapioverride/StorageRequestFactory.java
```diff
@@ -0,0 +1,30 @@
+package com.google.cloud.hadoop.gcsio.storageapioverride;
```
medb, line 1: It seems like all classes in the `com/google/cloud/hadoop/gcsio/storageapioverride/` package are quite small. I'm thinking that it maybe worth to move them all as static nested classes in the `StorageRequestFactory` class and make `StorageRequestFactory` package-private in the `com/google/cloud/hadoop/gcsio/` package - this will allow us to change it freely without worrying about breaking changes as we would do for public classes.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: `Storage` is still required for creating other request in `Objects` or `Buckets`. Ideally we can have the factory to create all the requests so that we have a unified interface to deal with them.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestToAccessBoundaryConverter.java
medb: If there already fast-fail for this, then it's OK.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: I think we can expose `storage` from `storageRequestFactory` via getter instead?

aalexx-S: Good idea, and it looks like `Storage` instance is only used for creating storage requests. I'll make the change. We can migrate the creation of other requests into the factory latter if we choose to do so, but this will clean it up a bit for now. (Migrating all request creation into it feels like deserve its own PR to me).

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: If we make the change above (remove storage and add getting to get it from factory), I think we can unify the two constructor. Since there is already other an other constructor that accept a credential instead, do you think it is acceptable for now until turning all of these into a builder? I'm imagine something like this: https://softwareengineering.stackexchange.com/questions/289439/how-should-i-handle-incompatible-configurations-with-the-builder-pattern

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: +1, will make the change.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -152,7 +153,12 @@
   @VisibleForTesting protected boolean metadataInitialized = false;
 
   /**
+<<<<<<< HEAD
```
aalexx-S, line 156: oh no

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
aalexx-S: It doesn't make sense to hide Storage now since some classes (GoogleCloudStorageReadChannel, for example) still requires a Storage to construct, and it doesn't make sense to hide it in Impl class, get it out, construct a read channel and hide it again in read channel. It might make relation between those objects a little bit strange.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -368,6 +372,7 @@ public GoogleCloudStorageImpl(
     this.storageOptions.throwIfNotValid();
 
     this.storage = checkNotNull(storage, "storage must not be null");
+    this.storageRequestFactory = new StorageRequestFactory(this.storage);
```
medb, line 375: nit: remove `this.` from storage parameter

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1103,31 +1104,47 @@ private HttpResponse handleExecuteMediaException(IOException e) throws IOExcepti
     throw new IOException(msg, e);
   }
 
-  private Storage.Objects.Get createDataRequest(String rangeHeader) throws IOException {
-    Storage.Objects.Get getObject = createRequest();
+  private Get createDataRequest(String rangeHeader) throws IOException {
```
medb, line 1107: Please revert import and use more descriptive `Storage.Objects.Get` type, instead less obvious `Get`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageRequestFactory.java
```diff
@@ -0,0 +1,87 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Please add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/StorageRequestFactoryTest.java
```diff
@@ -0,0 +1,45 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/StorageRequestFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.services.storage.Storage;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class StorageRequestFactoryTest {
```
medb, line 16: Add short class-level Javadoc

util/src/main/java/com/google/cloud/hadoop/util/AccessBoundary.java
```diff
@@ -30,16 +30,17 @@ public static AccessBoundary create(String bucketName, String objectName, Action
...
     GET_BUCKETS,
-    CREATE_BUCKETS,
-    DELETE_BUCKETS,
-    LIST_BUCKETS
+    GET_METADATA_OBJECTS,
+    LIST_BUCKETS,
+    LIST_OBJECTS,
+    READ_OBJECTS,
+    UNSPECIFIED_ACTION,
+    WRITE_OBJECTS
```
medb, line 44: nit: add trailing comma to minimize diff when adding new actions

medb (changes requested):

medb (approved):

PR #737, opened 2022-02-23
Migrate tests from using Mockito mocks except for mockBatchHelper and related mocks

aalexx-S:
Remove most of the mockito mocks. Mocks for Storage and requests are
completely removed. Only mocks for BatchHelper related objects remains.

Also moves the tests into their repective test class since they don't
use mockito now.


medb (changes requested):
Seems like we lost 0.30% percentage of test coverage. May you check what we do not test anymore?

aalexx-S:
Can't access Codecov's analysis: 
Unauthorized: Github API rate limit error: Forbidden
I'll take a look latter

aalexx-S:
Ref: https://app.codecov.io/gh/GoogleCloudDataproc/hadoop-connectors/compare/737/changes
Diff in `Watchdog`:
- L217-219: I have no idea what causes this :(

Diff in `GoogleCloudStorageImpl`:
- L372-373: Caused by not using mockito object, so `this.storage.getRequestFactory()` is not null anymore.
- L438-439, L448-449, L453-454: Unused package private utilities for injecting mock object.

Diff in `MockHttpTransport`:
- L112-136: Unused test function to obtain response from mock object.

Diff in `GoogleCloudStorageWriteChannel`:
- L89: It's now a real object, so `insert.getMediaHttpUploader()` is not null anymore.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -433,26 +430,11 @@ void setErrorExtractor(ApiErrorExtractor errorExtractor) {
...
 
-  @VisibleForTesting
-  void setSleeper(Sleeper sleeper) {
-    this.sleeper = sleeper;
-  }
-
-  @VisibleForTesting
-  void setBackOffFactory(BackOffFactory factory) {
-    backOffFactory = factory;
-  }
```
medb, line 454: nit: make these fields `final` if possible after removing these setters

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/MockGoogleCloudStorageImplFactory.java
```diff
@@ -0,0 +1,34 @@
+package com.google.cloud.hadoop.gcsio;
+
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.services.storage.Storage;
+import com.google.cloud.hadoop.util.RetryHttpInitializer;
+import com.google.cloud.hadoop.util.RetryHttpInitializerOptions;
+import com.google.cloud.hadoop.util.testing.FakeCredentials;
+
+class MockGoogleCloudStorageImplFactory {
```
medb, line 10: Probably better to add this class to the `testing` package: https://github.com/GoogleCloudDataproc/hadoop-connectors/tree/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing

util/src/main/java/com/google/cloud/hadoop/util/testing/MockHttpTransportHelper.java
```diff
@@ -107,35 +104,6 @@ public String getErrorDomain() {
...
-          }
-        };
-    HttpRequest request =
-        transport
-            .createRequestFactory()
-            .buildGetRequest(HttpTesting.SIMPLE_GENERIC_URL)
-            .setResponseReturnRawInputStream(true);
-    return request.execute();
-  }
-
```
medb, line 138: Hm, I think that it maybe better to not delete this code if it could be used for other tests? Or it's useful only for Mockito tests?

medb (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/MockGoogleCloudStorageImplFactory.java
aalexx-S: It uses `com.google.cloud.hadoop.util.testing.FakeCredentials`, which is under `testing` source root. We need to move both of them if we want to move it into `main` source root.

util/src/main/java/com/google/cloud/hadoop/util/testing/MockHttpTransportHelper.java
aalexx-S: I think the new added arbitrary behavior simulation method can achieve the same thing. This basically obtain response at runtime from the underlying input stream to allow inject custom input steam behavior (like throwing exception, returning arbitrary content from input stream, etc).

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/MockGoogleCloudStorageImplFactory.java
aalexx-S: Or maybe lets move it into the same testing package but under `testing` source root.

medb (approved):

PR #750, opened 2022-03-14
Add metrics to GCSIO apis

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -469,7 +484,21 @@ public GoogleCloudStorageOptions getOptions() {
   @Override
   public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOptions options)
       throws IOException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
     logger.atFiner().log("create(%s)", resourceId);
+    try {
+      AbstractGoogleAsyncWriteChannel<?> channel = createInternal(resourceId, options);
+      recordSuccessMetric(metricsRecorder, LATENCY_MS, stopwatch, "create", PROTOCOL_JSON);
```
medb, line 491: probably `stopwatch` should be stopped ASAP in the parent methods to improve accuracy.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -596,6 +638,20 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
   @Override
   public void createEmptyObject(StorageResourceId resourceId, CreateObjectOptions options)
       throws IOException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
```
medb, line 641: Do we really need to measure each method, if we can measure HTTP requests latency as we do for Hadoop Metrics?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Makes sense, it's cleaner that way as well

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Right now, we are not sure if we are spending time only a single HTTP calls or Batch/Higher level Op which involves multiple HTTP calls. This might be an overkill at this point. We will remove some/many of them after the debugging exercise 

I will be adding metrics for JSON apis at http layer as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -144,6 +150,16 @@
           "metadata");
 
   private static final String LIST_OBJECT_FIELDS_FORMAT = "items(%s),prefixes,nextPageToken";
+  private static final String METHOD_CREATE_BUCKET = "createBucket";
```
medb, line 153: Each of these constants used in a single method - not sure that it worth defining class-level constants for this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -16,6 +16,10 @@
 
 package com.google.cloud.hadoop.gcsio;
 
+import static com.google.cloud.hadoop.gcsio.CloudMonitoringMetricsRecorder.LATENCY_MS;
+import static com.google.cloud.hadoop.gcsio.CloudMonitoringMetricsRecorder.PROTOCOL_JSON;
+import static com.google.cloud.hadoop.gcsio.CloudMonitoringMetricsRecorder.recordErrorMetric;
+import static com.google.cloud.hadoop.gcsio.CloudMonitoringMetricsRecorder.recordSuccessMetric;
```
medb, line 22: Why is this static method? I think it should be instance method on of the `CloudMonitoringMetricsRecorder` instance.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: For instance method, we would need to move it to `MetricsRecorder` interface. Success and error metrics seemed too specific to add instance methods. Since their functionalities were on top of actual metric interfaces, I extracted them as util methods to `CloudMonitoringMetricsRecorder` (for lack of a better place)

We could still expose these methods as instance methods from `MetricsRecorder`, would you prefer that ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: Fair enough, makes sense.  I will revert them

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: If they are too specific for all types of metric syncs/recorders, then maybe we abstracted interface too soon? And should stick just with an impl class for just Cloud Monitoring sync first and extract common interface when we add more syncs and it will be more clear how common interface could look like? Also, I think that this defeats point of having `MetricsRecorder` interface, as we will need to wire similar static helper methods for other metric syncs in the future, while we should strive to opposite - ability to implement new syncs without need to do any changes at the call sites of `MetricsRecorder` interface and just add new implementation of the `MetricsRecorder` interface.

Alternative will be to create some wrapper class that takes `CloudMonitoringMetricsRecorder` instance as a constructor parameter so we can have instance helper methods for it that do not require passing `CloudMonitoringMetricsRecorder` instance for each method call - this seems to be more OOP-like, but still not as good as having these methods in the `CloudMonitoringMetricsRecorder` directly, if possible.

medb (changes requested):
(Replied on the pending comment)

PR #751, opened 2022-03-15
Enable HTTP trace event logs to Cloud Logging

mprashanthsagar:
Write structured logs to Cloud logging, to enable low level analysis via [log analytics preview](https://cloud.google.com/logging/docs/log-analytics)

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -27,6 +27,7 @@ import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.GlobAlgorithm;
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.OutputStreamType;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystemOptions;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions;
+import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.EventLogSink;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.MetricsSink;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions.Fadvise;
@@ -419,6 +420,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<MetricsSink> GCS_METRICS_SINK =
       new HadoopConfigurationProperty<>("fs.gs.metrics.sink", MetricsSink.NONE);
 
+  /** Configuration key to enable publishing event logs to Google cloud logging. */
+  public static final HadoopConfigurationProperty<EventLogSink> GCS_EVENT_LOG_SINK =
+      new HadoopConfigurationProperty<>("fs.gs.eventlog.sink", EventLogSink.NONE);
+
   // TODO(b/120887495): This @VisibleForTesting annotation was being ignored by prod code.
   // Please check that removing it is correct, and remove this comment along with it.
   // @VisibleForTesting
@@ -473,7 +478,8 @@ public class GoogleHadoopFileSystemConfiguration {
             GCS_GRPC_CHECK_INTERVAL_TIMEOUT_MS.get(config, config::getLong))
         .setDirectPathPreferred(GCS_GRPC_DIRECTPATH_ENABLE.get(config, config::getBoolean))
         .setTrafficDirectorEnabled(GCS_GRPC_TRAFFICDIRECTOR_ENABLE.get(config, config::getBoolean))
-        .setMetricsSink(GCS_METRICS_SINK.get(config, config::getEnum));
+        .setMetricsSink(GCS_METRICS_SINK.get(config, config::getEnum))
+        .setEventLogSink(GCS_EVENT_LOG_SINK.get(config, config::getEnum));
   }
 
   private static PerformanceCachingGoogleCloudStorageOptions getPerformanceCachingOptions(
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -27,8 +27,8 @@ import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.GlobAlgorithm;
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem.OutputStreamType;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystemOptions;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions;
-import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.EventLogSink;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.MetricsSink;
+import com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.TraceLogSink;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions.Fadvise;
 import com.google.cloud.hadoop.gcsio.PerformanceCachingGoogleCloudStorageOptions;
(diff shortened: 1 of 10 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -419,6 +420,10 @@
   public static final HadoopConfigurationProperty<MetricsSink> GCS_METRICS_SINK =
       new HadoopConfigurationProperty<>("fs.gs.metrics.sink", MetricsSink.NONE);
 
+  /** Configuration key to enable publishing event logs to Google cloud logging. */
+  public static final HadoopConfigurationProperty<EventLogSink> GCS_EVENT_LOG_SINK =
+      new HadoopConfigurationProperty<>("fs.gs.eventlog.sink", EventLogSink.NONE);
```
mayanks, line 425: nit: `event-log`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,69 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+
+  private final Map<HttpRequest, Stopwatch> requestTracker = new ConcurrentHashMap<>();
```
mayanks, line 24: requests that timeout will never receive a response and hence they will be stuck in this map for ever. Do you think we should clean this map after some timeout?

mayanks: One option is to set a custom request-header and populate it with current timestamp. In the `log` method then read that header and diff it with the current timestamp.

mprashanthsagar: The requests with timeouts should still be captured here. I don't follow the use-case here.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
mprashanthsagar: I am not particularly a fan of introducing `-` in flags/configurations.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.base.Stopwatch;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
```
medb, line 22: "EventLog" could be a bad name as this is what Spark uses for its history logs too.

gcsio/pom.xml
```diff
@@ -166,6 +166,10 @@
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-core</artifactId>
     </dependency>
+    <dependency>
```
medb, line 169: Why do we need to write directly to Cloud Logging instead of writing to log the same to local file which automatically pushed to Cloud Logging by fluentd?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
mprashanthsagar: Good call.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
mayanks: requests which timeout will never receive a HTTP response and hence the log method will not be called. Those request I guess will rot in this map.

mprashanthsagar: both client and server side timeouts should be captured by this interceptor. could you point me to docs/ why would you think timeout errors are skipped ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -333,10 +337,24 @@ public GoogleCloudStorageImpl(
 
     this.storageOptions = checkNotNull(options, "options must not be null");
     this.storageOptions.throwIfNotValid();
+    List<HttpRequestInitializer> httpRequestInitializers = new ArrayList<>();
+    httpRequestInitializers.add(new StatisticsTrackingHttpRequestInitializer(statistics));
+    httpRequestInitializers.add(
+        new RetryHttpInitializer(credentials, options.toRetryHttpInitializerOptions()));
+    if (storageOptions.getTraceLogSink() == TraceLogSink.CLOUD_LOGGING) {
+      checkNotNull(credentials, "credentials cannot be null when event logging enabled");
```
medb, line 345: `credentials` could be null for unauthenticated access.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -333,10 +337,24 @@ public GoogleCloudStorageImpl(
...
+      this.cloudLogger =
+          LoggingOptions.newBuilder()
+              .setProjectId(options.getProjectId())
+              .setCredentials(credentials)
```
medb, line 349: Does this setter has a null check?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2065,6 +2084,17 @@ public void close() {
     }
   }
 
+  private void closeCloudLogging() {
+    if (cloudLogger != null) {
+      try {
+        cloudLogger.flush();
```
medb, line 2090: This is synchronous operation that can increase shutdown time. Also, does it have timeout?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2065,6 +2084,17 @@ public void close() {
...
+        cloudLogger.close();
+      } catch (Exception e) {
+        logger.atWarning().withCause(e).log("Exception while closing cloud logging instance ");
```
medb, line 2093: Remove extra space at the end of message

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TraceLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,69 @@
...
+    jsonMap.put("protocol", "json");
+    LogEntry logEntry =
+        LogEntry.newBuilder(JsonPayload.of(jsonMap))
+            .setSeverity(Severity.INFO)
+            .setLogName("dataproc-gcs-performance")
+            .setResource(
+                MonitoredResource.newBuilder("global").addLabel("library", "gcsio").build())
+            .build();
+
+    logging.write(Collections.singleton(logEntry));
```
medb, line 67: Making this request synchronous will impact job execution, also what kind of QPS it can handle?

gcsio/pom.xml
medb: What queries are we talking about?

[Flogger supports structured logging](https://github.com/google/flogger/blob/11fcf148ca74dc098b225fff01ad69a40f5a4dd9/api/src/main/java/com/google/common/flogger/LoggingApi.java#L212-L260) which can be parsed by fluentd as tags/json fields if necessary.

Overall I'm not sure that maintenance cost and provided benefit worth native integration with Cloud Logging APIs.

I think that it could be better how to make it work on querying site, instead of implementing custom logging backend.

Additionally, I thought that for tracing are usually used specialized systems, not logs, like https://opentelemetry.io that we maybe already integrated with?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TraceLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @Override
+  public void initialize(HttpRequest request) throws IOException {
+    request.setInterceptor(
+        httpRequest -> requestTracker.put(httpRequest, Stopwatch.createStarted()));
+    request.setResponseInterceptor(this::log);
+  }
+
+  private void log(HttpResponse httpResponse) {
```
medb, line 38: Google API client library has already very low-level trace logging of executed request, did we explore if they are suitable for debugging purposes?

medb (changes requested):

PR #752, opened 2022-03-18
Clean up, fix up and refactor IO streams and channels

medb:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -17,202 +17,156 @@
...
-          byte b = singleReadBuf[0];
-          totalBytesRead++;
-          statistics.incrementBytesRead(1);
-          statistics.incrementReadOps(1);
-          result = (b & 0xff);
+      int numRead = channel.read(ByteBuffer.wrap(singleReadBuf));
+      if (numRead == -1) {
+        response = -1;
+      } else {
+        if (numRead != 1) {
```
mprashanthsagar, line 98: Can we collapse with ` else if ` instead

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -17,202 +17,156 @@
...
   public synchronized void seek(long pos) throws IOException {
+    logger.atFiner().log("seek(%d)", pos);
     long curPos = getPos();
     long diff = pos - curPos;
     if (diff > 0) {
       streamStatistics.seekForwards(diff);
     } else {
       streamStatistics.seekBackwards(diff);
     }
-    synchronized (this) {
```
mprashanthsagar, line 159: Any particular reason we are removing the synchronized block ( I am unaware of why it was there in the first place )

medb: This block is already in `synchronized` method, so this block is redundant.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
medb: Done.

mprashanthsagar (approved):

PR #753, opened 2022-03-23
Add test with increased batch size

mprashanthsagar:
Setup an allowlisted bucket with increased batch size to 5000. Adding an integration test to capture any regressions.
Creating 5K files on every test run is very costly. Files have been created outside the test.

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
@@ -647,7 +647,7 @@ public class GoogleCloudStorageFileSystemNewIntegrationTest {
   }
 
   @Test
-  public void listFileInfo_directory_increased_batch_size() throws Exception {
+  public void listFileInfo_directory_increased_page_size() throws Exception {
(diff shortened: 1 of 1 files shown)


mayanks (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
```diff
@@ -646,6 +646,29 @@ public void listFileInfo_directory() throws Exception {
             bucketUri.resolve(dirObject + "/file1"), bucketUri.resolve(dirObject + "/file2"));
   }
 
+  @Test
+  public void listFileInfo_directory_increased_batch_size() throws Exception {
```
medb, line 650: This seems to be list API result page size, not batch

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
mprashanthsagar: Ah my bad, didn't realise batch was used in another context. I do mean `pageSize` here

PR #755, opened 2022-04-08
Optimise metadata info to ReadChannel

mprashanthsagar:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -783,6 +772,25 @@ public SeekableByteChannel open(
       info = null;
     }
 
+    if (storageOptions.isGrpcEnabled()) {
+      if (info != null) {
```
medb, line 776: nit: use ternary operator.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -783,6 +772,25 @@ public SeekableByteChannel open(
...
+        return new GoogleCloudStorageGrpcReadChannel(
+            storageStubProvider,
+            info,
```
medb, line 779: Maybe you can make it nullable and check in constructor if info is null? This will allow to reduce number of constructors in the gRPC channel.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
mprashanthsagar: This would mean, the constructor would accept (StorageResourceId, Storage, and itemInfo) which seems redundant. 

If itemInfo is null, -- StorageResourceId and Storage cannot be nullable
If StorageResourceId and Storage are null -- then itemInfo cannot be null.

I would prefer to separate them

mayanks (approved):

PR #762, opened 2022-04-12
Run all gRPC integration tests also with TD enabled

Deependra-Patel:
This is needed to test with "forced direct path" endpoint that would actually fail the GoogleCloudStorageGrpcIntegrationTest#testOpenWithTDEnabled test otherwise it may have used the fallback

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
@@ -18,6 +18,7 @@ import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions.Fadvise;
 import com.google.cloud.hadoop.gcsio.StorageResourceId;
 import com.google.cloud.hadoop.gcsio.integration.GoogleCloudStorageTestHelper.TestBucketHelper;
 import com.google.cloud.hadoop.util.AsyncWriteChannelOptions;
+import com.google.common.flogger.GoogleLogger;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
@@ -28,6 +29,8 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class GoogleCloudStorageGrpcIntegrationTest {
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
 
   // Prefix this name with the prefix used in other gcs io integrate tests once it's whitelisted by
   // GCS to access gRPC API.
@@ -129,18 +132,14 @@ public class GoogleCloudStorageGrpcIntegrationTest {
 
   @Test
   public void testOpenWithTDEnabled() throws IOException {
-    String storageServicePath = System.getenv("GCS_TEST_STORAGE_SERVICE_PATH");
     String grpcServerAddress = System.getenv("GCS_TEST_GRPC_SERVER_ADDRESS");
-    String storageRootUrl = System.getenv("GCS_TEST_STORAGE_ROOT_URL");
-    System.err.println(
-        "Printing env vars " + storageServicePath + grpcServerAddress + storageRootUrl);
+    logger.atInfo().log("Printing env var for grpcServerAddress: " + grpcServerAddress);
+
     GoogleCloudStorageOptions.Builder optionsBuilder =
         GoogleCloudStorageTestHelper.getStandardOptionBuilder()
             .setGrpcEnabled(true)
             .setTrafficDirectorEnabled(true);
-    if (storageServicePath != null) optionsBuilder.setStorageServicePath(storageServicePath);
     if (grpcServerAddress != null) optionsBuilder.setGrpcServerAddress(grpcServerAddress);
-    if (storageRootUrl != null) optionsBuilder.setStorageRootUrl(storageRootUrl);
 
     GoogleCloudStorage rawStorage =
         new GoogleCloudStorageImpl(
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
@@ -59,7 +59,7 @@ public class GoogleCloudStorageGrpcIntegrationTest {
   private static GoogleCloudStorageOptions.Builder configureDefaultOptions() {
     GoogleCloudStorageOptions.Builder optionsBuilder =
         GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true);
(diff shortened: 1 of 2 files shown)


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -129,13 +132,18 @@ public void testOpen() throws IOException {
...
+    String grpcServerAddress = System.getenv("GCS_TEST_GRPC_SERVER_ADDRESS");
+    logger.atInfo().log("Printing env var for grpcServerAddress: " + grpcServerAddress);
+
+    GoogleCloudStorageOptions.Builder optionsBuilder =
         GoogleCloudStorageTestHelper.getStandardOptionBuilder()
             .setGrpcEnabled(true)
-            .setTrafficDirectorEnabled(true)
-            .build();
+            .setTrafficDirectorEnabled(true);
+    if (grpcServerAddress != null) optionsBuilder.setGrpcServerAddress(grpcServerAddress);
```
mprashanthsagar, line 142: This isn't ideal. We seem to test two different scenarios in one test.

we should relook at the structure of the test

Deependra-Patel: I have restructured it now, so each test would run twice with TD enabled/disabled. Let me know if you have any questions and we can discuss offline.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -37,31 +44,58 @@ public class GoogleCloudStorageGrpcIntegrationTest {
...
         GoogleCloudStorageTestHelper.getCredentials());
   }
 
   @BeforeClass
   public static void createBuckets() throws IOException {
-    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    GoogleCloudStorage rawStorage =
+        new GoogleCloudStorageImpl(
+            GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true).build(),
+            GoogleCloudStorageTestHelper.getCredentials());
```
mprashanthsagar, line 89: Do we need this change ?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -37,31 +44,58 @@ public class GoogleCloudStorageGrpcIntegrationTest {
...
     rawStorage.createBucket(BUCKET_NAME);
   }
 
   @AfterClass
   public static void cleanupBuckets() throws IOException {
-    GoogleCloudStorage rawStorage = createGoogleCloudStorage();
+    GoogleCloudStorage rawStorage =
+        new GoogleCloudStorageImpl(
+            GoogleCloudStorageTestHelper.getStandardOptionBuilder().setGrpcEnabled(true).build(),
+            GoogleCloudStorageTestHelper.getCredentials());
```
mprashanthsagar, line 98: Do we need this change ? Let's reuse the changes in the `createGoogleCloudStorage` method

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -37,31 +44,58 @@ public class GoogleCloudStorageGrpcIntegrationTest {
...
+  public GoogleCloudStorageGrpcIntegrationTest(boolean tdEnabled) {
+    this.tdEnabled = tdEnabled;
+  }
+
+  private GoogleCloudStorageOptions.Builder configureOptions() {
+    GoogleCloudStorageOptions.Builder optionsBuilder =
+        GoogleCloudStorageTestHelper.getStandardOptionBuilder()
+            .setGrpcEnabled(true)
+            .setTrafficDirectorEnabled(this.tdEnabled);
+    boolean runningTestsFromGCP = Boolean.parseBoolean(System.getenv("RUNNING_TESTS_FROM_GCP_VM"));
```
mprashanthsagar, line 64: Instead of accepting a boolean and hardcoding the endpoint, can we accept endpoint as a configuration with a default value?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -37,31 +44,58 @@ public class GoogleCloudStorageGrpcIntegrationTest {
...
 
-  private static GoogleCloudStorage createGoogleCloudStorage() throws IOException {
+  private final boolean tdEnabled;
+
+  @Parameters
+  // We want to test this entire class with both gRPC-LB and Traffic Director
+  public static Iterable<Boolean> tdEnabled() {
+    return Arrays.asList(false, true);
+  }
+
```
mprashanthsagar, line 54: This change covers testing with gslb and td with fallback. -- Since the env is not set as part of presubmit.

Effectively in one environment, we are testing TD with or without fallback, but never both. Can we cover all scenarios ?

mprashanthsagar (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
Deependra-Patel: Yes, these are needed because @BeforeClass method has to be static while `createGoogleCloudStorage` is no longer static because of using the class fields.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
Deependra-Patel: Setting TD should use direct path (not fallback) wherever it can, so it is not always with fallback. So if you run from a VM directpath should be used, it is just we will not be sure if fallback is used. That's when force-dp come into picture, force-dp was not accessible from where the presubmits are run. VPC-SC tests would have force-dp so we would know there. I am not sure how to make presubmits run with force-dp, and that is out of the scope of this PR because this PR makes sure td enabled/disabled runs for customer or not.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
Deependra-Patel: Refactored this now to have common endpoint logic for everyone. So when we want to run tests on pre-prod for example buckets will also be created in same universe.

mayanks (approved):

PR #765, opened 2022-04-14
shading new dependencies

davidrabinowitz:


gcs/pom.xml
```diff
@@ -286,15 +286,19 @@
                     <include>com.google.api.**</include>
                     <include>com.google.api.gax.**</include>
                     <include>com.google.auth.**</include>
+                    <include>com.google.cloud.*</include>
```
medb, line 289: Do we need this if we have subpackage-specific inlcudes?

util-hadoop/src/test/java/com/google/cloud/hadoop/util/HadoopCredentialConfigurationTest.java
```diff
@@ -132,6 +133,7 @@ public void exceptionIsThrownForNoCredentialOptions() {
   }
 
   @Test
+  @Ignore
```
medb, line 136: Probably we should not do this, because it can hide failures in the future.

medb (approved):

PR #767, opened 2022-04-18
Add env config for integ test to only run with TD enabled

Deependra-Patel:
Some of our internal endpoints don't work with gRPC-LB (which is a fallback option for us)

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
@@ -24,6 +24,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SeekableByteChannel;
 import java.util.Arrays;
+import java.util.List;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -48,8 +49,11 @@ public class GoogleCloudStorageGrpcIntegrationTest {
 
   @Parameters
   // We want to test this entire class with both gRPC-LB and Traffic Director
+  // Some of our internal endpoints only work with TD
   public static Iterable<Boolean> tdEnabled() {
-    return Arrays.asList(false, true);
+    return Boolean.parseBoolean(System.getenv("GCS_TEST_ONLY_RUN_WITH_TD_ENABLED"))
```
mprashanthsagar, line 54: Looks like the intent is to run for `directpath` not TD

Can we modify the env var to `ENFORCE_DIRECT_PATH` ? and add a comment that enforcing direct path with internal endpoint is only supported on TD

Deependra-Patel: `ENFORCE_DIRECT_PATH` is one of the usecase we are facing which is independent of what we are testing here. This piece of code only specifies whether we want to test with tdEnabled or not.
I have added the comment in our google3 codebase where we use this enforce endpoint.

mprashanthsagar: What is the use-case to test TD alone?

mprashanthsagar (approved):

PR #769, opened 2022-04-20
Handle overflow of test bucketName and reduce dependency on username

vinayakumarb:
https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java#L257

`makeBucketName()` considers `username` as well while constructing unique prefix for bucketname. While its limited to only  9 chars, there is no check to limit overall bucketname to 63 chars.

This PR adds a check and following changes to keep same behavior irrespective of usernames' length in bucket names.

1. Generate the unique string of length 13 chars for the bucket consisting of 
     1. Use first 8 characters of username in bucketname. If username is shorter, use complete username. i.e. No padding.
     2. Out of 12 char, after username (upto 8 chars) remaining are chosen from UUID. (min 4 chars)
4.  `username` and `UUID` together makes upto 15 characters including delimiters.
5. Remaining 48 characters can be utilized by `bucketPrefix` and `suffix` provided by the tests. So keep the cumulative length of `bucketPrefix` and `suffix` provided in `getUniqueBucketName(suffix)` to 48 characters. 

Reason to restrict cumulative instead of individual prefix and suffix : There are existing tests, where suffix of upto 34 chars is used in which provided prefix is only 10 chars. In other tests provided prefix is of bigger size and suffix is small.

In this way, behavior of test results doesnot depend on length of usernames.

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
@@ -246,22 +246,44 @@ public class GoogleCloudStorageTestHelper {
     private final String bucketPrefix;
     private final String uniqueBucketPrefix;
 
+    /**
+     * Create a Helper to handle bucket with specified prefix. <br>
+     * A Unique string of 13 chars will be generated as below.<br>
+     * [Max 8 chars of username]_[min 4 chars of random UUID]. Unique prefix will be created by
+     * appending this unique string to provided bucketPrefix.
+     *
+     * @param bucketPrefix Prefix for bucket name. Cumulative Max length of bucketPrefix and suffix
+     *     provided in getUniqueBucketName(suffix) will be 48 chars.
+     */
     public TestBucketHelper(String bucketPrefix) {
       this.bucketPrefix = bucketPrefix + DELIMITER;
-      this.uniqueBucketPrefix = makeBucketName(bucketPrefix);
+      this.uniqueBucketPrefix = makeUniqueBucketNamePrefix(bucketPrefix);
       checkState(
           this.uniqueBucketPrefix.startsWith(this.bucketPrefix),
           "uniqueBucketPrefix should start with bucketPrefix");
     }
 
-    private static String makeBucketName(String prefix) {
+    private static String makeUniqueBucketNamePrefix(String prefix) {
       String username = System.getProperty("user.name", "unknown").replaceAll("[-.]", "");
-      username = username.substring(0, min(username.length(), 9));
-      String uuidSuffix = UUID.randomUUID().toString().substring(0, 6);
+      // Total 13 characters of unique string will be created.
+      // Maximum only first 8 characters of username is considered. Remaining characters will be
+      // taken from UUID.
+      // Leaving room of 48 characters for bucketPrefix and suffix (excluding delimiters)
+      int usernamePrefixLen = min(username.length(), 8);
+      username = username.substring(0, usernamePrefixLen);
+      String uuidSuffix = UUID.randomUUID().toString().substring(0, 12 - usernamePrefixLen);
       return prefix + DELIMITER + username + DELIMITER + uuidSuffix;
     }
 
+    /**
+     * @param suffix Suffix to be added to the bucket name. Max cumulative length of bucketPrefix
+     *     and suffix is 48 chars.
+     * @return An unique bucket name with specified prefix and suffix.
+     */
     public String getUniqueBucketName(String suffix) {
+      checkArgument(
+          bucketPrefix.length() + suffix.length() <= 48,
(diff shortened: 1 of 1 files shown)


vinayakumarb:
> @vinayakumarb Could you please fix the check style violations ?
> 
> ```
> Run 'mvn spotless:apply' to fix these violations.
> ```

Have applied this and pushed changes again.
 Please review.

Deependra-Patel:
@vinayakumarb Tests are failing with error . PTAL.

"[ERROR] com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystemIntegrationTest  Time elapsed: 1.114 s  <<< ERROR!
java.lang.IllegalStateException: uniqueBucketPrefix should start with bucketPrefix"

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
```diff
@@ -256,12 +264,23 @@ public TestBucketHelper(String bucketPrefix) {
...
     private static String makeBucketName(String prefix) {
       String username = System.getProperty("user.name", "unknown").replaceAll("[-.]", "");
-      username = username.substring(0, min(username.length(), 9));
+      // Only first 6 characters of username is considered.
+      username = username.substring(0, min(username.length(), 6));
       String uuidSuffix = UUID.randomUUID().toString().substring(0, 6);
-      return prefix + DELIMITER + username + DELIMITER + uuidSuffix;
+      // [6 char username] + DELIMITER + [6 char UUID] => 13 chars of unique Id
+      // Leaving room of 48 characters for bucketPrefix and suffix (excluding delimiters)
+      return prefix + username + DELIMITER + uuidSuffix;
```
Deependra-Patel, line 272: Let's keep in same order

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
```diff
@@ -256,12 +264,23 @@ public TestBucketHelper(String bucketPrefix) {
 
     private static String makeBucketName(String prefix) {
       String username = System.getProperty("user.name", "unknown").replaceAll("[-.]", "");
-      username = username.substring(0, min(username.length(), 9));
+      // Only first 6 characters of username is considered.
+      username = username.substring(0, min(username.length(), 6));
```
Deependra-Patel, line 268: Can we keep here 9 (becuase if we take only 6 then there could be confusion like deepan(for deepanshu) and deepen(for deependra)) as before and change the 48 check below to lower number?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
vinayakumarb: It is in same order. Its just that `prefix` already have DELIMITER at the end. So duplication is removed.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
vinayakumarb: Considering upto 8 chars of username and remaining are being filled with unique UUID.
There may be cases all 8 characters are same for 2 different users, But UUID ensures prefix is unique.

vinayakumarb:
> 

I think this is fixed now.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
```diff
@@ -246,22 +246,44 @@ public static class TestBucketHelper {
...
       return prefix + DELIMITER + username + DELIMITER + uuidSuffix;
     }
 
+    /**
+     * @param suffix Suffix to be added to the bucket name. Max cumulative length of bucketPrefix
+     *     and suffix is 48 chars.
+     * @return An unique bucket name with specified prefix and suffix.
+     */
     public String getUniqueBucketName(String suffix) {
+      checkArgument(
```
Deependra-Patel, line 284: Checking total length of `uniqueBucketPrefix + DELIMETER + suffix` < 63 would have also worked.

Deependra-Patel (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTestHelper.java
vinayakumarb: I agree that this check serves the purpose. But again actual size of `uniqueBucketPrefix` depends on username of running env. Not fixed.
So checking on `bucketPrefix` and `suffix` which are controlled in code is better.

PR #772, opened 2022-04-28
Handle dropping of buffered data predictably

mayanks:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -252,24 +256,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
-          dataChunkMap.put(writeOffset, data);
           if (dataChunkMap.size() >= channelOptions.getNumberOfBufferedRequests()) {
-            dataChunkMap.remove(dataChunkMap.firstKey());
+            /*
+             If dataChunkMap is full, get the committedWriteOffset. This will make a API call to the
+             server and add latency in this path/context. The upload is happening asynchronously in
+             a different context. So this API will not add latency to overall upload. It will only
+             throttle the upstream write call, which is fine.
+            */
+            committedWriteOffset = getCommittedWriteSizeWithRetries(uploadId);
```
mprashanthsagar, line 271: I dont think this comment is accurate, this thread is responsible for uploads to gcs, (ref line270 -- `requestStreamObserver.onNext(insertRequest);` )

This is going to add significant latency to writes beyond the buffer capacity.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -252,24 +256,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+            logger.atFinest().log(
+                "Fetched committedWriteOffset: size:%d, numBuffers:%d, writeOffset:%d, committedWriteOffset:%d",
+                dataChunkMap.size(),
+                channelOptions.getNumberOfBufferedRequests(),
+                this.writeOffset,
+                committedWriteOffset);
+
+            // check and remove chunks from dataChunkMap
+            while (dataChunkMap.size() > 0 && dataChunkMap.firstKey() < committedWriteOffset) {
```
mprashanthsagar, line 280: Can we add additional tests to verify the change ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -252,24 +256,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+              logger.atFinest().log(
+                  "clearing dataChunkMap one buffer at a time, size: %d, firstKey:%d, committedwriteOffset:%d",
+                  dataChunkMap.size(), dataChunkMap.firstKey(), committedWriteOffset);
+              dataChunkMap.remove(dataChunkMap.firstKey());
```
mprashanthsagar, line 284: If we have a commit offset, why not delete all data upto that offset, instead of just the first key ?

This is add additional calls to server, and sub-optimal buffer clear performance

mprashanthsagar:
I am not in agreement with the changes in the PR, I suspect we are adding latency in the happy path in an attempt to optimise failure scenario handling. 

Can we re-look at the design offline before we get to implementation ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mprashanthsagar: nvm, misred while loop for if

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mprashanthsagar: what happens if the commit offset >= first key , potentially with offset .

we can end up with uncontrolled buffer size

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: Updated the comment. We now have data backed analysis on why this is not adding any latency to uploads.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -70,10 +71,12 @@
 
   private static final Duration START_RESUMABLE_WRITE_TIMEOUT = Duration.ofMinutes(1);
   private static final Duration QUERY_WRITE_STATUS_TIMEOUT = Duration.ofMinutes(1);
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
```
medb, line 74: nit: logger usually is a first class field declared even before constants

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -205,23 +208,23 @@ public WriteObjectResponse call() throws IOException {
...
+      Status.Code statusCode = status.getCode();
+      return (TRANSIENT_ERRORS.contains(statusCode));
     }
 
     private WriteObjectResponse doResumableUpload() throws IOException {
+      long committedWriteOffset = 0;
       // Only request committed size for the first insert request.
       if (writeOffset > 0) {
+        /*
+        TODO: writeOffset has to be multiple of MAX_BYTES_PER_MESSAGE. If not, we cannot upload
```
mprashanthsagar, line 222: please create a bug for the todo

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -205,23 +208,23 @@ public WriteObjectResponse call() throws IOException {
...
-      Throwable cause = throwable.getCause();
-      if (cause == null) return true;
-      return isRetriableError(cause);
+      Status status = Status.fromThrowable(throwable);
+      Status.Code statusCode = status.getCode();
+      return (TRANSIENT_ERRORS.contains(statusCode));
     }
 
     private WriteObjectResponse doResumableUpload() throws IOException {
+      long committedWriteOffset = 0;
```
mprashanthsagar, line 218: Why do we need both `writeOffset` and `committedWriteOffset` ? We should only rely on the offset from the server

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -252,24 +255,52 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
-            dataChunkMap.remove(dataChunkMap.firstKey());
+            /*
+             If dataChunkMap is full, get the committedWriteOffset. This will make a API call to the
+             server and add latency in this path/context. Since there are already chunks in flight,
+             calling this API will not reduce overall throughput. It will throttle the upstream
+             write call now rather than onFinalize, which is fine. This will also increase QPS to
+             the GCS backend. The increase will be linear to number of chunks written, so that
+             should also be fine.
+            */
+            committedWriteOffset = getCommittedWriteSizeWithRetries(uploadId);
```
mprashanthsagar, line 272: Can this be done asynchronously ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -425,16 +446,18 @@ public void onError(Throwable t) {
...
         if (transientError == null) {
-          nonTransientError =
-              new IOException(
-                  String.format(
-                      "Caught exception for '%s', while uploading to uploadId %s at writeOffset %d."
-                          + " Status: %s",
-                      resourceId, uploadId, writeOffset, status.getDescription()),
-                  t);
+          nonTransientError = t;
+          logger.atSevere().log(
```
mprashanthsagar, line 457: Logging is a way of handling the exception, we don't want to log and throw the exception (we are implicitly throwing as the observer awaits & checks the exception)

We should handle one exception only at one layer

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -252,24 +255,52 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+                channelOptions.getNumberOfBufferedRequests(),
+                this.writeOffset,
+                committedWriteOffset);
+
+            // check and remove chunks from dataChunkMap
+            while (dataChunkMap.size() > 0 && dataChunkMap.firstKey() < committedWriteOffset) {
+              logger.atFinest().log(
+                  "clearing dataChunkMap one buffer at a time, size: %d, firstKey:%d, committedwriteOffset:%d",
+                  dataChunkMap.size(), dataChunkMap.firstKey(), committedWriteOffset);
+              dataChunkMap.remove(dataChunkMap.firstKey());
```
mprashanthsagar, line 285: Is it safe to remove data from the map while iterating over it ? may be it is safer to use iterator/ stream API

```
map.entrySet().removeIf(entry -> entry.getKey() < committedWriteOffset));
```

OR
```
Iterator<Map.Entry<String, Integer>> it = map.entrySet().iterator();
        while (it.hasNext()) {
            if (it.next().getKey() < committedWriteOffset)){
                it.remove();
            } else {
               break;
           }
        }
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -177,7 +180,7 @@ private class UploadOperation implements Callable<WriteObjectResponse> {
     private InsertChunkResponseObserver responseObserver;
     // Holds list of most recent number of NUMBER_OF_REQUESTS_TO_RETAIN requests, so upload can
     // be rewound and re-sent upon transient errors.
-    private final TreeMap<Long, ByteString> dataChunkMap = new TreeMap<>();
+    private final TreeMap<Long, WriteObjectRequest> dataChunkMap = new TreeMap<>();
```
mprashanthsagar, line 183: Let's avoid making multiple changes in a single PR. Can we split this change

1. Dropping buffered data
2. Caching requests instead of ByteStrings -- I suspect this change can be avoided now if we reconsider the design itself skipping the extra memory copy
3. Retry handling with TRANSIENT_ERRORS

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: `writeOffset` tracks the client offset, it is used as we are writing the chunks. `committedWriteOffset` tracks server side offset and is used to clear the local buffers. Hence we need both of them.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: It could be done, but then we will need a block/signal code to get back control. Moreover, when we are calling this API, we (this thread) are idle, as we are not able to write subsequent buffers. So for simplicity sake, I would prefer this approach.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: I think all of these are related. The PR anyway logs the commit history, so although the branch-2.x.x will have one commit, the history will be preserved. Splitting to 3 PR will require a lot more work.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: Current implementation is safe. We are not really iterating, we are looking at the head(`firstKey`) and removing it if it is committed.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: `getResponseOrThrow` wraps the nonTransientError into IOException. So in `onError` if we do as previous implementation, then we end up wrapping the nonTransientError into IOException twice. Hence removing the wrapping.

Re: logging, we are not throwing any exception in this method, we are merely storing them. Hence added a log to indicate that an exception from the readResponse has occurred at this time.

PR #774, opened 2022-05-02
Add generated preprod client for GCS

mprashanthsagar:


medb (changes requested):
How this client was generated? Usually Maven build should be configured to generate this code during build instead of directly committing it to the repo.

mprashanthsagar:
> How this client was generated? Usually Maven build should be configured to generate this code during build instead of directly committing it to the repo.

This client was generated from https://github.com/google/apis-client-generator (from GCS team) based on the api config for Directory apis, the config isn't published to production yet.

medb:
> This client was generated from https://github.com/google/apis-client-generator (from GCS team) based on the api config for Directory apis, the config isn't published to production yet.

I think that the right approach will be to integrate running this script in generate sources phase of the Maven build: https://maven.apache.org/guides/mini/guide-generating-sources.html

mprashanthsagar:
> > This client was generated from https://github.com/google/apis-client-generator (from GCS team) based on the api config for Directory apis, the config isn't published to production yet.
> 
> I think that the right approach will be to integrate this script in Maven build and run it during `generate-sources` phase of the Maven build: https://maven.apache.org/guides/mini/guide-generating-sources.html


The code generator requires python 2.7 and `pip install requirements` to even build the generator. I don't think it is feasible to  integrate into the maven generate sources phase. It is a one time setup (temporary) till APIs get published, which would unblock us on development

Deependra-Patel (approved):

PR #776, opened 2022-05-05
Lazy prefetch of footer

mayanks:
prefetch footer only if there is a seek in the footer range

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -135,7 +135,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
 
   Fadvise readStrategy;
 
-  private final byte[] footerBuffer;
+  private byte[] footerBuffer = null;
 
   private final long footerStartOffsetInBytes;
 
@@ -175,8 +175,6 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     int prefetchSizeInBytes = readOptions.getMinRangeRequestSize() / 2;
     this.gRPCReadMessageTimeout = readOptions.getGrpcReadMessageTimeoutMillis();
     this.footerStartOffsetInBytes = max(0, (objectSize - prefetchSizeInBytes));
-    int footerSize = Math.toIntExact(min(prefetchSizeInBytes, objectSize));
-    this.footerBuffer = getFooterContent(footerStartOffsetInBytes, footerSize);
   }
 
   private void validate(GoogleCloudStorageItemInfo itemInfo) throws IOException {
@@ -233,8 +231,6 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     int prefetchSizeInBytes = readOptions.getMinRangeRequestSize() / 2;
     this.gRPCReadMessageTimeout = readOptions.getGrpcReadMessageTimeoutMillis();
     this.footerStartOffsetInBytes = max(0, (objectSize - prefetchSizeInBytes));
-    int footerSize = Math.toIntExact(min(prefetchSizeInBytes, objectSize));
-    this.footerBuffer = getFooterContent(footerStartOffsetInBytes, footerSize);
   }
 
   private GoogleCloudStorageItemInfo getObjectMetadata(StorageResourceId resourceId, Storage gcs)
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -730,6 +726,11 @@ public SeekableByteChannel position(long newPosition) throws IOException {
     cancelCurrentRequest();
     invalidateBufferedContent();
 
+    if (footerBuffer == null && newPosition > 0 && newPosition >= footerStartOffsetInBytes) {
```
mprashanthsagar, line 729: Should add the prefetch cost to `read` instead of `position` ? That would enable us to cache partial overlap with footer data as well ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -135,7 +135,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
 
   Fadvise readStrategy;
 
-  private final byte[] footerBuffer;
+  private byte[] footerBuffer = null;
```
medb, line 138: nit: `null` is a default value, usually we do not set it, I think

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
medb, line 729: I think that `position` should not perform any actual IO, because some clients can call it multiple times until they will actually call read.

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mayanks: Just curious. Why would a client call position (BTW this is `seek`) multiple times before reading?

I agree that semantically this is not the right place to add, but it keeps the code much cleaner. Let me know if you have a strong view on this. I can change the implementation.

mayanks: moved the prefetch to read under different condition. please review. 
I checked the parquet file format. It contains the footer length right at the end. So we may not be able to cache partial overlap in case footer is larger than our `rangerequestsize`. The first seek will happen right at the end of the footer.

mayanks: moved the prefetch to read

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -421,6 +417,18 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
+
+    /* resIterator is null on the first read (position = 0) or when a seek is performed (and when
+      there are exceptions). So if footerBuffer is null and we are trying to read into footer
+      region, we may just cache the footer
+    */
+    if ((resIterator == null)
+        && (footerBuffer == null)
+        && (effectivePosition >= footerStartOffsetInBytes)) {
+      this.footerBuffer =
+          getFooterContent(footerStartOffsetInBytes, (int) (objectSize - footerStartOffsetInBytes));
```
mprashanthsagar, line 429: use ` Math.toIntExact` instead

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -421,6 +417,18 @@ public int read(ByteBuffer byteBuffer) throws IOException {
...
 
     long effectivePosition = positionInGrpcStream + bytesToSkipBeforeReading;
+
+    /* resIterator is null on the first read (position = 0) or when a seek is performed (and when
+      there are exceptions). So if footerBuffer is null and we are trying to read into footer
+      region, we may just cache the footer
+    */
+    if ((resIterator == null)
+        && (footerBuffer == null)
+        && (effectivePosition >= footerStartOffsetInBytes)) {
```
mprashanthsagar, line 427: we are still missing out on overlap. How about 
```
effectivePosition + buffer.size() >= footerStartOffsetInBytes
```

This would avoid any redundant data transmission if the file footer offset is before configured footer .

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
mayanks: I thought about this. But I also read up a little on parquet file format. The length of the footer is stored at the end of the footer. So the first read typically happens at the very end. I don't think optimization will be helpful in reading larger footer.

mayanks:
closing this in favor of #788

PR #778, opened 2022-05-06
Make writes via GCS gRPC API more resilient

mayanks:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -248,24 +252,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+                this.writeOffset,
+                committedWriteOffset);
+
+            // check and remove chunks from dataChunkMap
+            while (dataChunkMap.size() > 0 && dataChunkMap.firstKey() < committedWriteOffset) {
+              logger.atFinest().log(
+                  "clearing dataChunkMap one buffer at a time, size: %d, firstKey:%d, committedwriteOffset:%d",
+                  dataChunkMap.size(), dataChunkMap.firstKey(), committedWriteOffset);
+              dataChunkMap.remove(dataChunkMap.firstKey());
+            }
```
mprashanthsagar, line 281: Could we extract this logic into a method ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -215,9 +217,11 @@ boolean isRetriableError(Throwable throwable) {
     }
 
     private WriteObjectResponse doResumableUpload() throws IOException {
+      long committedWriteOffset = 0;
```
mprashanthsagar, line 220: Do we need two trackers of writeOffset ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -248,24 +252,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+          // for errors.
+          if (dataChunkMap.size() < channelOptions.getNumberOfBufferedRequests()) {
+            ByteString data =
+                ByteString.readFrom(
+                    ByteStreams.limit(pipeSource, MAX_BYTES_PER_MESSAGE), MAX_BYTES_PER_MESSAGE);
+            insertRequest = buildInsertRequest(writeOffset, data, false);
+            dataChunkMap.put(writeOffset, insertRequest);
+            writeOffset += data.size();
+          }
+        }
```
mprashanthsagar, line 294: Instead of creating a busy loop, is there a way to use countdown latch ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -248,24 +252,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
           }
-          insertRequest = buildInsertRequest(writeOffset, data, false);
-          writeOffset += data.size();
-        }
-        requestStreamObserver.onNext(insertRequest);
-        objectFinalized = insertRequest.getFinishWrite();
 
+          // Pick up a chunk to write only if dataChunkMap has space. Else continue after looking
+          // for errors.
+          if (dataChunkMap.size() < channelOptions.getNumberOfBufferedRequests()) {
```
mprashanthsagar, line 286: Could we use `else` block instead of two disjoint `if` clauses ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -248,24 +252,50 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+                    ByteStreams.limit(pipeSource, MAX_BYTES_PER_MESSAGE), MAX_BYTES_PER_MESSAGE);
+            insertRequest = buildInsertRequest(writeOffset, data, false);
+            dataChunkMap.put(writeOffset, insertRequest);
+            writeOffset += data.size();
+          }
+        }
+        if (insertRequest != null) {
+          requestStreamObserver.onNext(insertRequest);
+          objectFinalized = insertRequest.getFinishWrite();
+        }
```
mprashanthsagar, line 298: Can this be moved to previous conditional block ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -341,17 +371,17 @@ private void updateObjectHash(ByteString dataChunk) {
     // This happens if a transient failure happens while uploading, and can be resumed by
     // querying the current committed offset.
     private WriteObjectRequest buildRequestFromBufferedDataChunk(
-        TreeMap<Long, ByteString> dataChunkMap, long writeOffset) throws IOException {
+        TreeMap<Long, WriteObjectRequest> dataChunkMap, long writeOffset) throws IOException {
       // Resume will only work if the first request builder in the cache carries an offset
       // not greater than the current writeOffset.
       WriteObjectRequest request = null;
       if (dataChunkMap.size() > 0 && dataChunkMap.firstKey() <= writeOffset) {
```
mprashanthsagar, line 378: Can we use a guard-clause to exit early instead ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -341,17 +371,17 @@ private void updateObjectHash(ByteString dataChunk) {
     // This happens if a transient failure happens while uploading, and can be resumed by
     // querying the current committed offset.
     private WriteObjectRequest buildRequestFromBufferedDataChunk(
```
mprashanthsagar, line 373: The method name does not reflect the new logic, could we refactor ?

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: insertRequest is being set in 2 out of 3 conditional blocks above. Code will repeat if I move it up.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: As discussed, this is not a busy-loop. we do wait on network IO for the committedOffset response.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -339,34 +366,22 @@ private void updateObjectHash(ByteString dataChunk) {
 
     // Handles the case when a writeOffset of data read previously is being processed.
     // This happens if a transient failure happens while uploading, and can be resumed by
-    // querying the current committed offset.
-    private WriteObjectRequest buildRequestFromBufferedDataChunk(
-        TreeMap<Long, ByteString> dataChunkMap, long writeOffset) throws IOException {
-      // Resume will only work if the first request builder in the cache carries an offset
-      // not greater than the current writeOffset.
+    // querying the writeRequest object at the current committed offset.
+    private WriteObjectRequest fetchRequestFromBufferedDataChunk(
```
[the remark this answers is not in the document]
mayanks: fixed

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -339,34 +366,22 @@ private void updateObjectHash(ByteString dataChunk) {
...
+        TreeMap<Long, WriteObjectRequest> requestChunkMap, long writeOffset) {
       WriteObjectRequest request = null;
-      if (dataChunkMap.size() > 0 && dataChunkMap.firstKey() <= writeOffset) {
-        for (Map.Entry<Long, ByteString> entry : dataChunkMap.entrySet()) {
-          if (entry.getKey() + entry.getValue().size() > writeOffset
+      if (requestChunkMap.size() > 0 && requestChunkMap.firstKey() <= writeOffset) {
+        for (Map.Entry<Long, WriteObjectRequest> entry : requestChunkMap.entrySet()) {
+          if (entry.getKey() + entry.getValue().getChecksummedData().getContent().size()
+                  > writeOffset
               || entry.getKey() == writeOffset) {
```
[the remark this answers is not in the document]
mayanks: We will revist this in a separate PR

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -248,24 +250,28 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
         } else {
-          ByteString data =
-              ByteString.readFrom(
-                  ByteStreams.limit(pipeSource, MAX_BYTES_PER_MESSAGE), MAX_BYTES_PER_MESSAGE);
-          dataChunkMap.put(writeOffset, data);
-          if (dataChunkMap.size() >= channelOptions.getNumberOfBufferedRequests()) {
-            dataChunkMap.remove(dataChunkMap.firstKey());
+          if (requestChunkMap.size() >= channelOptions.getNumberOfBufferedRequests()) {
+            freeUpCommittedRequests(requestChunkMap, writeOffset);
+          } else {
```
mprashanthsagar, line 260: Considering that this path is most accessed, can we ensure that it is the first conditional ? avoiding two failed `if` checks for most calls ?

mprashanthsagar (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: I tried. The logic is not very readable. The logic for first 2 conditions are quite infrequent but to the point. this else is more of a fallback. I'll leave this as is in favor of code readability.

mprashanthsagar: We have atleast 95% of requests following this codepath, we could extract the conditional into a one-line method for readability. The cost might outway in this scenario, we could add additional comments and methods wherever possible

Refer to https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/606 for how much conditionals can add to cpu cost.

mayanks: #606 is not relatable here. In #606 moreServerContent is an IO cost that we could have done without spending.

Here, the `if` condition will be called in each loop. It does not have any IO cost, it is just checking the occupancy of the buffer. What we save over here by re-ordering the code is `jmp` assembly instruction. I don't think we are doing a `assembly code` level optimization here.

To further illustrate if the code is like
```
while (cond) {
  if (cond2) {
    ... do x 5 % of the time ...
  } else {
   ... do y 95% of the time ...
  }
}
```

and if we try re-writing it as 
```

while (cond) {
  if (!cond2) {
   ... do y 95% of the time ...
  } else {
    ... do x 5 % of the time ...
  }
}
```

We do not save any CPU overhead. `cond2` will always be evaluated. What we will save here is below.

CPU will pipeline the assembly instructions to be executed even before evaluating `cond2`. If the condition evaluates to false, it flushes the pipeline and reloads it with the else block. 

This kind of optimization helps tremendously while writing kernel code or tight loops with huge iteration (like memcpy). It does not help us here.

PR #770, opened 2022-04-25
Extract GCSFS to interface

mprashanthsagar:

PR #780, opened 2022-05-10
Extract GCSFS to interface (#770)

mprashanthsagar:
(cherry picked from commit 111d06c48fdab0280e0aa4a379cfde9e069943c3)


medb (changes requested):
This is a breaking change - we can not merge it in subminor version release.

mprashanthsagar:
> This is a breaking change - we can not merge it in subminor version release.

Good catch, I missed that the class earlier would be an interface now. 

WDYT about a minor version upgrade, that is still compatible with Dataproc 2.0 but will allow for gcsio customers to upgrade separately ?

medb:
> > This is a breaking change - we can not merge it in subminor version release.
> 
> Good catch, I missed that the class earlier would be an interface now.
> 
> WDYT about a minor version upgrade, that is still compatible with Dataproc 2.0 but will allow for gcsio customers to upgrade separately ?

I think it depends when GCS feature will go GA - if it will not be GA before we will deprecate 2.0 images, then I do not think there is sense to invest in pre-2.1 images and better to focus on 2.1 images and serverless.

mprashanthsagar:
Discussed offline, will continue development on master branch and fix compatibility with Dataproc 2.0 for customers

PR #781, opened 2022-05-12
Handle partial write update from GCS

mayanks:
Connector supports resuming upload only from MAX_BYTES_PER_MESSAGE
boundary. If GCS reports a committed offset which is not a multiple
of this number, then we cannot resume.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
@@ -220,6 +220,13 @@ public final class GoogleCloudStorageGrpcWriteChannel
       // Only request committed size for the first insert request.
       if (writeOffset > 0) {
         writeOffset = getCommittedWriteSizeWithRetries(uploadId);
+        if (writeOffset % MAX_BYTES_PER_MESSAGE != 0) {
+          throw new IOException(
+              String.format(
+                  "Upstream requested data from a non chunk boundary for resource '%s'"
+                      + "offset:%d, chunkSize:%d",
+                  resourceId, writeOffset, MAX_BYTES_PER_MESSAGE));
+        }
       }
       StorageStub storageStub =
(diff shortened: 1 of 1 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -220,6 +220,13 @@ private WriteObjectResponse doResumableUpload() throws IOException {
       // Only request committed size for the first insert request.
       if (writeOffset > 0) {
         writeOffset = getCommittedWriteSizeWithRetries(uploadId);
+        if (writeOffset % MAX_BYTES_PER_MESSAGE != 0) {
```
mprashanthsagar, line 223: Please add tests to cover the code path.

Could we get a confirmation from GCS team on the behaviour as well ?

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: @veblush In one run of the benchmark we noticed this behavior. Towards the end of the file upload we received a transient error. On trying to resume, we fetched the committedWriteOffset in GCS. It returned a number which was not a multiple of 2MB(`MAX_BYTES_PER_MESSAGE`). Since we store uncommitted buffers in 2MB, we were not able to resume from that offset.

Can you confirm this?

mayanks: > Also, consider this scenario

In this case the complete file is written to the GCS. So when we retry, 
1. We get the file length as the committed Offset. 
2. We don't have a chunk beginning at that in our buffer, so 
3. We end up trying to read from the pipe, which returns EOF (0). 
4. This results in uploading `finalized` chunk to GCS with size as 0.

PR #783, opened 2022-05-13
Fix: prevent clobbering of SSL trustCertificates

Deependra-Patel:
Turns out NetHttpTransport.Builder#trustCertificates() clobbers any changes made to sslSocketFactory beforehand and this change is to handle modifying Socket keepAlive option while still supporting trustCertificates as expected.


medb:
I have fixed up original https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/782 PTAL

PR #785, opened 2022-05-16
Write stall on ready

mayanks:
Very rarely we observe that the write upload thread gets blocked on the `ready` latch and never returns. This causes the write job to stall indefinitely.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -239,7 +240,8 @@ private WriteObjectResponse doResumableUpload() throws IOException {
 
       // Wait for streaming RPC to become ready for upload.
       try {
-        responseObserver.ready.await();
+        // wait for 1 min for the channel to be ready. Else bail out
+        responseObserver.ready.await(60*1000, MILLISECONDS);
```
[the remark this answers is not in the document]
mayanks: ok. so this guy raises an exception if we are not `ready` in 60s. The `try-catch` block already catches the exception and bails out. we won't reach `observer.done.await`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -239,7 +240,13 @@ private WriteObjectResponse doResumableUpload() throws IOException {
 
       // Wait for streaming RPC to become ready for upload.
       try {
-        responseObserver.ready.await();
+        // wait for 1 min for the channel to be ready. Else bail out
+        if (!responseObserver.ready.await(60*1000, MILLISECONDS)) {
```
mprashanthsagar, line 244: let's add a test to cover the path, lgtm otherwise

mayanks: I tried looking into the test code and am not able to figure out how to trigger this failure. can you help @mprashanthsagar

abmodi: IMO, the issue seems that we are not doing countDown for ready on error. What happens if error happens before start itself?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
```diff
@@ -239,7 +240,13 @@ private WriteObjectResponse doResumableUpload() throws IOException {
...
+          throw new IOException(
```
arunkumarchacko, line 245: log?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mprashanthsagar: @abmodi Do you mean error from the api/connector ? 

If it is from the API, that should trigger the onError callback. Failing that the timeout should kick-in. 

If it is connector, that exception should bubble up if we have not reached the latch, If blocked on latch, timeout should kick-in again

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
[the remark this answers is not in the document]
mayanks: We don't have telemetry data on the delay we observe. The issue itself is fairly rare to reproduce. This timeout can be related to http-open timeout because here we are trying to establish grpc stream I guess. We could use the same timeout or as you suggest create a new configurable parameter. 

The fact that it is an extremely rare scenario, makes me lean towards not making it a configurable parameter. When this case is hit, we do retry it internally, so it is not a user facing consequence.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannel.java
mayanks: The responseObserver error is called after the channel is established. In this case we are observing that the channel establishment itself just hangs. There is no path in the code where onError is called while establishing the channel (IIRC)

abmodi: Sounds good.

abmodi (approved):

PR #787, opened 2022-05-17
Implement mkdir with directory support

Deependra-Patel:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
```diff
@@ -91,15 +113,17 @@ public GoogleCloudStorageNativeFileSystem(
    */
   @VisibleForTesting
   public GoogleCloudStorageNativeFileSystem(
```
mprashanthsagar, line 115: SInce the method is only used for testing, and we are modifying the signature. Can we make it package-private ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
```diff
@@ -40,6 +57,9 @@ public class GoogleCloudStorageNativeFileSystem implements GoogleCloudStorageFil
   // GCS access instance.
   private final GoogleCloudStorage gcs;
 
+  // Storage instance with directory support
+  private final Storage nativeStorage;
```
mprashanthsagar, line 61: nit : Should we stick to `storage` ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
```diff
@@ -188,7 +246,36 @@ public void close() {
...
+    Insert insertDir =
+        nativeStorage
+            .directories()
+            .insert(
+                resourceId.getBucketName(), new Directory().setName(resourceId.getObjectName()));
+    try {
+      insertDir.execute();
+    } catch (IOException e) {
+      if (ApiErrorExtractor.INSTANCE.itemAlreadyExists(e)) {
+        throw (FileAlreadyExistsException)
```
mprashanthsagar, line 273: Why do we need type casting here ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
```diff
@@ -188,7 +246,36 @@ public void close() {
 
   @Override
   public void mkdir(URI path) throws IOException {
```
mprashanthsagar, line 248: Can we add some unit tests to cover the functionality ?

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
Deependra-Patel: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNativeFileSystem.java
Deependra-Patel: `.initCause` returns a Throwable

PR #792, opened 2022-05-30
New release 2.1.8

mayanks:


gcs/CHANGES.md
```diff
@@ -1,5 +1,9 @@
 ### 2.1.7 - 2022-02-03
```
arunkumarchacko, line 1: Should this be 2.1.8? Also, should we update the date?

mayanks: Thanks for pointing this out Arun

PR #796, opened 2022-06-02
Add support for tracing GCS API calls

arunkumarchacko:
This is disabled by default behind <fs.gs.tracelog.enable> flag.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private static final Gson gson = new Gson();
+
+  // Using a ConcurrentHashMap with weak key reference to avoid potential memory leak during failure
+  // scenarios. We do not get an response interceptor callback if there is some failure while
+  // executing the
+  // HttpRequest.
+  private final Map<HttpRequest, HttpRequestResponseTimeTracker> requestTracker =
```
Deependra-Patel, line 28: static?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.common.flogger.GoogleLogger;
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
```
Deependra-Patel, line 20: How is this being enabled based on the flag?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
...
+public class EventLoggingHttpRequestInitializerTest {
+  private static final String requestUrl = "http://google.com";
+
+  @Mock private LowLevelHttpRequest mockLowLevelRequest;
+  @Mock private LowLevelHttpResponse mockLowLevelResponse;
+
+  private HttpRequest httpRequest;
+  private EventLoggingHttpRequestInitializer requestInitializer;
+  private EventLoggingHttpRequestInitializer requestInitializerSpy;
+  private ArgumentCaptor<HashMap<String, Object>> logDetailsCaptor;
```
Deependra-Patel, line 35: Maybe it is better to test the serialized json as argument, so we also test that logged format is indeed json.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Static also would have worked as well. I think having non-static is more natural here since there will be on tracker per GCSCloudStorage.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: That will be a separate change. Done that way since there will be conflicts b/w master and 2.2.x branch.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: 1. If we have to test this, then we will have to mock GoogleLogger static methods, which is not straightforward to do. Note that creating a method with string is not an option due to performance reasons (i.e. we want to pay the cost of creating the log only if the log is enabled).
2. The functionality of converting a map to Json is from GSON library and there is not much value testing that in this classes UT.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -198,6 +199,8 @@ public static Builder builder() {
 
   public abstract MetricsSink getMetricsSink();
 
+  public abstract Boolean getTraceLogEnable();
```
medb, line 202: Getters for boolean fields should have `is` prefix (see other configs)

medb, line 202: Use `Enabled` in the getter and setter method names (see other boolean configs)

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+public class EventLoggingHttpRequestInitializerTest {
+  private static final String requestUrl = "http://google.com";
+
+  @Mock private LowLevelHttpRequest mockLowLevelRequest;
+  @Mock private LowLevelHttpResponse mockLowLevelResponse;
```
medb, line 30: We try to not use mocks in the GCS connector code base because they are hard to maintain, use fakes, stubs, etc instead (see other test for examlpe).

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
medb, line 20: Add Javadoc

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.IOException;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private static final Gson gson = new Gson();
```
medb, line 22: Can it be non-static field (especially if we expect to have one instance of this class)?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+      requestTracker.remove(request);
+      jsonMap.put("response_time", tracker.getResponseTime());
+      jsonMap.put("request_start_time_utc", tracker.getStartTimeUtc());
+      jsonMap.put("request_finish_time_utc", tracker.getCurrentTimeUtc());
+    } else {
+      jsonMap.put("response_time", Integer.MAX_VALUE);
+      jsonMap.put("unexpected_error", "Zombie request. This is unexpected.");
+    }
+
+    logDetails(jsonMap);
```
mprashanthsagar, line 62: Let's separate logging and capturing the event logs, So that we have one class, that can be controlled via loggging configurations to capture all data.
That makes the fluentd config static

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @VisibleForTesting
+  void logDetails(Map<String, Object> jsonMap) {
+    logger.atInfo().log(gson.toJson(jsonMap));
```
mprashanthsagar, line 67: nit : Might be useful to accept log level as an argument as well, to control the different levels of logs for the future (in a separate class)

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+      return stopwatch.elapsed(MILLISECONDS);
+    }
+
+    String getStartTimeUtc() {
+      return dateTimeFormatter.format(startTime.toInstant());
+    }
+
+    String getCurrentTimeUtc() {
+      return dateTimeFormatter.format(new Date().toInstant());
+    }
```
mprashanthsagar, line 87: Do we want to hard code UTC ? why no rely on local time zone, this might conflict for the customers on-prem

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
mprashanthsagar, line 20: I don't see this request initializer getting used anywhere ? Is that intentional ? 

I think we are missing the wiring part of `isTraceEnabled`

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private static final Gson gson = new Gson();
+
+  // Using a ConcurrentHashMap with weak key reference to avoid potential memory leak during failure
+  // scenarios. We do not get an response interceptor callback if there is some failure while
+  // executing the
```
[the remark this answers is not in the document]
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: We will make the change if we add a second instance of this and if we find a need for this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: I think adding log level also as config might result in us having a lot of if-else. This can be taken up in future if we find a need for this. Not taking this now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: The main audience for this is GCS team. For them if might be useful to have a standard time, rather than the local time on the machine. Also it is quite easy to convert to any other time zone if required.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
arunkumarchacko: Done

arunkumarchacko: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Done. Thank you for catching this.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Done. Thank you for the suggestion.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Yes, it will be separate change. Split this into two that change will have conflict b/w master and 2.2.x branch and it would be cleaner to separate them into to PRs.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private final Gson gson = new Gson();
+
+  // Using a ConcurrentHashMap with weak key reference to avoid potential memory leak during failure
+  // scenarios. We do not get an response interceptor callback if there is some failure while
+  // executing the HttpRequest.
+  private final Map<HttpRequest, HttpRequestResponseTimeTracker> requestTracker =
+      new MapMaker().weakKeys().makeMap();
```
medb, line 46: Is this a thread-safe map?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+public class EventLoggingHttpRequestInitializerTest {
+  private static final String requestUrl = "http://google.com";
+
+  private EventLoggingHttpRequestInitializer requestInitializer;
+  private EventLoggingHttpRequestInitializer requestInitializerSpy;
+  private ArgumentCaptor<HashMap<String, Object>> logDetailsCaptor;
```
medb, line 40: I think that Flogger/Java Util Logging should have some ways to test log records that do not require intercepting - if they exist then it could be better to use it to test this e2e.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Test
+  public void testIOExceptionNoCallback() throws IOException {
+    HttpRequest httpRequest = getHttpRequestWithResponse(new IOException("test IOException"));
+    requestInitializerSpy.initialize(httpRequest);
+
+    try {
+      httpRequest.execute();
+    } catch (IOException ioe) {
+    }
```
medb, line 84: If exception is expected here, then we should use `assertThrows(...)`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,111 @@
...
+    requestInitializerSpy.initialize(httpRequest);
+    httpRequest.execute();
+
+    verify(requestInitializerSpy).logDetails(logDetailsCaptor.capture());
+    verify(requestInitializerSpy, times(1)).logDetails(any());
+    Map<String, Object> captured = logDetailsCaptor.getValue();
+
+    assertThat(captured.get("response_time")).isNotNull();
+    assertThat(captured.get("request_url")).isEqualTo(requestUrl);
+    assertThat(captured.get("response_status_code")).isEqualTo(200);
```
medb, line 73: Better to assert content of whole map, if possible, instead of individual keys

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: yes, it uses ConcurrentHashMap. Reference https://guava.dev/releases/21.0/api/docs/com/google/common/collect/MapMaker.html

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Many of the fields are not deterministic, like the response time, start/end time etc.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
mprashanthsagar: We already have a plan to add logs for gRPC path, might as well bite the bullet now.

This ensure that there is no changes downstream (fluentd) for the benchmarks

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
mprashanthsagar: This doesn't honor the jvm timezone property, and there will be discrepancies. I think we should rely on cloud logging to address the times (event time vs log time). I recommend that we stick to default timezone, instead of hard coding it. This also becomes difficult to evolve with time over multiple releases

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class EventLoggingHttpRequestInitializerTest {
+  private static final String REQUEST_URL = "http://google.com";
+  private static final Logger JDK_LOGGER =
```
mprashanthsagar, line 42: nit: Can we use a more indicative var name ?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -179,6 +179,19 @@ public void testOpenWithMetricsEnabled() throws IOException {
     assertObjectContent(rawStorage, objectToCreate, objectBytes);
   }
 
+  @Test
+  public void testOpenWithTracingLogEnabled() throws IOException {
```
mprashanthsagar, line 183: Let's add an integration test for the JSON path as well. Ensuring no regressions

mprashanthsagar (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: The timestamps (receiveTimestamp and logTimestamp) from Cloud Logging are there, which is also in UTC. Note that the logTimestamp is the time when the logger actually received the log, which is when the request actually ends. Someone looking at the logs will have to subtract the response_time from the logTimestamp to get when the request was fired. Hence it will be useful (even though it can be derived from latency and log time) to have the start and end times in json_payload itself. Also since the Cloud logging log times are in UTC it is better to have the same in the json_payload

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
mprashanthsagar: bump

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+  private void logDetails(Map<String, Object> jsonMap) {
+    logger.atInfo().log("%s", LazyArgs.lazy(() -> gson.toJson(jsonMap)));
+  }
+
+  private static class HttpRequestResponseTimeTracker {
+    private static final DateTimeFormatter dateTimeFormatter =
+        DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXXX").withZone(ZoneOffset.UTC);
+
+    private final Stopwatch stopwatch = Stopwatch.createStarted();
+    private final Date startTime = new Date();
```
medb, line 92: If we need to use UTC time, then probably Instant instead of Date will be better to use here

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+    long getResponseTime() {
+      return stopwatch.elapsed(MILLISECONDS);
+    }
+
+    String getStartTimeUtc() {
+      return dateTimeFormatter.format(startTime.toInstant());
+    }
+
+    String getCurrentTimeUtc() {
+      return dateTimeFormatter.format(new Date().toInstant());
```
medb, line 103: Will be better to use `Instant.now()` here.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+    jsonMap.put("request_url", request.getUrl().toString());
+    jsonMap.put("request_headers", request.getHeaders());
+    jsonMap.put("response_headers", httpResponse.getHeaders());
+    jsonMap.put("response_status_code", httpResponse.getStatusCode());
+    jsonMap.put("response_status_message", httpResponse.getStatusMessage());
+    jsonMap.put("thread_name", Thread.currentThread().getName());
+
+    HttpRequestResponseTimeTracker tracker = requestTracker.get(request);
+    if (tracker != null) {
+      requestTracker.remove(request);
```
medb, line 71: nit: you can use `remove` instead of `get` above - `return` returns removed element if it was present in the map.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+      jsonMap.put("response_time", tracker.getResponseTime());
+      jsonMap.put("request_start_time_utc", tracker.getStartTimeUtc());
+      jsonMap.put("request_finish_time_utc", tracker.getCurrentTimeUtc());
+    } else {
+      jsonMap.put("response_time", Integer.MAX_VALUE);
+      jsonMap.put("unexpected_error", "Zombie request. This is unexpected.");
```
medb, line 77: nit: use `Unknown request` - it maybe more clear than `Zombie request`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
arunkumarchacko: I will add this with the next change when we actually consume the tracer.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+  // scenarios. We do not get an response interceptor callback if there is some failure while
+  // executing the HttpRequest.
+  private final Map<HttpRequest, HttpRequestResponseTimeTracker> requestTracker =
+      new MapMaker().weakKeys().makeMap();
+
+  @Override
+  public void initialize(HttpRequest request) throws IOException {
+    request.setInterceptor(
+        httpRequest -> requestTracker.put(httpRequest, new HttpRequestResponseTimeTracker()));
+    request.setResponseInterceptor(this::logAndRemoveRequestFromTracking);
```
medb, line 53: Should we also track requests for which this method never was called because of some reason?

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
```diff
@@ -179,6 +179,19 @@ public void testOpenWithMetricsEnabled() throws IOException {
...
+    GoogleCloudStorage rawStorage =
+        new GoogleCloudStorageImpl(
+            configureDefaultOptions()
+                .setTraceLogEnabled(true)
+                .build(),
+            GoogleCloudStorageTestHelper.getCredentials());
+    StorageResourceId objectToCreate = new StorageResourceId(BUCKET_NAME, "testOpen_Object");
+    byte[] objectBytes = writeObject(rawStorage, objectToCreate, /* objectSize= */ 100);
+    assertObjectContent(rawStorage, objectToCreate, objectBytes);
```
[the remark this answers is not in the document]
arunkumarchacko: It is already verified in unit tests. I think that should be sufficient.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: If we have to implement this, then we need to have some sort of mechanism to remove the items from the tracker. Note that we do not get a call back if there is some form of error while executing the request.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializerTest.java
arunkumarchacko: Changed this to LOGGER

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+ * tracks the response time taken by the successful HttpRequests.
+ */
+class EventLoggingHttpRequestInitializer implements HttpRequestInitializer {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private final Gson gson = new Gson();
+
+  // Using a ConcurrentHashMap with weak key reference to avoid potential memory leak during failure
+  // scenarios. We do not get an response interceptor callback if there is some failure while
+  // executing the HttpRequest.
+  private final Map<HttpRequest, HttpRequestResponseTimeTracker> requestTracker =
```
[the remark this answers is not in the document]
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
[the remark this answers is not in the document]
medb: Agree that it's already checked in unit test, but if we do assert this in integration test, I'm not sure that we need this test at all, as it just sets option, but do not actually check if it has any effect, unless we want to test that it does not break anything else?

medb (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageGrpcIntegrationTest.java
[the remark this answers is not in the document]
arunkumarchacko: yes, this is a Sanity check. I do agree that most of the ground is covered by UT.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/EventLoggingHttpRequestInitializer.java
arunkumarchacko: YAGNI. We will cross the bridge when we reach there.

PR #801, opened 2022-06-06
Merge GoogleHadoopSyncableOutputStream functionality in the GoogleHadoopOutputStream

medb:


medb:
> `spark.history.fs.gs.outputstream.type=FLUSHABLE_COMPOSITE` is used for some benchmarks and as a work around for PHS today.
> 
> Is there an alternative for that, if we plan to deprecate the option ? cc : @abmodi @ranu010101

After this change this is an obsolete option because this functionality is available in a single output stream, so it will work as expected out of the box. We may still want to have PHS-specific sync interval, but I think that this is optional and we can add it later if necessary.

gcs/CHANGES.md
```diff
@@ -64,6 +64,13 @@
     fs.gs.auth.refresh.token
     ```
 
+1.  Merge all output stream types functionality in the default output stream,
+    and remove obsolete property:
```
cyxxy, line 68: Please add which of the original 3 types is the new default now.

cyxxy (approved):

gcs/CHANGES.md
medb: Done.

mprashanthsagar (approved):

PR #804, opened 2022-06-14
Updating minRangeRequestSize to take long instead of int

abmodi:
Updating configuration "fs.gs.inputstream.min.range.request.size" to take long instead of int.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -406,6 +406,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<FilesystemAPI> GCS_FILESYSTEM_API =
       new HadoopConfigurationProperty<>("fs.gs.filesystem.api", FilesystemAPI.OBJECT);
 
+  /** Configuration key to enable logging of additional trace details. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_TRACE_LOG_ENABLE =
+      new HadoopConfigurationProperty<>("fs.gs.tracelog.enable", false);
+
   // TODO(b/120887495): This @VisibleForTesting annotation was being ignored by prod code.
   // Please check that removing it is correct, and remove this comment along with it.
   // @VisibleForTesting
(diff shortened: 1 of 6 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -337,10 +337,10 @@ public class GoogleHadoopFileSystemConfiguration {
    * Minimum size in bytes of the HTTP Range header set in GCS request when opening new stream to
    * read an object.
    */
-  public static final HadoopConfigurationProperty<Integer> GCS_INPUT_STREAM_MIN_RANGE_REQUEST_SIZE =
+  public static final HadoopConfigurationProperty<Long> GCS_INPUT_STREAM_MIN_RANGE_REQUEST_SIZE =
       new HadoopConfigurationProperty<>(
           "fs.gs.inputstream.min.range.request.size",
-          GoogleCloudStorageReadOptions.DEFAULT_MIN_RANGE_REQUEST_SIZE);
+          (long) GoogleCloudStorageReadOptions.DEFAULT_MIN_RANGE_REQUEST_SIZE);
```
medb, line 343: Probably better change type of the `DEFAULT_MIN_RANGE_REQUEST_SIZE` class field to long too

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
abmodi: Yes - I tried that but some tests have hard dependency on this variable to be int. They are using this variable in creating ByteBuffer and that takes int as capacity.

mprashanthsagar: We could use `toIntExact` similar to https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L808

+1 to updating the default to long instead

abmodi: Fixed it. Thanks!

medb (approved):

PR #809, opened 2022-06-21
Grpc read optimization to not prematurely close existing requests

abmodi:
Hadoop FsDataInputStream exposes read APIs which will call seek back to original position after the read. Generally they are followed by another seek before actual read happens. We should only close the requests when we are actually reading rather than on seeks.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -308,7 +308,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
   }
 
   private byte[] getFooterContent() throws IOException {
-    int footerSize =  Math.toIntExact(objectSize - footerStartOffsetInBytes);
+    int footerSize = Math.toIntExact(objectSize - footerStartOffsetInBytes);
     ByteBuffer buffer = ByteBuffer.allocate(footerSize);
 
     // snapshot these variables because readFromGCS clobbers them
@@ -358,8 +358,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
   private int readBufferedContentInto(ByteBuffer byteBuffer) {
     // Handle skipping forward through the buffer for a seek.
     long bytesToSkip = positionForNextRead - positionInGrpcStream;
-    long bufferSkip =
-        min(bufferedContent.size() - bufferedContentReadOffset, bytesToSkip);
+    long bufferSkip = min(bufferedContent.size() - bufferedContentReadOffset, bytesToSkip);
     bufferSkip = max(0, bufferSkip);
     bufferedContentReadOffset += bufferSkip;
     positionInGrpcStream += bufferSkip;
@@ -435,7 +434,6 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
       return bytesRead > 0 ? bytesRead : -1;
     }
 
-
     /* resIterator is null on the first read (position = 0) or when a seek is performed (and when
       there are exceptions). So if footerBuffer is null and we are trying to read into footer
       region, we may just cache the footer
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -389,6 +382,22 @@ private int readBufferedContentInto(ByteBuffer byteBuffer) {
     return bytesToWrite;
   }
 
+  private boolean canReadFromExistingRequest(ByteBuffer byteBuffer) {
```
arunkumarchacko, line 385: can we have the private methods after the public methods? Will help with readability.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -389,6 +382,22 @@ private int readBufferedContentInto(ByteBuffer byteBuffer) {
...
+    if (resIterator == null) {
+      return false;
+    }
+
+    if (positionForNextRead < positionInGrpcStream) {
```
arunkumarchacko, line 390: when can this happen?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -389,6 +382,22 @@ private int readBufferedContentInto(ByteBuffer byteBuffer) {
     return bytesToWrite;
   }
 
+  private boolean canReadFromExistingRequest(ByteBuffer byteBuffer) {
+    if (resIterator == null) {
```
arunkumarchacko, line 386: does this method need to be synchronised? i.e. can resIterator be set to null from another thread? Same for other instance fields used.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -725,30 +728,17 @@ public SeekableByteChannel position(long newPosition) throws IOException {
...
     if (readStrategy == Fadvise.AUTO) {
-      if (seekDistance < 0 || seekDistance > readOptions.getInplaceSeekLimit()) {
+      if (positionForNextRead < positionInGrpcStream
+          || positionForNextRead - positionInGrpcStream > readOptions.getInplaceSeekLimit()) {
         readStrategy = Fadvise.RANDOM;
       }
     }
-
-    // Reset any ongoing read operations or local data caches.
-    cancelCurrentRequest();
```
arunkumarchacko, line 747: Is it ok to remove this? i.e. can b/210660938 regress?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
abmodi: In case of backward seek.

mprashanthsagar (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -363,11 +357,10 @@ private static void put(ByteString source, int offset, int size, ByteBuffer dest
 
   private int readBufferedContentInto(ByteBuffer byteBuffer) {
     // Handle skipping forward through the buffer for a seek.
-    long bufferSkip =
-        min(bufferedContent.size() - bufferedContentReadOffset, bytesToSkipBeforeReading);
+    long bytesToSkip = positionForNextRead - positionInGrpcStream;
+    long bufferSkip = min(bufferedContent.size() - bufferedContentReadOffset, bytesToSkip);
```
mayanks, line 361: while we are at it (simplifying the logic) we can look at getting rid of `bufferedContentReadOffset` variable as well. We can ensure that `bufferedContent` always has relevant data from offset 0. Else we can  do 
```
   bufferedContent = bufferedContent.substring(skipBytes)
```
This reduces the complexity of maintaining both `bufferedContent` and `bytesToSkipBeforeReading`

mayanks (approved):

PR #819, opened 2022-07-06
Added Time-Specific Metrics

guljain:
Added time-related metrics. 
Functions chosen are -
1. Open : Present in GoogleHadoopFileSystem class with unit test in GoogleHadoopFileSystemIntegrationTest class.
2. Rename : Presentin GoogleHadoopFileSystem class with unit test in GoogleHadoopFileSystemIntegrationTest class.
3. Delete : Present in GoogleHadoopFileSystem class with unit test in GoogleHadoopFileSystemIntegrationTest class.
4. Create : Present in GoogleHadoopFileSystem class with unit test in GoogleHadoopFileSystemIntegrationTest class.
5. Hsync : Present in GoogleHadoopOutputStream class with unit test in GoogleHadoopOutputStreamTest class.
6. Hflush : Present in GoogleHadoopOutputStream class with unit test in GoogleHadoopOutputStreamTest class.
7. Close : Present in GoogleHadoopOutputStream class with unit test in GoogleHadoopOutputStreamTest class.
8. Seek : Present in GoogleHadoopFSInputStream class with unit test in GoogleHadoopFSInputStreamIntegrationTest class.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Maps;
+import java.util.EnumSet;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/** Time Statistics which are collected in GCS. */
+@InterfaceStability.Unstable
+public enum GhfsTimeStatistic {
```
arunkumarchacko, line 12: Does this need to be public?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
+package com.google.cloud.hadoop.fs.gcs;
+
```
arunkumarchacko, line 2: please add license header

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Maps;
+import java.util.EnumSet;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/** Time Statistics which are collected in GCS. */
+@InterfaceStability.Unstable
+public enum GhfsTimeStatistic {
+  /** Total Time for open(). */
+  OPEN("timeElapsedOpenMusec", "total time required for open in µ seconds", 0),
```
arunkumarchacko, line 14: consider renaming this to micro (also at other places

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+  /** Total Time for hflush(). */
+  HFLUSH("timeElapsedHflushMusec", "total time required for hflush in µ seconds", 0),
+
+  /** Total Time for hsync(). */
+  HSYNC("timeElapsedHsyncMusec", "total time required for hsync in µ seconds", 0),
+
+  /** Total Time for seek(). */
+  SEEK("timeElapsedSeekMusec", "total time required for seek in µ seconds", 0);
+
+  public static final ImmutableSet<GhfsTimeStatistic> VALUES =
```
arunkumarchacko, line 37: why not https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableEnumSet.java?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+      ImmutableSet.copyOf(EnumSet.allOf(GhfsTimeStatistic.class));
+
+  /** A map used to support the {@link #fromSymbol(String)} call. */
+  public static final ImmutableMap<String, GhfsTimeStatistic> SYMBOL_MAP =
```
arunkumarchacko, line 41: can this be private?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+      Maps.uniqueIndex(Iterators.forArray(values()), GhfsTimeStatistic::getSymbol);
+
+  /**
+   * Statistic definition.
+   *
+   * @param symbol name
```
arunkumarchacko, line 47: comment is not adding much value.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+   * @param description description.
+   * @param value value
+   */
+  GhfsTimeStatistic(String symbol, String description, long value) {
+    this.symbol = symbol;
+    this.description = description;
+    this.value = value;
+  }
+  /** Statistic name. */
+  private final String symbol;
```
arunkumarchacko, line 57: please keep the fields above the constructor

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Statistic definition.
+   *
+   * @param symbol name
+   * @param description description.
+   * @param value value
+   */
+  GhfsTimeStatistic(String symbol, String description, long value) {
+    this.symbol = symbol;
```
arunkumarchacko, line 52: validate that the parameters are not null?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
...
+      Maps.uniqueIndex(Iterators.forArray(values()), GhfsTimeStatistic::getSymbol);
+
+  /**
+   * Statistic definition.
+   *
+   * @param symbol name
+   * @param description description.
+   * @param value value
+   */
+  GhfsTimeStatistic(String symbol, String description, long value) {
```
arunkumarchacko, line 51: this this be private?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,104 @@
+package com.google.cloud.hadoop.fs.gcs;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Maps;
+import java.util.EnumSet;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/** Time Statistics which are collected in GCS. */
```
arunkumarchacko, line 10: I think there is scope to improve this comment

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -126,7 +127,9 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
 
   @Override
   public synchronized void seek(long pos) throws IOException {
+    long startTime = System.nanoTime();
```
arunkumarchacko, line 130: why not use stopwatch?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
arunkumarchacko, line 12: I am not sure creating a mutable enum is the right thing to do. Why not create class instead? Also having this singleton make the testing harder

guljain: For the plugin GhfsTimeStatisticPlugin, I had references to loop over various metrics. This is the reason why mutable enum was used. Also by class, do you mean an implementation like https://github.com/LucaCanali/hadoop/blob/s3aAndHDFSTimeInstrumentation/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3ATimeInstrumentation.java ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
guljain: Do you mean 'timeElapsedOpenMicrosec" ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
guljain: I did not think of using stopwatch but nanoTime seems to be fast and precise.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
guljain: Yes, done/

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
guljain: Okay, removed.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
guljain: I was referring to GhfsStatistic class and implemented the same way. Changing this to ImmutableEnumSet will have no impact in the connector performance (as per my understanding) as it's role to list out enum object's symbols in the plugin. In connector, it is not called anywhere.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: I think Stopwatch might be the right choice. This is what we use at other places in the connector too.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
arunkumarchacko: I think mutable enums are an anti-pattern since it is sort of like global variables, which are hard to test etc. I was curious why can't we achieve the same functionality we need by using Java class instead of using Java Enum?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Maps;
+import java.util.EnumSet;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/** Time Statistics capturing total time taken by a function in micro seconds */
+@InterfaceStability.Unstable
+public enum GhfsTimeStatistic {
```
abmodi, line 29: any plans to capture read/write time?

guljain: I believe that is captures in hflush, hsync and close statistics from GoogleHadoopOutputStream. If not, can you point to specific functions ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -496,14 +496,21 @@ public String getScheme() {
 
   @Override
   public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOException {
+    long startTime = System.nanoTime();
```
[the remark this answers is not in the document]
vinayakumarb: Does it mean, need to wrap this entire method call inside lambda? or export existing contents of method to another method and call that from lambda?

Clearly cannot wrap the public method call itself!

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
guljain: Having enum class makes the metrics look more organized in my opinion. I agree with mutable enum issue so I have mapped them with AtomicLong and used it to store the metric's value. Does that seem like a good alternative ?

arunkumarchacko: This is practically a global mutable variable and testing this will be difficult. Hence I do not think this is the right thing to do.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStreamTest.java
```diff
@@ -278,6 +283,89 @@ public void hsync_statistics() throws IOException {
     assertThat(fout.getIOStatistics().counters().get(INVOCATION_HFLUSH.getSymbol())).isEqualTo(1);
   }
 
+  @Test
+  public void hsync_increment_timeStatistics() throws IOException {
+    Path objectPath = new Path(ghfs.getUri().resolve("/dir/object2.txt"));
+    FileSystem.Statistics statistics = new FileSystem.Statistics(ghfs.getScheme());
+    GoogleHadoopOutputStream fout =
```
arunkumarchacko, line 290: consider reducing code duplication in this file

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -138,6 +143,10 @@ public synchronized void seek(long pos) throws IOException {
       channel.position(pos);
     } catch (IllegalArgumentException e) {
       throw new IOException(e);
+    } finally {
+      GhfsTimeStatistic.incrementTimeElapsedMicrosec(
```
arunkumarchacko, line 147: can this throw exception? If yes, better to handle it since otherwise the original exception will be lost.

arunkumarchacko (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -516,30 +530,37 @@ public FSDataOutputStream create(
       long blockSize,
       Progressable progress)
       throws IOException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
     incrementStatistic(GhfsStatistic.INVOCATION_CREATE);
-    checkArgument(hadoopPath != null, "hadoopPath must not be null");
-    checkArgument(replication > 0, "replication must be a positive integer: %s", replication);
-    checkArgument(blockSize > 0, "blockSize must be a positive integer: %s", blockSize);
+    try {
```
medb, line 535: It maybe more readable and maintainable if we extract main logic in the `*Internal` methods and wrap calls to them in try-catch blocks:

```java
@Override
... create(. . .) {
  Stopwatch stopwatch = Stopwatch.createStarted();
  try {
    createInternal(. . .);
  } finally {
     GhfsTimeStatistic.incrementTimeElapsedMicrosec(
          CREATE, stopwatch.elapsed(TimeUnit.MICROSECONDS));
  }
}
```

Not sure that it will be more readable but an option to consider.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -516,30 +530,37 @@ public FSDataOutputStream create(
...
+                  CreateFileOptions.builder().setOverwriteExisting(overwrite).build(),
+                  /* append= */ false,
+                  /* minSyncInterval= */ Duration.ofMillis(
+                      GCS_OUTPUT_STREAM_SYNC_MIN_INTERVAL_MS.get(getConf(), getConf()::getInt)),
+                  statistics),
+              statistics);
+      instrumentation.fileCreated();
+      return response;
+    } finally {
+      GhfsTimeStatistic.incrementTimeElapsedMicrosec(
```
medb, line 560: I think that these stats should be FS instance-specific, not global for JVM.

medb (changes requested):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStreamTest.java
guljain: Created a single function to test all 3 statistics

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
guljain: Since I used an existing lambda, I think it should be more readable than before.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -16,16 +16,87 @@
...
     super(NAME, "Ghfs", ioStatistics);
+    this.ioStatistics = ioStatistics;
+  }
+
+  /**
+   * Map of minimums
+   *
+   * @return current map of minimums
+   */
+  private Map<String, Long> minimums() {
```
arunkumarchacko, line 51: nit: Please move private methods below public methods.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsTimeStatistic.java
arunkumarchacko: Please open a bug for this and assign to me. Also add a TODO comment in the code with the bug number.

arunkumarchacko (approved):

PR #820, opened 2022-07-11
Minor fixes and updates

yigress:
Minor string formats and annotation fixes on null checks and overrides.
Add in ubla bucket option.

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTestUtils.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTestUtils.java
@@ -16,8 +16,6 @@ package com.google.cloud.hadoop.gcsio;
 
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
-import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.gson.GsonFactory;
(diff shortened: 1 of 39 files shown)


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTestUtils.java
```diff
@@ -28,6 +30,7 @@
 public final class GoogleCloudStorageTestUtils {
 
   public static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
+  public static final JsonFactory JSON_FACTORY = GsonFactory.getDefaultInstance();
```
medb, line 33: Seems like we can delete this code

medb (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTestUtils.java
yigress: this is needed by the added test, the test is marked ignored as it fails 
https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/820/files#diff-9323db8d8583cefd2a02a3a84dfd7b5139e4ebb5c97a4b9f263dc556dc471f63R676

medb (approved):

PR #828, opened 2022-07-19
Dynamically compute HTTP requests statistic keys

medb:
This change simplifies implementation (no need to maintain enum for all possible HTTP methods that can change in the future) and improves `GoogleCloudStorageImpl` initialization speed at cost of string concatenation which should not be performance-critical taking into account IO latencies.


abmodi (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StatisticsTrackingHttpRequestInitializer.java
```diff
@@ -86,33 +53,12 @@ private void setHttpRequestFailureStats(HttpResponse response) {
...
-      case "POST":
-        increment(HTTP_POST_REQUEST_FAILURE);
-        break;
-      case "PUT":
-        increment(HTTP_PUT_REQUEST_FAILURE);
-        break;
-      default:
-        throw new IllegalStateException("Unknown request method: " + request.getRequestMethod());
-    }
+    String requestMethod = Ascii.toUpperCase(response.getRequest().getRequestMethod());
```
arunkumarchacko, line 56: nit: consider refactoring this logic to reduce the code duplication.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StatisticsTrackingHttpRequestInitializer.java
```diff
@@ -56,28 +43,8 @@ public void initialize(HttpRequest request) throws IOException {
...
-        increment(HTTP_POST_REQUEST);
-        break;
-      case "PUT":
-        increment(HTTP_PUT_REQUEST);
-        break;
-      default:
-        throw new IllegalStateException("Unknown request method: " + request.getRequestMethod());
-    }
+    String requestMethod = Ascii.toUpperCase(request.getRequestMethod());
+    increment("HTTP_" + requestMethod + "_REQUEST");
```
arunkumarchacko, line 47: nit: String.format might be bit faster.

arunkumarchacko (approved):
Can you add a description explaining WHY we need this change?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StatisticsTrackingHttpRequestInitializer.java
medb: `String.format` is actually slower than string concatenation: https://stackoverflow.com/a/31061425/3227693

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StatisticsTrackingHttpRequestInitializer.java
medb: Didn't find elegant way to minimize code duplication, this seems like a sweet spot.

PR #832, opened 2022-07-20
Adding support for readFully API

abmodi:
FSInputStream exposes readFully API to read entire content of file. With this implementation, we avoid unnecessary seeks while reading entire content.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
@@ -23,12 +23,14 @@ import static java.lang.Math.max;
 import com.google.cloud.hadoop.gcsio.FileInfo;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystem;
 import com.google.common.flogger.GoogleLogger;
+import java.io.EOFException;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SeekableByteChannel;
 import javax.annotation.Nonnull;
+import org.apache.hadoop.fs.FSExceptionMessages;
 import org.apache.hadoop.fs.FSInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.statistics.IOStatistics;
@@ -101,9 +103,7 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
   public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws IOException {
     streamStatistics.readOperationStarted(getPos(), length);
     checkNotNull(buf, "buf must not be null");
-    if (offset < 0 || length < 0 || length > buf.length - offset) {
-      throw new IndexOutOfBoundsException();
-    }
+    validatePositionedReadArgs(getPos(), buf, offset, length);
     int response = 0;
     try {
       // TODO(user): Wrap this in a while-loop if we ever introduce a non-blocking mode for the
@@ -122,6 +122,32 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
     streamStatistics.bytesRead(max(response, 0));
     streamStatistics.readOperationCompleted(length, max(response, 0));
     return response;
+  }
+
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    streamStatistics.readFullyOperationStarted(position, length);
+    checkNotNull(buffer, "buf must not be null");
+    validatePositionedReadArgs(position, buffer, offset, length);
+    if (length == 0) {
+      return;
+    }
+    int nread = 0;
+    synchronized (this) {
+      long oldPos = getPos();
+      try {
+        seek(position);
+        while (nread < length) {
+          int nbytes = read(buffer, offset + nread, length - nread);
+          if (nbytes < 0) {
+            throw new EOFException(FSExceptionMessages.EOF_IN_READ_FULLY);
+          }
+          nread += nbytes;
+        }
+      } finally {
+        seek(oldPos);
+      }
+    }
   }
 
   @Override
--- a/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamIntegrationTest.java
+++ b/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamIntegrationTest.java
@@ -24,6 +24,7 @@ import java.net.URI;
 import java.nio.channels.ClosedChannelException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import org.apache.hadoop.fs.FSExceptionMessages;
 import org.apache.hadoop.fs.FileSystem;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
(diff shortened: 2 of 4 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
     return response;
   }
 
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    streamStatistics.readFullyOperationStarted(position, length);
```
arunkumarchacko, line 129: why this is done before the validation?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
...
+    checkNotNull(buffer, "buf must not be null");
```
arunkumarchacko, line 130: consider adding additional details like position, offset and length

arunkumarchacko: Also this check is not done by validatePositionedReadArgs?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
...
+        seek(position);
+        while (nread < length) {
+          int nbytes = read(buffer, offset + nread, length - nread);
+          if (nbytes < 0) {
+            throw new EOFException(FSExceptionMessages.EOF_IN_READ_FULLY);
+          }
+          nread += nbytes;
+        }
+      } finally {
+        seek(oldPos);
```
arunkumarchacko, line 148: can this throw exception? if yes, can it mask the original?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
...
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    streamStatistics.readFullyOperationStarted(position, length);
+    checkNotNull(buffer, "buf must not be null");
+    validatePositionedReadArgs(position, buffer, offset, length);
+    if (length == 0) {
+      return;
+    }
+    int nread = 0;
+    synchronized (this) {
```
arunkumarchacko, line 136: why synchronized is required for this and not read(). Good to add a comment explaining WHY

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
     return response;
   }
 
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    streamStatistics.readFullyOperationStarted(position, length);
+    checkNotNull(buffer, "buf must not be null");
+    validatePositionedReadArgs(position, buffer, offset, length);
```
arunkumarchacko, line 131: why is this check not inside synchronized block? i.e why not make the method itself synchronized like other methods?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
     return response;
   }
 
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
```
arunkumarchacko, line 128: log.finer()?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +124,32 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
...
+      return;
+    }
+    int nread = 0;
+    synchronized (this) {
+      long oldPos = getPos();
+      try {
+        seek(position);
+        while (nread < length) {
+          int nbytes = read(buffer, offset + nread, length - nread);
+          if (nbytes < 0) {
```
arunkumarchacko, line 142: if it is 0, it is not EOF?

arunkumarchacko (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: No.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: This should be the protocol - statistics are for number of time API is being called not for how many times it was successful.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: validatePositionedRead doesn't check for null buffer. Rest of checks are done by validatePositionedReadArgs.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: You should synchronize the minimal critical block possible. I am not sure what requires extra explanation here. Unnecessary comments are anti-pattern.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: You should synchronize minimal block that requires synchronization.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: Can this cause infinite loop?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -145,7 +145,11 @@ public void readFully(long position, byte[] buffer, int offset, int length) thro
           nread += nbytes;
         }
       } finally {
-        seek(oldPos);
+        try {
+          seek(oldPos);
+        } catch (IOException ie) {
+          logger.atFiner().log("Ignoring IOE on seek to (%d): (%s)", oldPos, ie.getMessage());
```
arunkumarchacko, line 151: I think this should be an error

arunkumarchacko: Also now the Stream might be in an inconsistent state. I am not sure ignoring this is the right thing to do.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Fixed it.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Yes - if your server is not returning data, anyways it will be infinite loop for application. By default today also readFully can cause infinite waiting if server keeps on returning 0 bytes.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Not really - if you look into the implementation of seek - it will only throw error if your position is not correct. That will never happen in this scenario as we are setting it to previous old position only. That's why this exception should never occcur.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Statistics are for API calls and not for correct API calls.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: These are counters - so not sure how actual order matters here. And there are always tradeoffs and performance is critical in this case.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamIntegrationTest.java
```diff
@@ -107,6 +108,44 @@ public void testAvailable() throws Exception {
...
+
+    byte[] value = new byte[5];
+    byte[] expected = Arrays.copyOfRange(testContent.getBytes(StandardCharsets.UTF_8), 2, 7);
+
+    GoogleHadoopFSInputStream in = createGhfsInputStream(ghfs, path);
+    try (GoogleHadoopFSInputStream ignore = in) {
+      in.readFully(2, value);
+      assertThat(in.getPos()).isEqualTo(0);
+    }
+    assertThat(value).isEqualTo(expected);
```
medb, line 129: May you add test that demonstrates impact of this change (we have tests that assert number of GCS requests). For example, do we expect that number of the HTTP JSON GCS requests for `readFully` calls will be reduced after this change?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -124,6 +122,35 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
     return response;
   }
 
+  @Override
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
```
medb, line 126: PR comment mentions that this allows to avoid unnecessary `seek`. May you also clarify why is this important? Is this because we are sending new GCS request for each `seek`?

medb (changes requested):

PR #833, opened 2022-07-21
Disabling failFastOnRead by default

abmodi:
FailFastOnRead adds extra metadata calls for each read call and that is a huge overhead. Since Hive/Spark does file listing before reads, we don't need extra metadata calls for each read operation.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
@@ -39,7 +39,7 @@ public abstract class GoogleCloudStorageReadOptions {
   public static final double DEFAULT_BACKOFF_MULTIPLIER = 1.5;
   public static final int DEFAULT_BACKOFF_MAX_INTERVAL_MILLIS = 10 * 1000;
   public static final int DEFAULT_BACKOFF_MAX_ELAPSED_TIME_MILLIS = 2 * 60 * 1000;
-  public static final boolean DEFAULT_FAST_FAIL_ON_NOT_FOUND_ENABLED = true;
+  public static final boolean DEFAULT_FAST_FAIL_ON_NOT_FOUND_ENABLED = false;
   public static final boolean DEFAULT_GZIP_ENCODING_SUPPORT_ENABLED = true;
   public static final long DEFAULT_INPLACE_SEEK_LIMIT = 8 * 1024 * 1024;
   public static final Fadvise DEFAULT_FADVISE = Fadvise.SEQUENTIAL;


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
```diff
@@ -39,7 +39,7 @@ public enum Fadvise {
   public static final double DEFAULT_BACKOFF_MULTIPLIER = 1.5;
   public static final int DEFAULT_BACKOFF_MAX_INTERVAL_MILLIS = 10 * 1000;
   public static final int DEFAULT_BACKOFF_MAX_ELAPSED_TIME_MILLIS = 2 * 60 * 1000;
-  public static final boolean DEFAULT_FAST_FAIL_ON_NOT_FOUND_ENABLED = true;
+  public static final boolean DEFAULT_FAST_FAIL_ON_NOT_FOUND_ENABLED = false;
```
arunkumarchacko, line 42: is this a breaking change? Should this be part of release notes?

arunkumarchacko (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
abmodi: We will do it as part of new release. Is there any section for release notes.

medb: Usually we include docs and release notes changes in the same PR as change itself:
- https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CHANGES.md
- https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CONFIGURATION.md#io-configuration

medb, line 42: Is this change safe for all frameworks? If we want to make it only for Spark, then it maybe better to change override this default on Dataproc image Spark config.

medb (changes requested):

medb:
I think that #838 PR fixes the same issue for gRPC read channel.

Closing, please reopen if necessary.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadOptions.java
abmodi: I think we should do it in spark config.

PR #835, opened 2022-07-21
Seek should throw IOException instead of IllegalArgumentException

abmodi:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -742,16 +742,28 @@ public SeekableByteChannel position(long newPosition) throws IOException {
...
+  /** Validates that the given position is valid for this channel. */
+  protected void validatePosition(long position, long size) throws IOException {
+    if (position < 0) {
+      throw new EOFException(
+          String.format(
+              "Invalid seek offset: position value (%d) must be >= 0 for '%s'",
+              position, resourceId));
+    }
+
+    if (size >= 0 && position >= size) {
```
arunkumarchacko, line 759: Can size be less than 0? i.e. isn't the second condition enough?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -742,16 +742,28 @@ public SeekableByteChannel position(long newPosition) throws IOException {
...
-        newPosition);
+    validatePosition(newPosition, size());
     positionForNextRead = newPosition;
     return this;
   }
 
+  /** Validates that the given position is valid for this channel. */
+  protected void validatePosition(long position, long size) throws IOException {
+    if (position < 0) {
+      throw new EOFException(
```
arunkumarchacko, line 753: Should be throw IOException instead? @

abmodi: EOFException extends IOException only.

arunkumarchacko: Yes, but this is not really and EndOfFile case, correct? Why you think throwing an ArugmentException is not the right option here?

abmodi: Because that is a RuntimeException and clients are not expected to catch it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
abmodi: I don't think size should be less than 0, but still better to have it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
arunkumarchacko: Can the position be negative, unless there is some bug? I could not think of a case where position can be negative, unless there is a bug and hence it is ok to crash in that case.

I will leave it to you to decide.

arunkumarchacko (approved):

PR #838, opened 2022-07-25
Clean up fail fast on not found logic

medb:
This change also removes extra `getItemInfo` call in gRPC read channel

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -716,14 +716,7 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
     checkArgument(
         resourceId.isStorageObject(), "Expected full StorageObject id, got %s", resourceId);
 
-    // The underlying channel doesn't initially read data, which means that we won't see a
-    // FileNotFoundException until read is called. As a result, in order to find out if the
-    // object exists, we'll need to do an RPC (metadata or data). A metadata check should be a less
-    // expensive operation than a read data operation.
-    GoogleCloudStorageItemInfo itemInfo =
-        readOptions.isFastFailOnNotFoundEnabled() ? getItemInfo(resourceId) : null;
-
-    return open(resourceId, itemInfo, readOptions);
+    return open(resourceId, /* itemInfo= */ null, readOptions);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -752,25 +745,16 @@ private SeekableByteChannel open(
...
-              storageStubProvider,
-              itemInfo,
-              watchdog,
-              metricsRecorder,
-              readOptions,
-              BackOffFactory.DEFAULT,
-              this.storageOptions);
+      return new GoogleCloudStorageGrpcReadChannel(
+          storageStubProvider,
+          itemInfo == null ? getItemInfo(resourceId) : itemInfo,
```
abmodi, line 752: This PR is just cleaning up the metadata call but not adding support for failFast logic for gRPC. Is my understanding correct?

medb: Yes, I think that gRPC channel as it written right now always requires metadata call/info - this can be improved as follow up, if possible.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -177,8 +172,8 @@ public GoogleCloudStorageReadChannel(
 
     // Initialize metadata if available.
     GoogleCloudStorageItemInfo info = getInitialMetadata();
-    if (info != null) {
-      initMetadata(info);
+    if (info != null || readOptions.isFastFailOnNotFoundEnabled()) {
```
singhravidutt, line 175: Why is this required? IIUC, `info` can only be `null` iff fastFail is `disabled`. If `fastFail` is `enabled` then either client will get a `FileNotFoundException` or channel with NOT NULL `info`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -225,7 +212,7 @@ ExponentialBackOff createBackOff() {
    */
   @Nullable
   protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
```
singhravidutt, line 214: Is there any implication of removing this all together?

singhravidutt: nvm, this function is getting override too in `GoogleCloudStorageImpl.java`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -716,14 +716,7 @@ public SeekableByteChannel open(
     checkArgument(
         resourceId.isStorageObject(), "Expected full StorageObject id, got %s", resourceId);
 
-    // The underlying channel doesn't initially read data, which means that we won't see a
```
singhravidutt, line 719: Thank you for cleaning this up.

singhravidutt (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb: np

PR #839, opened 2022-07-28
Fail fast on read operations when underlying stream is closed.

abmodi:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -193,4 +203,16 @@ public int available() throws IOException {
...
+
+  /**
+   * Verify that the input stream is open. Non blocking; this gives the last state of the volatile
+   * {@link #closed} field.
+   *
+   * @throws IOException if the connection is closed.
+   */
+  private void checkNotClosed() throws IOException {
+    if (closed) {
+      throw new IOException(gcsPath + ": " + FSExceptionMessages.STREAM_IS_CLOSED);
```
arunkumarchacko, line 215: should this be ClosedChannelException?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -126,6 +134,7 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
 
   @Override
   public synchronized void seek(long pos) throws IOException {
+    checkNotClosed();
```
arunkumarchacko, line 137: should we add the same check for getPos() as well?

arunkumarchacko (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Yes - I tried this - it doesn't take a string as an argument. That's why I decided to go ahead with IOException.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
abmodi: Done.

PR #841, opened 2022-08-01
Do HTTP request tracing based on tracelogenable flag

arunkumarchacko:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -301,8 +303,10 @@ public GoogleCloudStorageImpl(
    * @param storage {@link Storage} to use for I/O.
    */
   @SuppressWarnings("CheckReturnValue")
+  @VisibleForTesting
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage) {
-    this(options, storage, /* downscopedAccessTokenFn= */ null);
+    this(options, storage, null, tryGetCredentialsFromStorage(storage));
+    warnIfTracingEnabled(storageOptions.isTraceLogEnabled());
```
abmodi, line 309: We should clean it up in future because this constructor should never be used.

abmodi (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -872,6 +870,78 @@ public void deleteObjects(List<StorageResourceId> fullObjectNames) throws IOExce
     }
   }
 
+  private void createGrpcStubIfEnabled(GoogleCloudStorageOptions options, Credentials credentials) {
+    if (!this.storageOptions.isGrpcEnabled()) {
```
medb, line 874: We usually use `this.` prefix only in constructors or setters when there is a conflict between class fields and method parameters

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -872,6 +870,78 @@ public void deleteObjects(List<StorageResourceId> fullObjectNames) throws IOExce
...
+      return;
+    }
+
+    this.watchdog =
+        Watchdog.create(Duration.ofMillis(options.getGrpcMessageTimeoutCheckInterval()));
```
medb, line 879: This is a non-obvious side effect for create method, can we move `watchdog` inside the gRPC stub object, for example?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -872,6 +870,78 @@ public void deleteObjects(List<StorageResourceId> fullObjectNames) throws IOExce
...
+
+    Credentials theCredentials = credentials;
```
medb, line 881: Do we really need `the` prefix? Usually we just use `credentials`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -872,6 +870,78 @@ public void deleteObjects(List<StorageResourceId> fullObjectNames) throws IOExce
...
+    if (theCredentials == null) {
+      checkArgument(
+          httpRequestInitializer instanceof RetryHttpInitializer,
+          "request initializer must be an instance of the RetryHttpInitializer class or credential information should be supplied when gRPC API enabled");
+      theCredentials = ((RetryHttpInitializer) httpRequestInitializer).getCredentials();
+    }
+
+    this.storageStubProvider =
```
medb, line 889: Probably it will be better if we return stub here and move watchdog in the stub

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -872,6 +870,78 @@ public void deleteObjects(List<StorageResourceId> fullObjectNames) throws IOExce
...
+        StorageStubProvider.newInstance(
+            this.storageOptions, this.backgroundTasksThreadPool, theCredentials);
```
medb, line 891: nit: remove `this.` prefixes.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
medb, line 309: It seems that logic in the constructors of this class becomes more and more complicated and hard to follow. I think that we should bite a bullet and refactor it by using Builder class (or possible factory methods, but this seems less optimal) to provide input parameters for this class initialization instead of maintaining multiple contradictory constructors.

medb (changes requested):

medb (approved):

PR #843, opened 2022-08-01
InvocationId for gRPC calls

singhravidutt:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall.SimpleForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import java.util.UUID;
+
+public final class InvocationIdGrpcChannelInterceptor implements ClientInterceptor {
```
arunkumarchacko, line 12: Please add copyright and also javadoc

arunkumarchacko: Does this have to be public?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+      super(call);
+    }
+
+    @Override
+    public void start(Listener<RespT> responseListener, Metadata headers) {
+      appendInvocationId(headers);
+      super.start(responseListener, headers);
+    }
+
+    private void appendInvocationId(final Metadata header) {
```
arunkumarchacko, line 42: Please follow the style used in the project. i.e. we do not include final keyword for parameters and local variables.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptorTest.java
```diff
@@ -0,0 +1,131 @@
+package com.google.cloud.hadoop.gcsio;
+
+import static com.google.common.truth.Truth.assertThat;
```
arunkumarchacko, line 3: include copyright header

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall.SimpleForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import java.util.UUID;
+
+public final class InvocationIdGrpcChannelInterceptor implements ClientInterceptor {
+
+  static final Metadata.Key<String> GOOG_API_CLIENT =
```
arunkumarchacko, line 14: add VisibleForTesting annotation?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+      Metadata.Key.of("x-goog-api-client", Metadata.ASCII_STRING_MARSHALLER);
+
+  static final String GCCL_INVOCATION_ID = "gccl-invocation-id";
```
arunkumarchacko, line 17: what is gccl?

arunkumarchacko: rename this to GCCL_INVOCATION_ID_PREFIX since this is not really the invocationId

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    return new InvocationIdAttachingClientCall<>(next.newCall(method, callOptions));
+  }
+
+  // Non private to avoid synthetic class
+  InvocationIdGrpcChannelInterceptor() {}
+
+  private final class InvocationIdAttachingClientCall<ReqT, RespT>
```
arunkumarchacko, line 28: can this be a static class?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+      extends SimpleForwardingClientCall<ReqT, RespT> {
+
+    // Non private to avoid synthetic class
+    InvocationIdAttachingClientCall(ClientCall<ReqT, RespT> call) {
+      super(call);
+    }
+
+    @Override
+    public void start(Listener<RespT> responseListener, Metadata headers) {
```
arunkumarchacko, line 37: why doing this onStart() and not onHeaders()?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+      appendInvocationId(headers);
+      super.start(responseListener, headers);
```
arunkumarchacko, line 39: why are we doing this before appendInvocationId(headers)?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private void appendInvocationId(final Metadata header) {
+      UUID uuid = UUID.randomUUID();
+      String invocationId = String.format("%s/%s", GCCL_INVOCATION_ID, uuid.toString());
+      final String value = header.get(GOOG_API_CLIENT);
```
arunkumarchacko, line 45: nit: same here. We do not use final as per the style.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+      if (value != null) {
+        // First remove the value from Metadata, as overrides are not supported
+        header.remove(GOOG_API_CLIENT, value);
+        // append space separated invocationId
+        String appendedValue = value + " " + invocationId;
+        header.put(GOOG_API_CLIENT, appendedValue);
```
arunkumarchacko, line 51: Is this the recommended approach? I did not see the key getting overwritten, from the implementation.

Also what will happen if another interceptor "puts" this header.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
```diff
@@ -63,6 +63,7 @@ private ManagedChannel buildManagedChannel() {
         .createChannelBuilder(target)
         .enableRetry()
         .userAgent(options.getAppName())
+        .intercept(new InvocationIdGrpcChannelInterceptor())
```
arunkumarchacko, line 66: can you add a comment explaining why we are adding a Channel interceptor and not a stub interceptor?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
```diff
@@ -463,20 +485,17 @@ public void retryInsertOnIOException() throws Exception {
...
-        newWriteChannel(options, writeConditions, /* requesterPaysProject= */ null);
+        newWriteChannel(
+            options,
+            writeConditions,
+            null,
+            () -> new RetryBoundedBackOff(new ExponentialBackOff(), /* maxRetries= */ 1),
+            false);
     fakeService.setInsertObjectExceptions(
         ImmutableList.of(
-            new StatusException(Status.DEADLINE_EXCEEDED),
```
arunkumarchacko, line 469: why are we changing this?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptorTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.ArgumentMatchers;
+
+@RunWith(JUnit4.class)
+public class InvocationIdGrpcChannelInterceptorTest {
+
+  private InvocationIdGrpcChannelInterceptor invocationIdInterceptor;
+
+  private Channel channel = mock(Channel.class);
```
arunkumarchacko, line 29: is there a way to avoid mock? We try to avoid mock as much as possible.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
```diff
@@ -117,7 +119,11 @@ public void setUp() throws Exception {
         .build()
         .start();
     stub =
-        StorageGrpc.newStub(InProcessChannelBuilder.forName(serverName).directExecutor().build());
+        StorageGrpc.newStub(
+            InProcessChannelBuilder.forName(serverName)
+                .intercept(new InvocationIdGrpcChannelInterceptor())
```
arunkumarchacko, line 124: why can't we use MetadataUtils.newAttachHeadersInterceptor?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
medb, line 12: Add javadoc

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: Updated.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptorTest.java
singhravidutt: Added

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: Using `MetadataUtils.HeaderAttachingClientInterceptor` as a reference which is also used to append headers. Why not using `HeaderAttachingClientInterceptor`? because we want to run our own custom logic about how we want insert invocationIds.

`onHeader` is to intercept the response. Here we are intercepting the request for which `start` is used.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: > why are we doing this before appendInvocationId(headers)?

I don't understand what is asked here.

We have to append the custom logic before making call to actual `start` which will eventually send the request over channel.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: That is correct keys are not overwritten. That's what I commented too. Metadata don't verify if key is already there or not and just appends it at the end of an array of headers.

Just trying to be in sync with how this is being done as library (manual client) where we are appending it to existing header instead of duplicating the key.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
singhravidutt: Adding InvocationId to header is generic thing which will apply to any RPC call made from gRPC client. That's the reason it's not added to a specific stub which could differ for different RPCs.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
singhravidutt: This UT was broken where we are trying to test the retries but retry strategy `BackOff.STOP_BACKOFF` which will not retry anything and this UT was no different that any other UT in this class.
As you can see I updated the retry strategy and hence updated the number of calls to this `fakeService`.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptorTest.java
singhravidutt: Not sure why? In UT's we should mock all the dependent calls and just test the `unit` isn't it?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
singhravidutt: explained above.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: Updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
singhravidutt: I am not sure if it is even required here. I am able to test the `private` functionality using the public functions.

arunkumarchacko: Not required since we made this package private

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptor.java
arunkumarchacko: For this to work the "value" in the header cannot have  space correct? I am not sure if there is any such restrictions.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageStubProvider.java
arunkumarchacko: Can we add this as a comment in the code since this is not trivial and would help future readers of the code.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcWriteChannelTest.java
arunkumarchacko: Can we have a separate CL for that? It is better if we do not club unrelated changes in the same CL.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/InvocationIdGrpcChannelInterceptorTest.java
arunkumarchacko: Mocks are hard to maintain. Refer GoogleCloudStorageGrpcReadChannelTest. We are using FakeService. Why cant we do something similar?

PR #844, opened 2022-08-02
fAdvise should be updated in second read.

abmodi:
CSV files can be opened at different offsets. gRPC flow updates fAdvise to Random even when first read is beyond InPlaceSeekLimit.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
@@ -742,14 +742,26 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
     if (!isOpen()) {
       throw new ClosedChannelException();
     }
-    checkArgument(newPosition >= 0, "Read position must be non-negative, but was %s", newPosition);
-    checkArgument(
-        newPosition < size(),
-        "Read position must be before end of file (%s), but was %s",
-        size(),
-        newPosition);
+    validatePosition(newPosition, size());
     positionForNextRead = newPosition;
     return this;
+  }
+
+  /** Validates that the given position is valid for this channel. */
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -104,7 +104,7 @@ public class GoogleCloudStorageGrpcReadChannel implements SeekableByteChannel {
   private boolean channelIsOpen = true;
 
   // Current position in the object.
-  private long positionInGrpcStream;
+  private long positionInGrpcStream = -1;
```
arunkumarchacko, line 107: Wouldn't this cause the index to go out of sync since at many places we are using +=.

eg: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java#L530

It might be better to have another variable to track this rather than overloading this variable. Single Responsibility  i.e.

abmodi: The reason I used this is because of same reason. This should tell us the positionInGrpcStream. Until we open the grpcStream position is undefined (denoted by -1). Once we open it, this variable is set to correct position. Similar implementation is present in json implementation.

arunkumarchacko: The sequential coupling here is the one that worries me. Can we add some asserts where we expect this to be initialized?

abmodi: Sure - let me do that.

arunkumarchacko (approved):

PR #846, opened 2022-08-04
Using common defaults in gcs and gcsio

abmodi:


arunkumarchacko (commented):
Do we need any testing for this since this can have some ramifications. Also if you cant include some rationale behind changing some of the defaults, it would be helpful.

arunkumarchacko (approved):
Discussed offline with Abhishek. We are not changing the defaults, but making GCSIO defaults the same as GCS defaults.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -16,6 +16,28 @@
 
 package com.google.cloud.hadoop.fs.gcs;
 
+import static com.google.cloud.hadoop.gcsio.GoogleCloudStorageOptions.BATCH_THREADS_DEFAULT;
```
medb, line 19: Probably we need to converge on the default vars naming and prefix all of them with `DEFAULT_` prefix for consistency?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -326,11 +356,12 @@ public class GoogleHadoopFileSystemConfiguration {
 
   /** Configuration key for enabling use of the gRPC API for read/write. */
   public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_ENABLE =
-      new HadoopConfigurationProperty<>("fs.gs.grpc.enable", false);
+      new HadoopConfigurationProperty<>("fs.gs.grpc.enable", ENABLE_GRPC_DEFAULT);
```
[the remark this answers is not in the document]
medb: +1

medb (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
abmodi: All these are public constants previously defined. I will change for gRPC variables as it is not being.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -293,27 +320,30 @@ public class GoogleHadoopFileSystemConfiguration {
...
-   * If true, reading a file with GZIP content encoding (HTTP header "Content-Encoding: gzip") will
+   * If false, reading a file with GZIP content encoding (HTTP header "Content-Encoding: gzip") will
    * result in failure (IOException is thrown).
    */
   public static final HadoopConfigurationProperty<Boolean>
       GCS_INPUT_STREAM_SUPPORT_GZIP_ENCODING_ENABLE =
           new HadoopConfigurationProperty<>(
-              "fs.gs.inputstream.support.gzip.encoding.enable", false);
+              "fs.gs.inputstream.support.gzip.encoding.enable",
+              DEFAULT_GZIP_ENCODING_SUPPORT_ENABLED);
```
abmodi, line 334: @medb  could you please look into this - I have enabled gzip by default. Previously, it was disabled here but was enabled in gcsio. Does it look correct?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -339,28 +370,33 @@ public class GoogleHadoopFileSystemConfiguration {
 
   /** Configuration key for check interval (in millisecond) for gRPC request timeout to GCS. */
   public static final HadoopConfigurationProperty<Long> GCS_GRPC_CHECK_INTERVAL_TIMEOUT_MS =
-      new HadoopConfigurationProperty<>("fs.gs.grpc.checkinterval.timeout.ms", 1_000L);
+      new HadoopConfigurationProperty<>(
+          "fs.gs.grpc.checkinterval.timeout.ms", GRPC_MESSAGE_TIMEOUT_CHECK_INTERVAL);
```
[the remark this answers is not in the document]
abmodi: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
abmodi: Disabled it by default now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
```diff
@@ -35,10 +35,10 @@
...
   /** Default setting for enabling use of GCS gRPC API. */
-  public static final boolean ENABLE_GRPC_DEFAULT = false;
+  public static final boolean DEFAULT_GRPC_ENABLED = false;
 
   /** Default setting for enabling use of the Traffic Director for GCS gRPC API. */
-  public static final boolean ENABLE_TRAFFIC_DIRECTOR_DEFAULT = true;
+  public static final boolean DEFAULT_TRAFFIC_DIRECTOR_ENABLED = true;
 
   /** Default setting to prefer DirectPath for gRPC. */
   public static final boolean DIRECT_PATH_PREFERRED_DEFAULT = true;
```
medb, line 44: Seems like we need to rename all default constants to have `DEFAULT_` prefix

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageMockitoTest.java
```diff
@@ -77,7 +77,13 @@ public void setUp() {
     MockitoAnnotations.initMocks(this);
 
     MockHttpTransport transport = mockTransport();
-    gcs = mockedGcs(transport);
+    final GoogleCloudStorageOptions GCS_OPTIONS =
```
medb, line 80: We usually do not use final inside the methods, and all vars should be camelCase except class-level constants.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/testing/MockGoogleCloudStorageImplFactory.java
```diff
@@ -48,4 +48,20 @@ public static GoogleCloudStorageImpl mockedGcs(HttpTransport transport) {
...
+        new Storage(
+            transport,
+            GsonFactory.getDefaultInstance(),
+            new TrackingHttpRequestInitializer(
+                new RetryHttpInitializer(
+                    new FakeCredentials(),
+                    RetryHttpInitializerOptions.builder()
+                        .setDefaultUserAgent("gcs-io-unit-test")
+                        .build()),
+                false));
```
medb, line 64: I think that this logic could be de-duplicated if you will call this new method inside the above `mockedGcs` method.

medb (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
abmodi: Done. Thanks!

PR #849, opened 2022-08-11
Fixing issue of the read from the buffer from gRPC

abmodi:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -528,10 +528,9 @@ private int readObjectContentFromGCS(ByteBuffer byteBuffer) throws IOException {
       int skipBytes = Math.toIntExact(positionForNextRead - positionInGrpcStream);
       if (skipBytes >= 0 && skipBytes < content.size()) {
         content = content.substring(skipBytes);
-        positionInGrpcStream = positionForNextRead;
+        positionInGrpcStream += skipBytes;
       } else if (skipBytes >= content.size()) {
         positionInGrpcStream += content.size();
-        positionForNextRead = positionInGrpcStream;
```
arunkumarchacko, line 534: it wont hurt if we keep this correct?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -528,10 +528,9 @@ private int readObjectContentFromGCS(ByteBuffer byteBuffer) throws IOException {
...
         return bytesRead;
```
arunkumarchacko, line 534: this will always be 0, correct? Can we change this to 0 and move byteRead variable closer to where it is getting used?

arunkumarchacko (commented):
Can we also add a short description about the bug. It would help future readers of code.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
abmodi: It will - that  would be wrong. positionForNextRead could be higher than positionInGrpcStream

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
abmodi: Will refactor this code.

arunkumarchacko (approved):

PR #850, opened 2022-08-11
adding support for invocationId in Json path

singhravidutt:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
```diff
@@ -169,12 +171,33 @@ public ImmutableList<String> getAllRequestStrings() {
...
+  }
+
   public ImmutableList<String> getAllRawRequestStrings() {
     return requests.stream()
         .map(GoogleCloudStorageIntegrationHelper::requestToString)
         .collect(toImmutableList());
   }
 
+  private String getInvocationId(HttpHeaders header) {
+    String apiClientHeader = (String) header.get(InvocationIdInterceptor.GOOG_API_CLIENT);
```
arunkumarchacko, line 188: can this be null?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private final HttpExecuteInterceptor interceptor;
+
+  public InvocationIdInterceptor(HttpExecuteInterceptor interceptor) {
+    this.interceptor = interceptor;
+  }
+
+  @Override
+  public void intercept(HttpRequest request) throws IOException {
+    if (this.interceptor != null) {
```
arunkumarchacko, line 23: when can this be null?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    String existing = (String) headers.get(GOOG_API_CLIENT);
+    if (isInvocationIdPresent(existing)) {
+      // As invocationId is already present,It's a retried request.
+      // TODO: add support for attempt_count
+      return;
+    }
+    UUID invocationId = UUID.randomUUID();
+    final String signatureKey = "Signature="; // For V2 and V4 signedURLs
+    final String builtURL = request.getUrl().build();
+    if (invocationId != null && !builtURL.contains(signatureKey)) {
```
arunkumarchacko, line 36: can invocationId be null?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
+package com.google.cloud.hadoop.util.interceptors;
```
arunkumarchacko, line 1: Please add copyright header

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.api.client.http.HttpExecuteInterceptor;
+import com.google.api.client.http.HttpHeaders;
+import com.google.api.client.http.HttpRequest;
+import java.io.IOException;
+import java.util.UUID;
+
+public final class InvocationIdInterceptor implements HttpExecuteInterceptor {
```
arunkumarchacko, line 9: can this be not private.

arunkumarchacko: Also add javadoc

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+        newValue = existing + " " + invocationEntry;
+      } else {
+        newValue = invocationEntry;
+      }
+      headers.set("x-goog-api-client", newValue);
+    }
+  }
+
+  private boolean isInvocationIdPresent(String apiClientHeader) {
+    return apiClientHeader != null ? apiClientHeader.contains(GCCL_INVOCATION_ID_PREFIX) : false;
```
arunkumarchacko, line 49: nit: apiClientHeader != null && apiClientHeader.contains(GCCL_INVOCATION_ID_PREFIX) can be a bit more  neater

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+      if (existing != null && !existing.isEmpty()) {
+        newValue = existing + " " + invocationEntry;
+      } else {
+        newValue = invocationEntry;
+      }
+      headers.set("x-goog-api-client", newValue);
+    }
+  }
+
+  private boolean isInvocationIdPresent(String apiClientHeader) {
```
arunkumarchacko, line 48: can this be static?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+      this.interceptor.intercept(request);
+    }
+    HttpHeaders headers = request.getHeaders();
+    String existing = (String) headers.get(GOOG_API_CLIENT);
+    if (isInvocationIdPresent(existing)) {
+      // As invocationId is already present,It's a retried request.
+      // TODO: add support for attempt_count
+      return;
+    }
+    UUID invocationId = UUID.randomUUID();
```
arunkumarchacko, line 33: did we consider having this based on timestamp, rather than having this to be fully random? i.e. using the invocation id, we can get some idea about then the call was made for eg.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko, line 36: can we rely on methods in URL class to check this, rather than doing builtURL.contains?

arunkumarchacko, line 36: why we are doing this only if Signature= is present?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko, line 33: why not move this inside the if?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    final String builtURL = request.getUrl().build();
+    if (invocationId != null && !builtURL.contains(signatureKey)) {
+      String invocationEntry = GCCL_INVOCATION_ID_PREFIX + invocationId;
+      final String newValue;
+      if (existing != null && !existing.isEmpty()) {
+        newValue = existing + " " + invocationEntry;
+      } else {
+        newValue = invocationEntry;
+      }
+      headers.set("x-goog-api-client", newValue);
```
arunkumarchacko, line 44: please use the constant which is already there.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    this.interceptor = interceptor;
+  }
+
+  @Override
+  public void intercept(HttpRequest request) throws IOException {
+    if (this.interceptor != null) {
+      this.interceptor.intercept(request);
+    }
+    HttpHeaders headers = request.getHeaders();
+    String existing = (String) headers.get(GOOG_API_CLIENT);
```
arunkumarchacko, line 27: consider renaming this to googleApiClientHeader

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
```diff
@@ -169,12 +171,33 @@ public ImmutableList<String> getAllRequestStrings() {
         .collect(toImmutableList());
   }
 
+  public ImmutableList<String> getAllRequestInvocationIds() {
```
arunkumarchacko, line 174: where is this getting used?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Sure.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: I don't see any benefit of it.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Done.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: same as above.

singhravidutt: Replicated what we are doing in manual client build by GCS library team https://github.com/googleapis/java-storage/pull/1309/files#diff-803f9af86ecb181c65a6855974a3561f99ef409ad758f92e00890760188e4ea5

singhravidutt: My bad. Will remove null check.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: This is not the requirement. GCS do get timestamp information attached to every request already.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: I guess it's in alignment with what is followed in this package.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: when there is no interceptor.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Sure.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
singhravidutt: I don't expect the  httpHeader to be `null` for any request.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Yeah, Will do.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: added java doc.
It's a `public` class.

arunkumarchacko: I meant, can it be not public

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
arunkumarchacko: I think it is not in our control and better to be defensive in this case.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko: static methods are slightly more efficient.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
arunkumarchacko: We should try to avoid adding code just for enabling testing. Aren't there other alternatives?

arunkumarchacko (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
singhravidutt: I am confused.
This is the code in `test` directory. 
It is serving the purpose of extracting specific information from Http request header.
It is a required piece of code.
____
What is your proposal to `alternatives`?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: No.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
arunkumarchacko: My bad. I misread.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
arunkumarchacko: Ignore this comment since this is for test code.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
```diff
@@ -169,12 +171,33 @@ public ImmutableList<String> getAllRequestStrings() {
...
+    return requests.stream()
+        .map(r -> getInvocationId(r.getHeaders()))
+        .filter(id -> id != null && !id.isEmpty())
```
arunkumarchacko, line 177: nit: can change this to .filter(not(Strings::isNullOrEmpty))

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
```diff
@@ -169,12 +171,33 @@ public ImmutableList<String> getAllRequestStrings() {
...
     return requests.stream()
         .map(GoogleCloudStorageIntegrationHelper::requestToString)
         .collect(toImmutableList());
   }
 
+  private String getInvocationId(HttpHeaders header) {
+    String apiClientHeader = (String) header.get(InvocationIdInterceptor.GOOG_API_CLIENT);
+    int beginIndex = apiClientHeader.indexOf(InvocationIdInterceptor.GCCL_INVOCATION_ID_PREFIX);
+    if (beginIndex >= 0) {
+      beginIndex = beginIndex + InvocationIdInterceptor.GCCL_INVOCATION_ID_PREFIX.length();
```
arunkumarchacko, line 191: it would be helpful if we can provide a sample header as comment.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko: Please add @VisibleForTesting annotation to indicate that the class is public only for testing purposes.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -95,6 +96,7 @@ public void initialize(HttpRequest request) throws IOException {
       headers.setUserAgent(options.getDefaultUserAgent());
     }
     headers.putAll(options.getHttpHeaders());
+    request.setInterceptor(new InvocationIdInterceptor(request.getInterceptor()));
```
arunkumarchacko, line 99: It is not obvious why we are passing one interceptor to another. Is it because this will override the existing interceptor? If yes, please add a comment since it is not very obvious. Also we use ChainingInterceptors at other places for similar scenarios.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+import java.util.UUID;
+
+/**
+ * HTTP request interceptor to attach unique identifier i.e. invocationId to each new request and
+ * make sure to pass in the same identifier if request is retried.
+ */
+public final class InvocationIdInterceptor implements HttpExecuteInterceptor {
+
+  public static final String GCCL_INVOCATION_ID_PREFIX = "gccl-invocation-id/";
```
arunkumarchacko, line 31: Please add @VisibleForTesting

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko: I think there is scope for improving the variable name since the current one is very generic. I will leave it up to you

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+        newValue = existing + " " + invocationEntry;
+      } else {
+        newValue = invocationEntry;
+      }
+      headers.set(GOOG_API_CLIENT, newValue);
+    }
+  }
+
+  private static boolean isInvocationIdPresent(String apiClientHeader) {
+    return apiClientHeader != null && apiClientHeader.contains(GCCL_INVOCATION_ID_PREFIX)
```
arunkumarchacko, line 70: why not just return?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    HttpHeaders headers = request.getHeaders();
+    String existing = (String) headers.get(GOOG_API_CLIENT);
+    if (isInvocationIdPresent(existing)) {
+      // As invocationId is already present,It's a retried request.
+      // TODO: add support for attempt_count
+      return;
+    }
+
+    final String signatureKey = "Signature="; // For V2 and V4 signedURLs
```
arunkumarchacko, line 54: I think you mentioned that we took this logic from another repo. It would be useful to point to that link as comment.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
singhravidutt: Nothing is getting overridden. All interceptors are getting chained via this. Unlike gRPC in HTTP there is no way to provide more than one interceptor for request and this feels like the only way via which we can have multiple interceptors(one after another) act on a single request.

`ChainingInterceptors` which you are referring  to is not a interceptor but an `initializer:ChainingHttpRequestInitializer`  which chains multiple `initializers` and `interceptors` using the same exact logic which I used (or other way round).

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Sure.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Because `invocationId` is not the only thing present in `x-goog-api-client` header. As we are just interested in making sure `invocationId` is present in header hence just boolean.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Sure.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: Sure.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
singhravidutt: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/TrackingHttpRequestInitializer.java
singhravidutt: Sure.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
arunkumarchacko: yes, I was referring about ChainingInterceptors. Please do add a comment explaining that this is doing a chaining interceptor. Also why not make use of existing chaining interceptor?

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
arunkumarchacko: I was wondering why we need the ternary operator. Why can't we just return the boolean value?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
singhravidutt: There is NO `ChainingInterceptors`  there is `ChainingHttpRequestInitializer`. I am not creating another `HTTPInitializer` just an interceptor and hence there is no way(or no need to) I can use `ChainingHttpRequestInitializer`. `Interceptor` and `Initializer` are two different things.
IIUC, If someone is creating a new interceptor and they can add it to any of the `HTTPInitializer` used.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
singhravidutt: aah, My bad I overlooked it.

arunkumarchacko (approved):

PR #854, opened 2022-08-16
Delegating retries to GCS client library

singhravidutt:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2213,33 +2213,8 @@ private boolean canIgnoreExceptionForEmptyObject(
...
-          } catch (InterruptedException e) {
-            // We caught an InterruptedException, we should set the interrupted bit on
-            // this thread.
-            Thread.currentThread().interrupt();
-            nextSleep = BackOff.STOP;
-          }
-        }
-        existingInfo = getItemInfo(resourceId);
-        nextSleep = nextSleep == BackOff.STOP ? BackOff.STOP : backOff.nextBackOffMillis();
-      } while (!existingInfo.exists() && nextSleep != BackOff.STOP);
```
medb, line 2242: I think that whole purpose of this logic is to wait until object will be created potentially by some other worker, i.e. we need to wait when 404 Not Found responses are returned. And it seems like GCS client library will not retry 404 responses, so in this case I think that this will not work as it should/expected.

Do we know that this not necessary logic anymore?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
singhravidutt: I wasn't aware that we are intentionally retrying 404s. Got ome context from https://github.com/GoogleCloudDataproc/hadoop-connectors/commit/7ddca407ecf0be16504d2f117926c9cadd02159e leaving it as it is for now.

Have one question though, dows it make sense to not retry 429s or 503s for `emptyObjectCreation` then?

medb: I think that this is a safeguard that even if 429s and 503 are retried there is still a possibility that all retries will fail, but empty object will be created by some other worker. Not sure how important this is now, maybe we can delete this logic, if there were no retries for 429 and 503 responses when it was added.

PR #863, opened 2022-08-23
Enforce early socket read timeout setting.

cnauroth:
Set socket read timeout (`fs.gs.http.read-timeout`) as early as possible
on new sockets returned from the custom `SSLSocketFactory`. This
guarantees the timeout is enforced during TLS handshakes when using
Conscrypt as the security provider.

See also https://github.com/google/conscrypt/issues/864 .

### DIFF
--- a/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
+++ b/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
@@ -137,6 +137,10 @@ public class HadoopCredentialsConfiguration {
   public static final HadoopConfigurationProperty<RedactedString> PROXY_PASSWORD_SUFFIX =
       new HadoopConfigurationProperty<>(".proxy.password");
 
+  /** Key suffix for setting the read timeout (in milliseconds) for HTTP request. */
+  public static final HadoopConfigurationProperty<Integer> READ_TIMEOUT_SUFFIX =
+      new HadoopConfigurationProperty<>(".http.read-timeout", 20 * 1000);
+
   /**
    * Configuration key for defining the OAUth2 client ID. Required when the authentication type is
    * USER_CREDENTIALS
@@ -313,7 +317,8 @@ public class HadoopCredentialsConfiguration {
             return HttpTransportFactory.createHttpTransport(
                 PROXY_ADDRESS_SUFFIX.withPrefixes(keyPrefixes).get(config, config::get),
                 PROXY_USERNAME_SUFFIX.withPrefixes(keyPrefixes).getPassword(config),
-                PROXY_PASSWORD_SUFFIX.withPrefixes(keyPrefixes).getPassword(config));
+                PROXY_PASSWORD_SUFFIX.withPrefixes(keyPrefixes).getPassword(config),
+                READ_TIMEOUT_SUFFIX.withPrefixes(keyPrefixes).get(config, config::getInt));
           } catch (IOException e) {
             throw new UncheckedIOException(e);
           }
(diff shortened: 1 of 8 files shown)


util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -73,6 +73,32 @@ public static HttpTransport createHttpTransport(
...
+   *     empty, no socket read timeout will be applied.
+   * @return The resulting HttpTransport.
+   * @throws IllegalArgumentException If the proxy address is invalid.
+   * @throws IOException If there is an issue connecting to Google's Certification server.
+   */
+  public static HttpTransport createHttpTransport(
+      @Nullable String proxyAddress,
+      @Nullable RedactedString proxyUsername,
+      @Nullable RedactedString proxyPassword,
+      @Nullable Integer readTimeoutMillis)
```
medb, line 100: Let's change type to `Duration`

util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
```diff
@@ -137,6 +137,10 @@ public class HadoopCredentialsConfiguration {
   public static final HadoopConfigurationProperty<RedactedString> PROXY_PASSWORD_SUFFIX =
       new HadoopConfigurationProperty<>(".proxy.password");
 
+  /** Key suffix for setting the read timeout (in milliseconds) for HTTP request. */
+  public static final HadoopConfigurationProperty<Integer> READ_TIMEOUT_SUFFIX =
+      new HadoopConfigurationProperty<>(".http.read-timeout", 20 * 1000);
```
medb, line 142: Seems like we have this config property declared 2 times now?

medb (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
cnauroth: Done, thank you.

util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
cnauroth: Thanks for catching this. I removed the duplicated definition under `/gcs`.

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -186,12 +216,14 @@ static URI parseProxyAddress(@Nullable String proxyAddress) {
...
-  static class SslKeepAliveSocketFactory extends SSLSocketFactory {
+  static final class CustomSslSocketFactory extends SSLSocketFactory {
 
     private final SSLSocketFactory wrappedSockedFactory;
+    private final Integer readTimeoutMillis;
 
-    public SslKeepAliveSocketFactory(SSLSocketFactory wrappedSocketFactory) {
+    public CustomSslSocketFactory(SSLSocketFactory wrappedSocketFactory, Duration readTimeout) {
       this.wrappedSockedFactory = wrappedSocketFactory;
+      this.readTimeoutMillis = readTimeout != null ? (int) readTimeout.toMillis() : null;
```
medb, line 226: Probably will be safer to use `Math.toExactInt` instead of `(int)` cast

medb (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
cnauroth: Done, thanks!

medb (approved):

cnauroth:
Thanks for the reviews, @medb and @Deependra-Patel !

PR #865, opened 2022-08-30
Enforce early socket read timeout setting.

cnauroth:
Set socket read timeout (`fs.gs.http.read-timeout`) as early as possible
on new sockets returned from the custom `SSLSocketFactory`. This
guarantees the timeout is enforced during TLS handshakes when using
Conscrypt as the security provider.

See also https://github.com/google/conscrypt/issues/864 .

(cherry picked from commit 61f8629ed3aa0c19629e072e3ff6445df6e211e0)

### DIFF
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -1,5 +1,10 @@
 ### 2.2.8 - 2022-XX-XX
 
+1.  Set socket read timeout (`fs.gs.http.read-timeout`) as early as possible on
+    new sockets returned from the custom `SSLSocketFactory`. This guarantees the
+    timeout is enforced during TLS handshakes when using Conscrypt as the
+    security provider.
+
 ### 2.2.7 - 2022-06-01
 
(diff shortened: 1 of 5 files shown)


Deependra-Patel (approved):

cnauroth:
@Deependra-Patel , thank you for the review.

@medb , this was mostly the same as the master branch pull request you approved, so I went ahead with merge based on Deependra's approval. I'm happy to respond if you have any follow-up requests though. Thanks!

scwhittle:
Thanks for fixing this bug Chris! Can you confirm that this will be in the 2.2.8 release and what the expected timeline for the release is? I would like to update Apache Beam dependency once it is released (currently 2.2.6) as we encountered this in a pipeline.

cnauroth:
@scwhittle , yes, this change will ship in release 2.2.8. I'm still working on a timeline for that release. I can give you an update when I know more.

scwhittle:
Great, thanks Chris!

PR #867, opened 2022-09-05
Added Time-Specific Metrics

arunkumarchacko:
Original PR: https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/819


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsInstrumentation.java
```diff
@@ -569,7 +573,6 @@ public void seekForwards(long skipped) {
       if (skipped > 0) {
         bytesSkippedOnSeek.addAndGet(skipped);
       }
-      seekOperations.incrementAndGet();
```
singhravidutt, line 572: Why we are removing this? Similar comment fir `seekBackwards` too.

nit: update the java doc of this function.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
```diff
@@ -82,10 +82,10 @@ public enum GhfsStatistic {
       TYPE_COUNTER),
   INVOCATION_COPY_FROM_LOCAL_FILE(
       StoreStatisticNames.OP_COPY_FROM_LOCAL_FILE, "Calls of copyFromLocalFile()", TYPE_COUNTER),
-  INVOCATION_CREATE(StoreStatisticNames.OP_CREATE, "Calls of create()", TYPE_COUNTER),
+  INVOCATION_CREATE(StoreStatisticNames.OP_CREATE, "Calls of create()", TYPE_DURATION),
```
singhravidutt, line 85: nit: should this be "Duration of.." ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -16,16 +16,88 @@
...
 
+  /** Exention for minimum */
+  private static final String MINIMUM = StoreStatisticNames.SUFFIX_MIN;
+  /** Exention for maximum */
+  private static final String MAXIMUM = StoreStatisticNames.SUFFIX_MAX;
+  /** Exention for mean */
+  private static final String MEAN = StoreStatisticNames.SUFFIX_MEAN;
+
+  /** IOStatistics Instance */
+  private final IOStatistics ioStatistics;
```
singhravidutt, line 39: nit: use of "this" in whole class.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -16,16 +16,88 @@
...
     super(NAME, "Ghfs", ioStatistics);
+    this.ioStatistics = ioStatistics;
+  }
+
+  /**
+   * Map of minimums
+   *
+   * @return current map of minimums
+   */
+  private Map<String, Long> minimums() {
```
singhravidutt, line 52: Do any of these `private` functions adding any value?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
arunkumarchacko: This is pattern that is already getting used. Even otherwise the naming look ok to me.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
arunkumarchacko: We could have inlined this. Having this method is not bad either.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
arunkumarchacko: you mean remove "this"?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
singhravidutt: Yeah, but we changed it from capturing the count to capturing duration of operation. Hence we should fix the description for this too, shouldn't we?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
singhravidutt: Yes.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
arunkumarchacko: Made this consistent with similar existing metrics.

arunkumarchacko (commented):
PTAL

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -106,60 +111,82 @@ public synchronized int read() throws IOException {
...
-        statistics.incrementReadOps(1);
-      }
-      response = numRead;
-    } catch (IOException e) {
-      streamStatistics.readException();
-    }
-    streamStatistics.bytesRead(max(response, 0));
-    streamStatistics.readOperationCompleted(length, max(response, 0));
-    return response;
+    return trackDuration(
```
singhravidutt, line 114: It would be much cleaner if we can use AOP and annotation to capture the "duration" of each API.

singhravidutt (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: Using the pattern from hadoop. Adding AOP will be more work and I do no think it is worth the effort.

PR #870, opened 2022-09-07
Support existing access token when using UserCredentials auth type

### LINKED ISSUE #869: Support existing access token when using UserCredentials auth type
When using the UserCredentials auth type, we can't set a valid access token at the same time.
A bit of a shame if we just did the authorisation code grant flow, which returns an access token with the refresh token.

Ideally, the hadoop connector should use this access token until it expires and only then, refresh the token.

Currently, the connector will refresh straight away.

qcastel:
In Dataiku, we found a small optimisation of the UserCredentials auth type that we thought we could share with you guys.

Saving a bit of network calls to Google OAuth2 provider.

Fixes #869

### DIFF
--- a/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
+++ b/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
@@ -164,6 +164,20 @@ public class HadoopCredentialsConfiguration {
       new HadoopConfigurationProperty<>(".auth.refresh.token");
 
   /**
+   * Configuration key for defining the OAuth2 access token. Optional when the authentication type
+   * is USER_CREDENTIALS
+   */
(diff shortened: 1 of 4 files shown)


util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java
```diff
@@ -341,7 +363,10 @@ private static GoogleCredentials configureCredentials(
     }
     if (credentials instanceof UserCredentials) {
       return ((UserCredentials) credentials)
-          .toBuilder().setTokenServerUri(URI.create(tokenServerUrl)).build();
+          .toBuilder()
+              .setAccessToken(credentials.getAccessToken())
```
qcastel, line 367: I had to do that, to workaround https://github.com/googleapis/google-auth-library-java/pull/993

qcastel:
@medb any news on this one?

qcastel:
@medb Whats your view on this one? Want me to close it as wont do?

PR #886, opened 2022-10-06
adding trace logging for connector and gcs-grpc APIs

singhravidutt:
Using trace logging to capture read API monitored at connector level as well as latency observed by connector from gcs side (for grpc path).


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -18,19 +18,32 @@
...
 
 /** A seekable and positionable FSInputStream that provides read access to a file. */
 class GoogleHadoopFSInputStream extends FSInputStream {
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  static final String READ_METHOD = "gcsFSRead";
+  static final String POSITIONAL_READ_METHOD = "gcsFSReadPositional";
+  static final String SEEK_METHOD = "gcsFSSeek";
+
+  private final Gson gson = new Gson();
```
arunkumarchacko, line 44: This can be static.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -98,6 +112,17 @@ public synchronized int read() throws IOException {
     return (b & 0xff);
   }
 
+  private void captureDuration(String suffix, Stopwatch stopwatch, int byteRead) {
```
arunkumarchacko, line 115: Please move private methods after Public methods for easier readability.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -98,6 +112,17 @@ public synchronized int read() throws IOException {
...
+    if (isTraceLoggingEnabled) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      // Capturing time in nanos because majority of the reads will be done from cache
+      long nanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
```
arunkumarchacko, line 119: What is the value in capturing this at the nano second granularity. It might be better to keep this consistent with other places where we are using milli seconds.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -174,8 +201,10 @@ public synchronized long getPos() throws IOException {
   @Override
   public synchronized void seek(long pos) throws IOException {
     logger.atFiner().log("seek(%d)", pos);
+    Stopwatch stopwatch = Stopwatch.createStarted();
     try {
       channel.position(pos);
+      captureDuration(SEEK_METHOD, stopwatch, 0);
```
arunkumarchacko, line 207: bytes_read do not make sense for seek, correct?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -142,8 +168,9 @@ public synchronized int read(byte[] buf, int offset, int length) throws IOExcept
   @Override
   public synchronized int read(long position, byte[] buf, int offset, int length)
       throws IOException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
     int result = super.read(position, buf, offset, length);
-
+    captureDuration(POSITIONAL_READ_METHOD, stopwatch, result);
```
arunkumarchacko, line 173: why not include offset, position and length information also in the logs?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -98,6 +112,17 @@ public synchronized int read() throws IOException {
...
   }
 
+  private void captureDuration(String suffix, Stopwatch stopwatch, int byteRead) {
+    if (isTraceLoggingEnabled) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      // Capturing time in nanos because majority of the reads will be done from cache
+      long nanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
+      jsonMap.put(suffix + "_latencyNs", nanos);
+      jsonMap.put(suffix + "_byteRead", byteRead);
+      logger.atInfo().log("%s", LazyArgs.lazy(() -> gson.toJson(jsonMap)));
```
arunkumarchacko, line 122: why not log the file path as well?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko, line 115: why not capture the duration taken by close() as well?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamIntegrationTest.java
```diff
@@ -79,14 +94,26 @@ public void read_singleBytes() throws Exception {
...
+      assertingHandler.verifyReadAPILogFields(GoogleHadoopFSInputStream.READ_METHOD);
+      assertingHandler.flush();
+
       assertThat(in.read(1, value, 1, 1)).isEqualTo(1);
+      // Total 4 calls for positional read
+      // 1. positional Read call
+      // 2. Seek to desired position
+      // 3. actual read from buffer after seek
+      // 4. Seek to old position
+      assertingHandler.assertLogCount(4);
```
arunkumarchacko, line 116: Do we want to validate the size read and the fact that logs for different APIs are logged?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -98,6 +112,17 @@ public synchronized int read() throws IOException {
     return (b & 0xff);
   }
 
+  private void captureDuration(String suffix, Stopwatch stopwatch, int byteRead) {
+    if (isTraceLoggingEnabled) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      // Capturing time in nanos because majority of the reads will be done from cache
+      long nanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
+      jsonMap.put(suffix + "_latencyNs", nanos);
+      jsonMap.put(suffix + "_byteRead", byteRead);
```
arunkumarchacko, line 121: please have "suffix" as a separate field and name it "method" so that it is easier to query.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -98,6 +112,17 @@ public synchronized int read() throws IOException {
     return (b & 0xff);
   }
 
+  private void captureDuration(String suffix, Stopwatch stopwatch, int byteRead) {
+    if (isTraceLoggingEnabled) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      // Capturing time in nanos because majority of the reads will be done from cache
+      long nanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
+      jsonMap.put(suffix + "_latencyNs", nanos);
```
arunkumarchacko, line 120: Please name this "elapsed" instead of latency.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamIntegrationTest.java
```diff
@@ -48,6 +54,14 @@ public static void afterClass() {
     gcsFsIHelper.afterAllTests();
   }
 
+  @Before
+  public void setup() {
+    grpcTracingLogger = Logger.getLogger(GoogleHadoopFSInputStream.class.getName());
+    grpcTracingLogger.setUseParentHandlers(false);
+    grpcTracingLogger.addHandler(assertingHandler);
```
arunkumarchacko, line 61: Please remove the handler as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -279,11 +285,26 @@ private GoogleCloudStorageItemInfo getObjectMetadata(StorageResourceId resourceI
...
   }
 
+  private void recordGCSAPILatency(
+      MeasureLong measure, long time, String status, String method, String protocol) {
+    if (storageOptions.isTraceLogEnabled()) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      jsonMap.put("method", method);
+      jsonMap.put("status", status);
+      jsonMap.put("protocol", protocol);
+      jsonMap.put("duration", time);
```
arunkumarchacko, line 295: The method has latency and this is named duration. Please keep this consistent with what we already have and name the elapsed?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
```diff
@@ -279,11 +285,26 @@ private GoogleCloudStorageItemInfo getObjectMetadata(StorageResourceId resourceI
...
+  }
+
   private void recordSuccessMetric(
       MeasureLong measure, Stopwatch stopwatch, String method, String protocol) {
     long time = stopwatch.elapsed(MILLISECONDS);
     TagKey[] keys = new TagKey[] {METHOD, STATUS, PROTOCOL};
     String[] values = new String[] {method, STATUS_OK, protocol};
+    recordGCSAPILatency(measure, time, STATUS_OK, method, protocol);
     metricsRecorder.recordLong(keys, values, measure, time);
     logger.atFinest().log(
```
arunkumarchacko, line 309: why not convert this into a JSON and log, rather than adding another log statement? Now the same information will be logged twice.

arunkumarchacko (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: > Capturing time in nanos because majority of the reads will be done from cache

As it is being read from in-memory cache it will not be captured as part of millisec.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: yes, that's why passed 0 to reuse the same private method.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: I don't think it will be useful as we are not interested in tracking each individual request but in accumulation.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: It was not of interest but can add it for `close` as well.

arunkumarchacko: This can be interesting in the write() case.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: I doubt we care much about the local cache access latencies. It would be better to keep this consistent with other places and keep this in MS.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: Position and offset is interesting, especially in the case of Parquet

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: It is better to add another method, rather than having a field which do not have a meaning in the case of seek()

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: We do care for it if calls are in magnitude 10^7. As far as other places i.e. calls to gcs, is concerned it's always in millisecond magnitude. So, we can bypass the nanosec precision but same argument can't be made for connector level APIs, specifically `read`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: Sure.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
singhravidutt: I will make it consistent it with latency.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageGrpcReadChannel.java
singhravidutt: Because we also want to control the trace logging with flag.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: I will stick to capturing the latency and read bytes as those are the one which we are interested in as of now. As far as capturing Request/Responses per API is concerned, we already have that in our `master` and hopefully we will be able to make use of it and deprecate this change (or whole 2.2.x branch) all together.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: making it consistent with `latency`

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: Added.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: This is not really latency.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: Are you talking about Guls changes? If yes, it has information only at the aggregate level.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -110,12 +129,18 @@ public synchronized int read() throws IOException {
    */
   @Override
   public synchronized int read(byte[] buf, int offset, int length) throws IOException {
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    Map<String, Object> apiTraces = new HashMap<>();
```
arunkumarchacko, line 133: It would be great if we can avoid creating this object in case tracing in not enabled. We are paying the price in all code paths.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -223,4 +264,25 @@ public int available() throws IOException {
...
+      apiTraces = new HashMap<>();
+    }
+    // Capturing time in nanos because majority of the reads will be done from cache
+    apiTraces.put(DURATION_NS, stopwatch.elapsed(TimeUnit.NANOSECONDS));
+  }
+
+  private void captureAPITraces(String method, Map<String, Object> apiTraces) {
+    if (isTraceLoggingEnabled) {
+      Map<String, Object> jsonMap = new HashMap<>();
+      jsonMap.put(String.format("%s_%s", method, GCS_PATH), gcsPath);
```
arunkumarchacko, line 279: It might be easier to query if we have method and path as separate fields

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -223,4 +264,25 @@ public int available() throws IOException {
...
+      if (apiTraces != null && !apiTraces.isEmpty()) {
+        for (Map.Entry item : apiTraces.entrySet()) {
+          jsonMap.put(String.format("%s_%s", method, item.getKey()), item.getValue());
```
arunkumarchacko, line 282: method is already logged. Do we really need to add method again?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
arunkumarchacko: I think it is ok to have another method to save on this. Note that tracing will be disabled on most of the cases.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -18,19 +18,41 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.fs.FSInputStream;
 import org.apache.hadoop.fs.FileSystem;
 
 /** A seekable and positionable FSInputStream that provides read access to a file. */
 class GoogleHadoopFSInputStream extends FSInputStream {
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  static final String READ_METHOD = "gcsFSRead";
```
arunkumarchacko, line 40: can this be private?

arunkumarchacko (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: marked all of them with @VisibleForTesting

PR #894, opened 2022-10-31
Add GCS Audit Log Headers

meharanjan318:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2342,6 +2345,11 @@ <RequestT extends StorageRequest<?>> RequestT initializeRequest(
       String token = downscopedAccessTokenFn.apply(accessBoundaries);
       request.getRequestHeaders().setAuthorization("Bearer " + token);
     }
+    if (storageOptions.isGcsAuditLogEnabled()
+        && !storageOptions.getHttpRequestHeaders().containsKey(CUSTOM_AUDIT_USER_HEADER)) {
+      String user = getUgiUserName();
+      if (!user.isEmpty()) request.getRequestHeaders().set(CUSTOM_AUDIT_USER_HEADER, user);
```
arunkumarchacko, line 2351: single line if statements can be avoided. It is error prone and harder to read.

arunkumarchacko, line 2351: user can be null if there is an exception.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2442,4 +2450,14 @@ public Map<String, Long> getStatistics() {
...
         .collect(toImmutableMap(Map.Entry::getKey, e -> e.getValue().get()));
   }
+
+  /** Helper method to get the UGI short user name */
+  private static String getUgiUserName() {
+    try {
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+      return ugi.getShortUserName();
+    } catch (IOException e) {
+      return null;
```
arunkumarchacko, line 2460: log. Why are we ignoring this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2342,6 +2345,11 @@ <RequestT extends StorageRequest<?>> RequestT initializeRequest(
       String token = downscopedAccessTokenFn.apply(accessBoundaries);
       request.getRequestHeaders().setAuthorization("Bearer " + token);
     }
+    if (storageOptions.isGcsAuditLogEnabled()
+        && !storageOptions.getHttpRequestHeaders().containsKey(CUSTOM_AUDIT_USER_HEADER)) {
```
arunkumarchacko, line 2349: why we need this check? Customer is supposed to set this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2342,6 +2345,11 @@ <RequestT extends StorageRequest<?>> RequestT initializeRequest(
       String token = downscopedAccessTokenFn.apply(accessBoundaries);
       request.getRequestHeaders().setAuthorization("Bearer " + token);
     }
+    if (storageOptions.isGcsAuditLogEnabled()
```
arunkumarchacko, line 2348: Please add integration tests for this

arunkumarchacko (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -436,6 +436,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_TRACE_LOG_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.tracelog.enable", false);
 
+  /** Configuration key to enable logging of custom audit details. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_AUDIT_LOG_ENABLE =
+      new HadoopConfigurationProperty<>("fs.gs.log-ugi-to-header.enable", false);
```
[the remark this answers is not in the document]
meharanjan318: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2442,4 +2450,14 @@ public Map<String, Long> getStatistics() {
     return statistics.entrySet().stream()
         .collect(toImmutableMap(Map.Entry::getKey, e -> e.getValue().get()));
   }
+
+  /** Helper method to get the UGI short user name */
+  private static String getUgiUserName() {
+    try {
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+      return ugi.getShortUserName();
```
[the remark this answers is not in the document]
meharanjan318: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: Logged error.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -264,6 +267,9 @@ public Boolean load(String bucketName) {
   // Function that generates downscoped access token.
   private final Function<List<AccessBoundary>, String> downscopedAccessTokenFn;
 
+  // Short username of current logged in user.
+  private String ugiUserName;
```
singhravidutt, line 271: nit: If we are extracting it in constructor can we make it `final`?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -567,4 +576,43 @@ private static List<String> getExpectedRequestsForCreateObject(
...
+          storage, resourceId, /* partitionSize= */ expectedSize, /* partitionsCount= */ 1);
+
+      // 3 types of log records are present in the job handler
+      // 1. Geting the object - GET call
+      // 2. Starting object upload (First write call) - POST call
+      // 3. Writing the chunk (Handled by the library internally therefore does not carry the storage options we set for HTTP request, thererfore we get the audit logs at the initial step and not one for each chunk written) - PUT call
+      jsonLogHander.getAllLogRecords().stream().filter(logRecord -> !logRecord.get("request_method").equals("PUT")).collect(
+          Collectors.toList()).forEach(logRecord -> {
+            LinkedTreeMap<String, Object> requestHeaders = (LinkedTreeMap<String, Object>) logRecord.get("request_headers");
+            assertEquals(ugiUser, requestHeaders.get("x-goog-custom-audit-user"));
```
singhravidutt, line 611: nit: Can we use this from gcsIml. CUSTOM_AUDIT_USER_HEADER ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
arunkumarchacko: why we need an additional flag (GcsAuditLogEnabled) if having this header is controlled by user

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2442,4 +2456,15 @@ public Map<String, Long> getStatistics() {
     return statistics.entrySet().stream()
         .collect(toImmutableMap(Map.Entry::getKey, e -> e.getValue().get()));
   }
+
+  private static String getUgiUserName() {
+    try {
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+      return ugi.getShortUserName();
+    } catch (IOException e) {
+      logger.atInfo().log(
```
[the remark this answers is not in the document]
arunkumarchacko: +1

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -347,6 +353,8 @@ public GoogleCloudStorageImpl(
               this.storageOptions, this.backgroundTasksThreadPool, credentials);
     }
     this.downscopedAccessTokenFn = downscopedAccessTokenFn;
+
+    this.ugiUserName = getUgiUserName();
```
arunkumarchacko, line 357: Do we need to call this if AuditLog is disabled?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -567,4 +576,43 @@ private static List<String> getExpectedRequestsForCreateObject(
...
+  @Test
+  public void initializeRequest_withGcsAuditLogEnabled() throws IOException {
+    AssertingLogHandler jsonLogHander = new AssertingLogHandler();
+    Logger jsonTracingLogger =
+        jsonLogHander.getLoggerForClass(EventLoggingHttpRequestInitializer.class.getName());
+
+    try {
+
+      String ugiUser = UUID.randomUUID().toString();
+      UserGroupInformation ugi = UserGroupInformation.createRemoteUser(ugiUser);
```
arunkumarchacko, line 589: where will this create the user?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: If GcsAuditLogEnabled is true we check whether the user has set the header or not. If they have we give them priority, if they have not we set it ourselves.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -436,6 +436,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_TRACE_LOG_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.tracelog.enable", false);
 
+  /** Configuration key to enable logging of custom audit details. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_AUDIT_LOGS_HEADER_ENABLE =
```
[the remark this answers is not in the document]
meharanjan318: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -567,4 +576,43 @@ private static List<String> getExpectedRequestsForCreateObject(
...
+                  .setTraceLogEnabled(true)
+                  .build(),
+              GoogleCloudStorageTestHelper.getCredentials());
+
+      int expectedSize = 5 * 1024 * 1024;
+      StorageResourceId resourceId = new StorageResourceId(TEST_BUCKET, name.getMethodName());
+      writeObject(
+          storage, resourceId, /* partitionSize= */ expectedSize, /* partitionsCount= */ 1);
+
+      // 3 types of log records are present in the job handler
```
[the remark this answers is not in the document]
meharanjan318: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
meharanjan318: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: Added check.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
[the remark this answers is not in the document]
arunkumarchacko: Also make changes to CHANGES.md

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -126,6 +126,7 @@ public class GoogleHadoopFileSystemConfigurationTest {
           put("fs.gs.storage.root.url", "https://storage.googleapis.com/");
           put("fs.gs.storage.service.path", "storage/v1/");
           put("fs.gs.tracelog.enable", false);
+          put("fs.gs.log-ugi-to-header.enable", false);
```
arunkumarchacko, line 129: Why not follow the pattern used for other configs? None of them has hiphen.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2442,4 +2456,15 @@ public Map<String, Long> getStatistics() {
...
   }
+
+  private static String getUgiUserName() {
+    try {
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+      return ugi.getShortUserName();
+    } catch (IOException e) {
+      logger.atInfo().log(
+          "Ignoring IOException; failure to fetch current user info");
+      return null;
```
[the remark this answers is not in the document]
meharanjan318: Adding this as a warning, because if we throw error it might break workflows which we do not want as this is an extra check that we have added for only our use case and were not checking before.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
meharanjan318: Sets the kerberos user for this process.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: This class has multiple constructors, if I make it final will have to send this as a parameter in all of them which will require changes in multiple places wherever the constructors are being called.

meharanjan318: Done

gcs/CONFIGURATION.md
```diff
@@ -131,6 +131,10 @@
     fs.gs.storage.http.headers.another-custom-header=another_custom_value
     ```
 
+*   `fs.gs.audit.logs.header.enable` (default: `false`)
```
medb, line 134: Is there are reason we need to add this functionality in the GCE connector itself? We intentionally added support for generic HTTP headers so we do not need to implement per-header logic in GCS connector.

medb (changes requested):

gcs/CONFIGURATION.md
meharanjan318: This is added for the case when user submits the job through open source API and does not set audit log headers but sets this property as enabled. In that case we pick the Kerberos/OS level user and set that in the header. We are not doing any changes to the HTTP headers that the user has set using fs.gs.storage.http.headers

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
meharanjan318: Done.

gcs/CONFIGURATION.md
medb: Do we need to cover this though? For example, there will be no audit headers if users submit job that does not use GCS connector, or users gsutil. I think it's maybe OK to have a limitation that job should be submitted via Dataproc API to have this feature, especially taking into account that this is a temporary solution, IIUC.

gcsio/pom.xml
```diff
@@ -203,6 +203,10 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-testing</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
```
arunkumarchacko, line 208: do we have to shade this JAR?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2442,4 +2455,23 @@ public Map<String, Long> getStatistics() {
...
+  }
+
+  private static String getCurrentLoggedInUser() {
+    try {
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
+      return ugi.getShortUserName();
+    } catch (IOException e) {
+      logger.atWarning().withCause(e).log(
+          "Ignoring IOException; failure to fetch current user info");
+      return null;
```
arunkumarchacko, line 2474: Let us re throw the exception if this is not expected to fail. That way we will not hide the bugs.

gcs/CONFIGURATION.md
arunkumarchacko: Thank you for the suggestion Igor. Meha, can we re-use the existing header for this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
meharanjan318: We do not want workflows to fail which were working fine before in case a functionality that we have added fails.

arunkumarchacko: Not all workflows will be affected correct. Only if they had the audit flag enabled, only then it will hit this code path, correct?

gcs/CONFIGURATION.md
medb: Btw, did we test if Hadoop's support for resolving env vars will work for us w/o any code changes?

I.e. if we will add this config to the Hadoop XML config (not sure about exact syntax, we may need to escape `$`):
```xml
<configuration>
  . . .
  <property>
    <name>fs.gs.storage.http.headers.X-Goog-Custom-Audit-User</name>
    <value>${user.name}</value>
  </property>
  . . .
</configuration>
```

If this works, then I think that we may not need even Dataproc Agent changes that sets this property manually for each user.

cc @cnauroth

cnauroth: As per Javadoc in the [`Configuration`](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java) class, configuration property expansion can pull from environment variables (prefixed as `${env.FOO}`), or other properties defined in the same file, or general JVM system properties. In the example above, using `${user.name}`, it would use the user name that launched the JVM process (e.g. Spark driver or YARN containers running Spark executors). In the context of Dataproc, the user running that JVM process varies depending on the cluster configuration and method of job submission:

| Cluster Configuration | Job Submission | User |
| - | - | - |
| Unsecured | API | yarn |
| Unsecured | ssh | yarn |
| Kerberos | API | dataproc |
| Kerberos | ssh | <user authenticated by `kinit`> |
| Personal Auth | API | <user authenticated by `kinit`> |
| Personal Auth | ssh | N/A, ssh blocked for Personal Auth clusters |

Does this satisfy the requirements? If so, then I expect you can make use of Hadoop configuration property expansion.

However, the earlier conversation discusses wanting to pick the "OS level user." If that is a requirement, then this solution would be insufficient, because:

1. In unsecured clusters, audit log headers would show the user as "yarn" instead of the submitting user.
2. In Kerberos clusters (but not using Personal Auth), for jobs submitted through the Dataproc API, audit log headers would show the user as "dataproc" instead of the submitting user.

If we need to satisfy that, then I can't think of any other solution other than the agent injecting a property.

PR #896, opened 2022-11-04
Use the Maven release profile to deploy using nexus-staging-maven-plugin

arunkumarchacko:
Skip the the deploy-maven-plugin. Use the release profile to deploy nexus-staging-maven-plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,8 @@
     <junit.version>4.13.2</junit.version>
     <mockito.version>3.11.2</mockito.version>
     <system-lambda.version>1.2.0</system-lambda.version>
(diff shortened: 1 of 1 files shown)


Deependra-Patel:
Title is unclear. 
`Use the release profile to deploy using nexus-staging-maven-plugin` Is this what we are achieving?

Also, we are skipping to do final release to keep that step manual right?

arunkumarchacko:
> 

We were using the default "deploy" plugin in Maven to deploy to Nexus. With this change we are using the plugin from Nexus.

Yes, this will release only to staging. Promoting to Maven will be manual.

Deependra-Patel (approved):

PR #902, opened 2022-11-11
dependencies for Veneer Client

singhravidutt:
Also upgraded `io.grpc` version from `1.43.1` to `1.50.1` because `com.google.cloud` is introducing [`io.grpc:grpc-googleapis`](https://mvnrepository.com/artifact/com.google.cloud/google-cloud-storage/2.14.0) dependency with `1.50.1` and does make sense to have all `io.grpc` to be consistent. Btw we already have `requireSameVersions` plugin against `io.grpc`.

### DIFF
--- a/gcsio/pom.xml
+++ b/gcsio/pom.xml
@@ -98,6 +98,10 @@
       <artifactId>grpc-protobuf</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>grpc-google-cloud-storage-v2</artifactId>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -274,7 +274,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-storage-bom</artifactId>
-        <version>2.14.0</version>
+        <version>${google.cloud-storage.bom.version}</version>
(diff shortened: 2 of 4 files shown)


arunkumarchacko (commented):
Please update CHANGES.md as well.

pom.xml
```diff
@@ -271,6 +271,13 @@
         <artifactId>google-http-client-gson</artifactId>
         <version>${google.http-client.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-storage-bom</artifactId>
+        <version>2.14.0</version>
```
medb, line 277: It should be `${google.cloud-storage.bom.version}` here.

gcsio/pom.xml
```diff
@@ -97,6 +97,10 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-protobuf</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-storage</artifactId>
+    </dependency>
```
medb, line 103: After adding this dependency you need to make sure that all jars that it brings are shaded.

medb (changes requested):

pom.xml
singhravidutt: Mybad. Updated.

gcsio/pom.xml
singhravidutt: Thanks. Added it in shading.

arunkumarchacko (approved):

medb (approved):

PR #905, opened 2022-11-15
added Implementation for google-cloud-storage client

singhravidutt:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -436,6 +436,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_TRACE_LOG_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.tracelog.enable", false);
 
+  /** Configuration key to enable using Veneer client for connecting with GCS */
+  public static final HadoopConfigurationProperty<Boolean> GCS_VENEER_CLIENT_ENABLE =
+      new HadoopConfigurationProperty<>("fs.gs.veneer.enable", false);
```
arunkumarchacko, line 441: I am not sure if the name Veneer is used outside. Consider changing this to something more meaningful

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSVeneerImpl.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Provides read/write access to Google Cloud Storage (GCS), using Java nio channel semantics. This
+ * is a basic implementation of the GoogleCloudStorage interface that mostly delegates through to
+ * the appropriate API call(s) google-cloud-storage client.
+ */
+public class GCSVeneerImpl implements GoogleCloudStorage {
```
arunkumarchacko, line 19: I was thinking if we should instead care a Builder and it will return the correct GoogleCloudStorage client. It will simplify the constructor and also can make this class to be not public

arunkumarchacko, line 19: Here also, can we not use name Veneer?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSVeneerImpl.java
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Having an instance of gscImpl to redirect calls to Json client while new client implementation
+   * is in WIP.
+   */
+  private GoogleCloudStorageImpl oldGCSClient;
```
arunkumarchacko, line 24: rename this to gcsClientDelegate?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GCSVeneerImplTest.java
```diff
@@ -0,0 +1,156 @@
...
+    // Set up the mock Insert to throw an exception when execute() is called.
+    Error fakeError = new Error("Fake error");
+    MockHttpTransport transport =
+        mockTransport(
+            jsonErrorResponse(ErrorResponses.NOT_FOUND),
+            inputStreamResponse(
+                CONTENT_LENGTH,
+                /* headerValue = */ 1,
+                new ThrowingInputStream(/* readException = */ null, fakeError)));
+    GoogleCloudStorageImpl gcs = mockedGcs(transport);
```
arunkumarchacko, line 58: why is this not Veneer?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GCSVeneerIntegrationTest.java
```diff
@@ -0,0 +1,549 @@
...
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests that require a particular configuration of GCSVeneerImpl. */
+@RunWith(JUnit4.class)
+public class GCSVeneerIntegrationTest {
```
arunkumarchacko, line 52: why can't we reuse the existing Integration tests?

arunkumarchacko (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
singhravidutt: Yeah discussed it with Ben too. Will use `ManualClient` instead of `Veneer`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSVeneerImpl.java
singhravidutt: sure.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GCSVeneerIntegrationTest.java
singhravidutt: Not the one under `gcsImplTest` maybe we can reuse integration test (if there are) for `GoogleCloudStorageFileSystemImpl`. Reason being GCSImpl object is getting initialised as part of individual test.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GCSVeneerImplTest.java
singhravidutt: Yeah, found many discrepancies here will fix it and also try to increase the coverage.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSVeneerImpl.java
singhravidutt: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Provides read/write access to Google Cloud Storage (GCS), using Java nio channel semantics. This
+ * is a basic implementation of the GoogleCloudStorage interface that mostly delegates through to
+ * the appropriate API call(s) google-cloud-storage client.
+ */
+public class GCSManualClientImpl implements GoogleCloudStorage {
```
arunkumarchacko, line 22: Can this be not public?

arunkumarchacko, line 22: "manual" client do not sound right. Should we call this GoogleCloudStorageImplV2?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Having an instance of gscImpl to redirect calls to Json client while new client implementation
+   * is in WIP.
+   */
+  private GoogleCloudStorageImpl gcsClientDelegate;
+
+  private GoogleCloudStorageOptions storageOptions;
+  private Credentials credentials;
+
+  public GCSManualClientImpl(GCSManualClientImplBuilder builder) throws IOException {
```
arunkumarchacko, line 33: once we have a builder, then we can make the constructor public

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+    this.storageOptions = checkNotNull(builder.storageOptions, "options must not be null");
```
arunkumarchacko, line 34: these checks can go to Builder.build() method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @VisibleForTesting
+    public GCSManualClientImplBuilder withHttpRequestInitializer(
+        HttpRequestInitializer httpRequestInitializer) {
+      this.httpRequestInitializer = httpRequestInitializer;
+      return this;
+    }
+
+    public GCSManualClientImpl build() throws IOException {
+      return new GCSManualClientImpl(this);
```
arunkumarchacko, line 219: Usually the builder will create the instance, rather than pasisng the builder to the constructor

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+  public Map<String, Long> getStatistics() {
+    return gcsClientDelegate.getStatistics();
+  }
+
+  @Override
+  public void close() {
+    gcsClientDelegate.close();
+  }
+
+  public static class GCSManualClientImplBuilder {
```
arunkumarchacko, line 187: It might be a good idea to move this to GoogleCloudStorage class so that we can hide this Impl class details.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+        Function<List<AccessBoundary>, String> downscopedAccessTokenFn) {
+      this.storageOptions = storageOptions;
+      this.credentials = credentials;
+      this.downscopedAccessTokenFn = downscopedAccessTokenFn;
+    }
+
+    @VisibleForTesting
+    public GCSManualClientImplBuilder withApairyClientStorage(
+        com.google.api.services.storage.Storage storage) {
+      this.storage = storage;
```
arunkumarchacko, line 207: can validate this to be non-null. For others as well

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -436,6 +436,10 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_TRACE_LOG_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.tracelog.enable", false);
 
+  /** Configuration key to enable using google-cloud-storage client for connecting with GCS */
+  public static final HadoopConfigurationProperty<Boolean> GCS_MANUAL_CLIENT_ENABLE =
+      new HadoopConfigurationProperty<>("fs.gs.manualclient.enable", false);
```
arunkumarchacko, line 441: rename this to fs.gs.gcs.javaclient.enable? "manual" do not sound right.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: yeah, there is no harm doing it here.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: yeah, my bad It should be private now.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: I am delegating this check to the constructor as it has all info to figure out for what all fields null check is required.
Although I agree in this case it wouldn't matter much but I find this null checks better to be placed in constructor.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: I don't prefer "V2" as it's vague, I can go with 'java client' as it's manual java client and that way it will be consistent with our flag name too.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
singhravidutt: sure.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: As discussed offline. This Nested class setup seems fine.
Only caveat being the `Impl` class need to public for integrations test.
There are two options via which we can make this `package-protected`
1. extract out the builder from the `Impl` class 
    * Have to make `Impl` public.
2. Created another wrapper Builder
    * Will have somany classees tied to each other but will serve the purpose.
    
For now I am annotating this class with `@VisibleForTesting`. Will delegate this decision to some later point in time.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
```diff
@@ -0,0 +1,223 @@
...
+   * is in WIP.
+   */
+  private GoogleCloudStorageImpl gcsClientDelegate;
+
+  private GoogleCloudStorageOptions storageOptions;
+  private Credentials credentials;
+
+  private GCSJavaClientImpl(GCSJavaClientImplBuilder builder) throws IOException {
+    this.storageOptions = checkNotNull(builder.storageOptions, "options must not be null");
+    this.credentials = checkNotNull(builder.credentials, "credentials must not be null");
```
arunkumarchacko, line 36: if httpRequestInitializer is not null, this can be null?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
```diff
@@ -0,0 +1,223 @@
...
+
+    if (builder.httpRequestInitializer != null) {
```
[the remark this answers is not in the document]
singhravidutt: Nope. Overriding both `storage` and  `httpRequestInitializer` are part of testing use cases. I made sure that there ill not be a case where both will be provided. 

IMO, We already have code specific for testing, don't wanna extend it. If it results in some unwanted behaviour it will be for testing and should be handled accordingly.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
```diff
@@ -0,0 +1,223 @@
...
+    return gcsClientDelegate.getStatistics();
+  }
+
+  @Override
+  public void close() {
+    gcsClientDelegate.close();
+  }
+
+  public static class GCSJavaClientImplBuilder {
+
```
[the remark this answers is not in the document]
singhravidutt: done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
singhravidutt: Didn't understand it fully.
`httpRequestInitializer` will and should be null for all production usecases, We exposed a constructor in GCSImpl for test use cases only.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
```diff
@@ -0,0 +1,223 @@
...
+  private Credentials credentials;
+
+  private GCSJavaClientImpl(GCSJavaClientImplBuilder builder) throws IOException {
+    this.storageOptions = checkNotNull(builder.storageOptions, "options must not be null");
+    this.credentials = checkNotNull(builder.credentials, "credentials must not be null");
+
+    if (builder.httpRequestInitializer != null) {
+      logger.atWarning().log(
+          "Overriding httpRequestInitializer. ALERT: Should not be hit in production");
+      gcsClientDelegate =
```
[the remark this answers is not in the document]
singhravidutt: I believe it's not required here as there is no conflict but will add it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * Provides read/write access to Google Cloud Storage (GCS), using Java nio channel semantics. This
+ * is a basic implementation of the GoogleCloudStorage interface that mostly delegates through to
+ * the appropriate API call(s) google-cloud-storage client.
+ */
+@VisibleForTesting
+public class GCSJavaClientImpl implements GoogleCloudStorage {
```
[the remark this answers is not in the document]
singhravidutt: It's will unnecessary make class name super huge.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSManualClientImpl.java
singhravidutt: response below should cover this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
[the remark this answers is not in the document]
arunkumarchacko: I will leave it up to you. It should be GcsJavaClientImpl

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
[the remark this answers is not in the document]
arunkumarchacko: It is good to have the preconditions checked in the constructor.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
arunkumarchacko: if builder. httpRequestInitializer is not null, the credentials are not used. So I was wondering why one need to provide credentials if httpRequestInitializer is provided.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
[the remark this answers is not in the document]
singhravidutt: ok.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GCSJavaClientImpl.java
singhravidutt: `Short ans`: because we don't want to have unnecessary checks in code for just testing usecases.
`Long ans`: If we are overriding `httpRequestInitializer` we must have used credentials to create it (there is no workaround to that). Now, we can handle this requirement in two ways
1. Create constructor specifically for this testing usecase. That is what is done in GCSImpl.
2. Second is, don't introduce extra code in production and instead try to align testing as per production code. That's what I am trying to achieve here by imposing testing code to pass in credentials too which they any how would have if they are overriding `httpRequestInitializer`.

arunkumarchacko (approved):

PR #908, opened 2022-11-18
dependencies for Java Client (#902)

singhravidutt:


gcs/CHANGES.md
```diff
@@ -82,6 +82,8 @@
     fs.gs.grpc.read.zerocopy.enable (default : true)
     ```
 
+1.  Added dependency on new storage client library: [google-cloud-storage](https://github.com/googleapis/java-storage/tree/main/google-cloud-storage).
```
arunkumarchacko, line 85: why are we adding this here? This should be added after the "next" tag on the top

PR #911, opened 2022-11-23
Fix the issue where connector init is failing if both tracing and met…

arunkumarchacko:
…rics are enabled

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -387,13 +387,7 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
 
     this.httpRequestInitializer = this.storage.getRequestFactory().getInitializer();
 
-    this.metricsRecorder =
-        MetricsSink.CLOUD_MONITORING == this.storageOptions.getMetricsSink()
-            ? CloudMonitoringMetricsRecorder.create(
-                options.getProjectId(),
-                new CredentialAdapter(
-                    ((RetryHttpInitializer) httpRequestInitializer).getCredential()))
-            : new NoOpMetricsRecorder();
+    this.metricsRecorder = getMetricsRecorder(options, credential, this.httpRequestInitializer);
 
     // Create the gRPC stub if necessary;
     if (this.storageOptions.isGrpcEnabled()) {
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -424,6 +418,24 @@ private GoogleCloudStorageImpl(
     this.downscopedAccessTokenFn = downscopedAccessTokenFn;
   }
 
+  private static MetricsRecorder getMetricsRecorder(
```
singhravidutt, line 421: IIUC, What we are trying to do here is to enforce the use the `credential` which was passed in to constructor instead of extracting it from `RetryHttpInitializer`.
In production code flow we ALWAYS pass in the `credentials` which is used in creating the HTTP transport for `Storage`. So, `credential` is/can never `null` in production code flow. I am up for adding a null check for that.

Now, coming to testing, we agree that we need multiple GCSImpl constructors so that we can mock or override the storage OR transport layer. In any case we must have used some credentials (either a fake or real), why can't we update our test case to make sure to use the constructor where `credential` is getting passed instead of just passing in the overriden transport layer? I do agree that it will require a bit more changes in our tests but IMO I will still prefer that instead of having hacks in production code where God knows which credential is getting used for what.
Another point in favour of not adding this hack would be that GCSImpl is going to be here long enough (even after integrating new java-client) if possible let's try not to add hack in here yet.

On high level I believe extracting `credential` from http transport is wrong (And maybe extracting Http transport form storage too). It will never fall under production code flow and for testing it's `test` code responsibility to pass in the correct credential.
Will let @medb @cnauroth Comment on this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -387,13 +387,7 @@ private GoogleCloudStorageImpl(
...
     this.httpRequestInitializer = this.storage.getRequestFactory().getInitializer();
 
-    this.metricsRecorder =
-        MetricsSink.CLOUD_MONITORING == this.storageOptions.getMetricsSink()
-            ? CloudMonitoringMetricsRecorder.create(
-                options.getProjectId(),
-                new CredentialAdapter(
-                    ((RetryHttpInitializer) httpRequestInitializer).getCredential()))
-            : new NoOpMetricsRecorder();
+    this.metricsRecorder = getMetricsRecorder(options, credential, this.httpRequestInitializer);
```
singhravidutt, line 390: In `getMetricRecorder` we are giving priority to `credential` over the credential stored in `httpRequestInitializer` (extracted from Storage object).
This is a behaviour change and might be breaking change for customer. How we are validating that this is a correct thing to do?

One can see a similar pattern for When gRPC is enabled where passed(passed in constructor call) in credentials are used.I can see the cain of priority as `credentials >credential > httpRequestInitializer`.

Why not extract the logic of which credentials gets the priority and and just use that credential in rest of the function instead of deciding the priority at two different places?

I do understand that in `getMetricsRecorder` too we are maintaining the priority as `credential> httpRequestInitializer` but having this same logic segregated at various place will create issue in maintaining consistency.

I will let you take the call.

singhravidutt (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
arunkumarchacko: The priority is the same as the existing code. 

We have refactored this piece of code in the master branch. If we have to refactor this, it will be a bigger change, which is not related to this bug. It can be taken up as a separate change if you feel strongly about it.

PR #923, opened 2022-12-12
Release 2.2.10

arunkumarchacko:
Release 2.2.10


Deependra-Patel (approved):
Arun will create a new PR to fix regex issues and also fix automation that this doesn't happen in next release.

PR #929, opened 2022-12-21
java storage write channel

singhravidutt:


gcsio/pom.xml
```diff
@@ -180,6 +180,10 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+    </dependency>
```
medb, line 186: We do not use `easymock` in Dataproc - this is google3 policy that we adhere in our OSS projects too.

medb (changes requested):

gcsio/pom.xml
singhravidutt: Removed. Got added as part of some experiment.

PR #932, opened 2022-12-26
dependencies for Veneer Client (#902)

singhravidutt:


pom.xml
```diff
@@ -93,6 +93,7 @@
     <google.auto-value.version>1.8.2</google.auto-value.version>
     <google.cloud-bigquerystorage.version>2.25.0</google.cloud-bigquerystorage.version>
     <google.cloud-core.version>2.5.4</google.cloud-core.version>
+    <google.cloud-storage.bom.version>2.15.0</google.cloud-storage.bom.version>
     <google.cloud-storage.grpc.version>2.2.2-alpha</google.cloud-storage.grpc.version>
```
medb, line 97: Is this property used anywhere after this change?

medb (changes requested):

pom.xml
singhravidutt: It is getting used to pick the version for `google-cloud-storage-bom` artifact.
We are integrating new java-client created by gcs team.

medb: I meant `google.cloud-storage.grpc.version` property. Seems we removed dependency that used it directly, so we can remove this property too?

singhravidutt: aah. Yeah It can be removed.

pom.xml
```diff
@@ -468,11 +475,6 @@
         <version>${google.flogger.version}</version>
         <scope>runtime</scope>
       </dependency>
-      <dependency>
```
arunkumarchacko, line 471: why we need to remove this? This is getting added indirectly from GCS Client library?

pom.xml
```diff
@@ -468,11 +475,6 @@
...
-        <groupId>com.google.api.grpc</groupId>
-        <artifactId>grpc-google-cloud-storage-v2</artifactId>
```
arunkumarchacko, line 473: this is present in gcs/pom.xml. Is it still safe to remove?

singhravidutt: Yes. As it is getting added as part of `google-cloud-storage`.

pom.xml
singhravidutt: > This is getting added indirectly from GCS Client library?

Yes.

arunkumarchacko (approved):

medb (approved):

PR #937, opened 2022-12-29
Misc cleanup

medb:


cjac (approved):
LGTM

cjac:
It would be nice if you could bump the coverage up to 90%

PR #941, opened 2023-01-05
Prefetch footer when creating/opening `GoogleCloudStorageReadChannel`

medb:
This allows to eliminate metadata request when opening object for reading at cost of always pre-fetching footer.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -170,8 +170,8 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
 
     // Initialize metadata if available.
     GoogleCloudStorageItemInfo info = getInitialMetadata();
-    if (info != null || readOptions.isFastFailOnNotFoundEnabled()) {
-      initMetadata(info == null ? fetchInitialMetadata() : info);
+    if (info != null) {
+      initMetadata(info);
     } else if (readOptions.isFastFailOnNotFoundEnabled()) {
       prefetchFooterAndInitMetadata();
     }
@@ -207,7 +207,8 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
 
   /**
    * Returns {@link GoogleCloudStorageItemInfo} used to initialize metadata in constructor or {@code
-   * null} if {@link GoogleCloudStorageReadOptions#isFastFailOnNotFound()} is set to {@code false}.
+   * null} if {@link GoogleCloudStorageReadOptions#isFastFailOnNotFoundEnabled()} is set to {@code
+   * false}.
    */
   @Nullable
   protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
@@ -241,6 +242,7 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
     checkState(metadataInitialized, "metadata should be initialized already for '%s'", resourceId);
 
     // Do not cache footer for empty and gzip-encoded files
+    // TODO: cache footer if whole gzipped object were read.
     if (size == 0 || gzipEncoded) {
       return;
     }
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -35,8 +35,6 @@ import com.google.api.services.storage.model.StorageObject;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions.Fadvise;
 import com.google.cloud.hadoop.util.ApiErrorExtractor;
 import com.google.cloud.hadoop.util.ClientRequestHelper;
-import com.google.cloud.hadoop.util.ResilientOperation;
-import com.google.cloud.hadoop.util.RetryDeterminer;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.google.common.flogger.GoogleLogger;
@@ -174,6 +172,8 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
     GoogleCloudStorageItemInfo info = getInitialMetadata();
     if (info != null || readOptions.isFastFailOnNotFoundEnabled()) {
       initMetadata(info == null ? fetchInitialMetadata() : info);
+    } else if (readOptions.isFastFailOnNotFoundEnabled()) {
+      prefetchFooterAndInitMetadata();
     }
   }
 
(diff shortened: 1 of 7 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -215,39 +215,39 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
...
-  /** Returns {@link GoogleCloudStorageItemInfo} used to initialize metadata in constructor. */
-  private GoogleCloudStorageItemInfo fetchInitialMetadata() throws IOException {
-    StorageObject object;
+  /** Method used to cache footer and initializes metadata in constructor. */
+  private void prefetchFooterAndInitMetadata() throws IOException {
+    checkState(!metadataInitialized, "metadata should not be initialized yet for '%s'", resourceId);
+
+    // TODO: use metadata request if `readOptions.getMinRangeRequestSize()` is 0.
+    Storage.Objects.Get getObject =
+        createDataRequest(/* rangeHeader= */ "bytes=-" + readOptions.getMinRangeRequestSize());
```
cnauroth, line 224: Previously, the metadata request used `setFields` to select only the required metadata. Is it possible to preserve that behavior on this data request?

medb: Fields filtering applicable only for metadata requests that return object metadata in JSON format in HTTP response body. It's not applicable to data request used for footer pre-fetching as it returns raw bytes in HTTP body and we take necessary object metadata from HTTP headers instead that already very limited and can not be filtered.

medb:
> 1. We will be incurring this cost for no-parquet files as well, correct?
> 2. Do we have some benchmark numbers for how much improvement this is bringing in?
> 3. Given that we are having footer caching, how much benefit are we getting?

1. In terms of $ cost - data and metadata requests cost the same as both are GET requests. There is a little higher cost for 2 MiB data prefetch in terms of latency +20-50ms, but taking into account that in majority of case it saves 1 request with ~100ms@p99 latency it should not be considered a significant drawback for rest of the cases where this optimization does not apply.
2. It saves 1 HTTP GET request (~100ms@p99) when reading structured data formats (Parquet, ORC, etc), it may not have big impact on e2e perf, but it allows to significantly reduce number of overall HTTP request to GCS.
3. Spark footer caching done on top of this optimization and benefits from it too, i.e. right now Spark needs 2 requests (1 metadata and 1 data) to cache footer, with this change it will require just one data request.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
     }
   }
 
+  private void cacheFooterWithRetries(Storage.Objects.Get getObject, HttpResponse response)
```
singhravidutt, line 1063: Wouldn't it be better to put retires inside `cacheFooter` while reading footer content from response's `contentStream`  instead of retrying the process?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
...
+        }
+        try {
+          response = getObject.executeMedia();
+          // TODO(b/110832992): validate response range header against expected/request range.
+        } catch (IOException e) {
+          response = handleExecuteMediaException(e);
+        }
+      }
+    }
+    checkState(
```
singhravidutt, line 1092: Even if perfecting of footer failed is it necessary to fail the `open` request? We already have got metadata. If there is a read request to fetch footer content it can and will be retried at that time.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
...
+        break;
+      } catch (IOException footerException) {
+        logger.atInfo().withCause(footerException).log(
+            "Failed to prefetch footer (retry #%s/%s) for '%s'",
+            retriesCount + 1, maxRetries, resourceId);
+        if (retriesCount == 0) {
+          readBackOff.get().reset();
+        }
+        if (retriesCount == maxRetries) {
+          resetContentChannel();
```
singhravidutt, line 1081: Why we need to reset the Channel? Channel is not even created till this point. I do understand there is no harm in resetting it but it's adding a confusion about channel existence.

IIUC, contentChannel would only exist if there is already a `openStream`.

cnauroth (commented):
Hello @medb . I haven't checked closely yet, but wanted to drop a quick note that I'm wondering if this change would conflict with #724. I believe that one was important for GCS Ranger Plugin support. CC: @aalexx-S , @cyxxy .

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -215,39 +215,39 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
...
   }
 
-  /** Returns {@link GoogleCloudStorageItemInfo} used to initialize metadata in constructor. */
-  private GoogleCloudStorageItemInfo fetchInitialMetadata() throws IOException {
-    StorageObject object;
+  /** Method used to cache footer and initializes metadata in constructor. */
+  private void prefetchFooterAndInitMetadata() throws IOException {
+    checkState(!metadataInitialized, "metadata should not be initialized yet for '%s'", resourceId);
+
+    // TODO: use metadata request if `readOptions.getMinRangeRequestSize()` is 0.
```
arunkumarchacko, line 222: what is the point making this call when MinRangeRequestSize is 0. i.e. what will happen with this change if MinRangeRequestSize is 0?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -215,39 +215,39 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
...
+      response = getObject.executeMedia();
+      // TODO(b/110832992): validate response range header against expected/request range
     } catch (IOException e) {
-      throw errorExtractor.itemNotFound(e)
-          ? createFileNotFoundException(resourceId, e)
-          : new IOException("Error reading " + resourceId, e);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-      throw new IOException("Thread interrupt received.", e);
+      if (errorExtractor.rangeNotSatisfiable(e)) {
```
arunkumarchacko, line 230: will this fail if the actual size is less than MinRangeRequestSize?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
...
+      throws IOException {
+    for (int retriesCount = 0; retriesCount < maxRetries; retriesCount++) {
+      try {
+        cacheFooter(response);
+        if (retriesCount != 0) {
+          logger.atInfo().log(
+              "Successfully cached footer after %s retries for '%s'", retriesCount, resourceId);
+        }
+        break;
+      } catch (IOException footerException) {
```
arunkumarchacko, line 1073: do we want to retry on all exceptions?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
     }
   }
 
+  private void cacheFooterWithRetries(Storage.Objects.Get getObject, HttpResponse response)
+      throws IOException {
+    for (int retriesCount = 0; retriesCount < maxRetries; retriesCount++) {
```
arunkumarchacko, line 1065: Usually we retry with exponential backoff. Since this is nice to have, can we make this simpler by not retrying?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -215,39 +215,39 @@ protected GoogleCloudStorageItemInfo getInitialMetadata() throws IOException {
...
-      Storage.Objects.Get getObject =
-          createMetadataRequest().setFields("contentEncoding,generation,size");
-      object =
-          ResilientOperation.retry(
-              getObject::execute,
-              readBackOff.get(),
-              RetryDeterminer.SOCKET_ERRORS,
-              IOException.class,
-              sleeper);
+      response = getObject.executeMedia();
```
arunkumarchacko, line 227: there is code duplication here. Do we want to refactor and reuse?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1060,6 +1060,42 @@ protected InputStream openStream(long bytesToRead) throws IOException {
...
+        }
+        if (retriesCount == maxRetries) {
+          resetContentChannel();
+          throw footerException;
+        }
+        try {
+          response = getObject.executeMedia();
+          // TODO(b/110832992): validate response range header against expected/request range.
+        } catch (IOException e) {
+          response = handleExecuteMediaException(e);
```
arunkumarchacko, line 1088: This method never return response. A bit confusing.

aalexx-S:
@medb Can we use the get metadata request type in **com.google.cloud.hadoop.gcsio.StorageRequestFactory** instead of the raw object get request? I'm afraid this may break the ranger plugin. @cyxxy fyi

PR #947, opened 2023-01-23
Fix a minor issue with README when we do the release

arunkumarchacko:
Removing the release tag for BQ connector. Will add it back when we automate the release for BQ connector

### DIFF
--- a/README-template.md
+++ b/README-template.md
@@ -78,7 +78,7 @@ To add a dependency on one of the connectors using Maven, use the following:
     <dependency>
       <groupId>com.google.cloud.bigdataoss</groupId>
       <artifactId>gcs-connector</artifactId>
-      <version>${next-gcs-connector-release-tag}</version>
+      <version>hadoop2-${next-gcs-connector-release-tag}</version>
     </dependency>
     ```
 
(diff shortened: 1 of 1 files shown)


README-template.md
```diff
@@ -88,7 +88,7 @@ To add a dependency on one of the connectors using Maven, use the following:
     <dependency>
       <groupId>com.google.cloud.bigdataoss</groupId>
       <artifactId>bigquery-connector</artifactId>
-      <version>${next-bigquery-connector-release-tag}</version>
+      <version>hadoop2-1.2.0</version>
```
Deependra-Patel, line 91: Why are we hardcoding the value here?

README-template.md
```diff
@@ -78,7 +78,7 @@ To add a dependency on one of the connectors using Maven, use the following:
     <dependency>
       <groupId>com.google.cloud.bigdataoss</groupId>
       <artifactId>gcs-connector</artifactId>
-      <version>${next-gcs-connector-release-tag}</version>
+      <version>hadoop2-${next-gcs-connector-release-tag}</version>
```
Deependra-Patel, line 81: We also publish for hadoop3 right?

arunkumarchacko: Yes, we publish for both hadoop2 and hadoop3 from this branch.

README-template.md
arunkumarchacko: This is the version for BQ connector. we have not released this in a while and probably will not. We have removed this from master for e.g. We can add this back if we plan to add automation for Hive BQ connector release.

Deependra-Patel (approved):

PR #964, opened 2023-02-16
Run integration tests against new Storage Client

singhravidutt:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
```diff
@@ -81,6 +82,13 @@ public class GoogleCloudStorageFileSystemIntegrationTest {
   // Name of the test object.
   protected String objectName = "gcsio-test.txt";
 
+  @Parameterized.Parameter public boolean testStorageClientImpl;
+
+  @Parameters
+  public static Iterable<Boolean> getTesStorageClientImplParameter() {
+    return List.of(false, true);
```
arunkumarchacko, line 89: Instead of true, false, why not provide the actual values (i.e. the Impl class) as parameter

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemTest.java
```diff
@@ -126,7 +153,11 @@ public void testConstructor() throws IOException {
         options.getCloudStorageOptions().toBuilder().setAppName("appName").build());
 
     // Verify that credentials == null works - this is required for unauthenticated access.
-    new GoogleCloudStorageFileSystemImpl((Credentials) null, optionsBuilder.build());
+    // TODO: this seems to be not the case with java-storage.
+    // Confirm and remove this `if`.
+    if (!testStorageClientImpl) {
+      new GoogleCloudStorageFileSystemImpl((Credentials) null, optionsBuilder.build());
```
arunkumarchacko, line 159: What is the purpose of this code?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageIntegrationHelper.java
```diff
@@ -49,7 +49,7 @@
 public abstract class GoogleCloudStorageIntegrationHelper {
 
   // Prefix used for naming test buckets.
-  public static final String TEST_BUCKET_NAME_PREFIX = "gcsio-test";
+  public static final String TEST_BUCKET_NAME_PREFIX = "dataproc-gcs-gcsio";
```
arunkumarchacko, line 52: why changing this?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNewIntegrationTest.java
```diff
@@ -102,18 +109,14 @@ public static void beforeClass() throws Throwable {
...
     gcsfsIHelper = new GoogleCloudStorageFileSystemIntegrationHelper(gcsfs);
     gcsfsIHelper.beforeAllTests();
+    gcsRequestsTracker = new TrackingHttpRequestInitializer(httpRequestsInitializer);
   }
 
-  @AfterClass
-  public static void afterClass() {
+  @After
+  public void after() {
     gcsfsIHelper.afterAllTests();
```
arunkumarchacko, line 117: why changing this?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNewIntegrationTest.java
```diff
@@ -692,7 +695,9 @@ public void listObjectInfo_includePrefix_objectInBucket() throws Exception {
   public void listObjectInfo_includePrefix_implicitDirInBucket() throws Exception {
     GoogleCloudStorage gcs = createGoogleCloudStorage(gcsOptions);
 
-    String testBucket = gcsfsIHelper.createUniqueBucket("lst-objs_incl-pfx_impl-dir-in-bckt");
+    String testBucket =
```
arunkumarchacko, line 698: why changing this?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNewIntegrationTest.java
```diff
@@ -1337,7 +1345,11 @@ private GoogleCloudStorage createGoogleCloudStorage(GoogleCloudStorageOptions op
   }
 
   private String getTestResource() {
-    return name.getMethodName() + "_" + UUID.randomUUID();
+    return getMethodName() + "_" + UUID.randomUUID();
+  }
+
+  private String getMethodName() {
+    return name.getMethodName().replace('[', '-').replace(']', '-');
```
arunkumarchacko, line 1352: why is this required? Is there any behaviour change in Veneer?

util/src/main/java/com/google/cloud/hadoop/util/ErrorType.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.util;
+
+public enum ErrorType {
```
arunkumarchacko, line 19: Can we avoid having this as public?

util/src/main/java/com/google/cloud/hadoop/util/ErrorTypeExtractor.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.util;
+
+public interface ErrorTypeExtractor {
```
arunkumarchacko, line 19: why we need this as public Interface?

util/src/main/java/com/google/cloud/hadoop/util/GrpcErrorTypeExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.util;
+
+import io.grpc.Status;
+
+public class GrpcErrorTypeExtractor implements ErrorTypeExtractor {
```
arunkumarchacko, line 21: why public?

arunkumarchacko (changes requested):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemXAttrsIntegrationTest.java
```diff
@@ -59,13 +61,20 @@ public final class GoogleHadoopFileSystemXAttrsIntegrationTest {
 
   private String bucketName;
 
+  @Parameterized.Parameter public boolean testStorageClientImpl;
```
medb, line 64: Taking into account that we would want to test all our clients, and in future there will be more than 2 of the, should we use exiting `ClientType` enum for parametrizing tests, as boolean looks a little weird and harder maintain going forward?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -541,6 +545,19 @@ private static void validate(GoogleCloudStorageItemInfo itemInfo) throws IOExcep
     }
   }
 
+  private IOException convertError(Exception error) {
```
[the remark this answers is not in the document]
singhravidutt: no we can't it sues state variable.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -80,6 +84,7 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
             .setDownscopedAccessTokenFn(downscopedAccessTokenFn)
             .build());
     this.storageOptions = options;
+    this.errorExtractor = GrpcErrorTypeExtractor.INSTANCE;
```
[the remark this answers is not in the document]
singhravidutt: Because we are supporting only gRPC via new client.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemNewIntegrationTest.java
```diff
@@ -44,29 +44,36 @@
...
   @Rule public TestName name = new TestName();
 
   private TrackingHttpRequestInitializer gcsRequestsTracker;
 
-  @BeforeClass
-  public static void beforeClass() throws Throwable {
+  @Parameterized.Parameter public boolean testStorageClientImpl;
+
+  @Parameters
+  public static Iterable<Boolean> getTesStorageClientImplParameter() {
```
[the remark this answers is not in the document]
singhravidutt: yes, https://junit.org/junit4/javadoc/4.12/org/junit/runners/Parameterized.html

util/src/main/java/com/google/cloud/hadoop/util/GrpcErrorTypeExtractor.java
singhravidutt: This is getting defined in `util` package and used in `gcsio`. It has to be public.

util/src/main/java/com/google/cloud/hadoop/util/ErrorTypeExtractor.java
singhravidutt: This is getting defined in `util` package and used in `gcsio`. It has to be public.

util/src/main/java/com/google/cloud/hadoop/util/ErrorType.java
singhravidutt: This is getting defined in `util` package and used in `gcsio`. It has to be public.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNewIntegrationTest.java
singhravidutt: With parameterization each `Test` method will gets converted, for eg if test method name was `testMe` with parameterization it will get converted into `testMe[0]`, `testMe[1]` etc depending upon number of variations are there.

As we were using `methodName` as a objectPath for gcs. It was creating some issues in request validation due to url-encoding. To bypass this I have added additional filtering/replacements.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageIntegrationHelper.java
singhravidutt: Because now we are running integration test against gRPC too.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageNewIntegrationTest.java
singhravidutt: With bucket prefix changing from `gcsio-test` to `dataproc-gcs-gcsio` this was causing overall bucket name to go beyond limits.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemTest.java
singhravidutt: `null` credentials were not allowed with new java-storage client.

Although there is a way to send unauthorized request to new storage client will add code changes for it.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
singhravidutt: Sure. Instead of using boolean using `ClientType`. Which will make sure that these tests will run against new `ClientType` automatically.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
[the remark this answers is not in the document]
singhravidutt: We can. Will do.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -52,6 +54,8 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
   private final GoogleCloudStorageOptions storageOptions;
   private final Storage storage;
 
+  private final ErrorTypeExtractor errorExtractor;
```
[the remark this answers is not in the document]
singhravidutt: added.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -65,10 +67,12 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
   public GoogleCloudStorageClientReadChannel(
       Storage storage,
       GoogleCloudStorageItemInfo itemInfo,
-      GoogleCloudStorageReadOptions readOptions)
+      GoogleCloudStorageReadOptions readOptions,
+      ErrorTypeExtractor errorExtractor)
       throws IOException {
     validate(itemInfo);
     this.storage = storage;
+    this.errorExtractor = errorExtractor;
```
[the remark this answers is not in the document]
singhravidutt: As it will not be accessed/initialized outside the scope of this library kind of avoiding necessary null checks.

cloudbuild/cloudbuild.yaml
```diff
@@ -25,6 +25,7 @@ steps:
     entrypoint: 'bash'
     args: ['/hadoop-connectors/cloudbuild/presubmit.sh', 'unittest']
     env:
+      - 'GCS_TEST_PROJECT_ID=$PROJECT_ID'
```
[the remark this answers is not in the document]
singhravidutt: It was required for one of the unit-test. Just for sanity I can provide a dummy value against unit-test projectId. It's just should not be `null`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -115,6 +116,26 @@ public class GoogleCloudStorageFileSystemImpl implements GoogleCloudStorageFileS
...
       comparing(FileInfo::getPath, PATH_COMPARATOR);
 
+  private static GoogleCloudStorage createCloudStorage(
+      GoogleCloudStorageFileSystemOptions options,
+      Credentials credentials,
+      Function<List<AccessBoundary>, String> downscopedAccessTokenFn)
+      throws IOException {
+    checkNotNull(options, "options must not be null");
+    if (options.getClientType() == ClientType.STORAGE_CLIENT) {
+      return GoogleCloudStorageClientImpl.builder()
```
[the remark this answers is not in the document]
singhravidutt: configuration for ClientType will not change at run time. Just logging what client is getting used based on a static config seems unnecessary to me.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemXAttrsIntegrationTest.java
singhravidutt: Updated.

cloudbuild/cloudbuild.yaml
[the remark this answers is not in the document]
medb: How it run before w/o this var then?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
```diff
@@ -1044,7 +1056,7 @@ public void testGlobStatus() throws IOException {
     createFile(new Path("/directory1/subdirectory2/file1"), data);
     createFile(new Path("/directory1/subdirectory2/file2"), data);
 
-    FileStatus[] rootDirectories = ghfs.globStatus(new Path("/d*"));
+    FileStatus[] rootDirectories = ghfs.globStatus(new Path("/directory1*"));
```
medb, line 1059: Why do we need this change?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -48,13 +50,13 @@
 class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
-
```
medb, line 51: nit: revert

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -535,19 +539,33 @@ private static void validate(GoogleCloudStorageItemInfo itemInfo) throws IOExcep
...
           String.format(
               "Cannot read GZIP-encoded file (%s) (not supported via gRPC API): %s",
               contentEncoding, resourceId));
     }
   }
 
+  private IOException convertError(Exception error) {
+    String msg = String.format("Error reading '%s'", resourceId);
+    switch (errorExtractor.getErrorType(error)) {
+      case NON_FOUND:
```
medb, line 559: NOT_FOUND?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -115,6 +116,26 @@ public class GoogleCloudStorageFileSystemImpl implements GoogleCloudStorageFileS
   static final Comparator<FileInfo> FILE_INFO_PATH_COMPARATOR =
       comparing(FileInfo::getPath, PATH_COMPARATOR);
 
+  private static GoogleCloudStorage createCloudStorage(
+      GoogleCloudStorageFileSystemOptions options,
+      Credentials credentials,
+      Function<List<AccessBoundary>, String> downscopedAccessTokenFn)
+      throws IOException {
+    checkNotNull(options, "options must not be null");
+    if (options.getClientType() == ClientType.STORAGE_CLIENT) {
```
medb, line 125: Use switch as this is enum

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -124,12 +145,7 @@ public class GoogleCloudStorageFileSystemImpl implements GoogleCloudStorageFileS
    */
   public GoogleCloudStorageFileSystemImpl(
       Credentials credentials, GoogleCloudStorageFileSystemOptions options) throws IOException {
-    this(
-        GoogleCloudStorageImpl.builder()
-            .setOptions(checkNotNull(options, "options must not be null").getCloudStorageOptions())
-            .setCredentials(credentials)
-            .build(),
-        options);
+    this(createCloudStorage(options, credentials, null), options);
```
medb, line 148: nit: add comment to `null` param: `/* downscopedAccessTokenFn= */`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemHTTPClientNewIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemHTTPClientNewIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.BeforeClass;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Test suites of integration tests for {@link GoogleCloudStorageFileSystemImpl} class configured to
+ * use Apiary client to connect to gcs server.
+ */
+@RunWith(JUnit4.class)
+public class GoogleCloudStorageFileSystemHTTPClientNewIntegrationTest
```
medb, line 23: nit: do not capitalize `Http`: `GoogleCloudStorageFileSystemHttpClientNewIntegrationTest`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemJavaStorageNewIntegrationTest.java
```diff
@@ -0,0 +1,68 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
```diff
@@ -174,10 +184,13 @@ private void validateFileInfoInternal(
...
 
+      // Causing flakiness for unittest
+      // Comparing just upto milliseconds as that's the least count we have for
+      // `fileModificationTime`
       assertWithMessage(
               "getModificationTime for bucketName '%s' objectName '%s'", bucketName, objectName)
-          .that(fileModificationTime)
-          .isAtLeast(testStartTime);
+          .that(fileModificationTime.toEpochMilli())
+          .isAtLeast(testStartTime.toEpochMilli());
```
medb, line 193: This looks like a hack. Can we fix root cause of the flakiness somehow?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageHTTPClientNewIntegrationTest.java
```diff
@@ -0,0 +1,59 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageHTTPClientNewIntegrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.BeforeClass;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Integration tests for GoogleCloudStorageFileSystem. GoogleCloudStorageFileSystemIntegrationHelper
+ * is configured to use Apiary client to connect GCS server.
+ */
+@RunWith(JUnit4.class)
+public class GoogleCloudStorageHTTPClientNewIntegrationTest
```
medb, line 21: Do not capitalize `HTTP`

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageJavaStorageNewIntegrationTest.java
```diff
@@ -0,0 +1,152 @@
+package com.google.cloud.hadoop.gcsio;
```
medb, line 1: Add license header

util/pom.xml
```diff
@@ -83,5 +83,9 @@
       <groupId>com.google.truth</groupId>
       <artifactId>truth</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+    </dependency>
```
medb, line 89: Why do we need this dep?

util/src/main/java/com/google/cloud/hadoop/util/ErrorType.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.util;
+
+/** Various error-types Exceptions would be mapped against. */
+public enum ErrorType {
```
medb, line 20: Maybe we can declare this as a nested enum in `ErrorTypeExtractor`?

medb (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
singhravidutt: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
singhravidutt: Sure. Updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
singhravidutt: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemHTTPClientNewIntegrationTest.java
singhravidutt: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemHTTPClientNewIntegrationTest.java
singhravidutt: added.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemJavaStorageNewIntegrationTest.java
singhravidutt: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageHTTPClientNewIntegrationTest.java
singhravidutt: added.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageHTTPClientNewIntegrationTest.java
singhravidutt: Done.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageJavaStorageNewIntegrationTest.java
singhravidutt: Added.

util/pom.xml
singhravidutt: We are using `io.grpc.Status` while extracting error type.

util/src/main/java/com/google/cloud/hadoop/util/ErrorType.java
singhravidutt: Yeah, we can do that.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
singhravidutt: So, basically we are running this against integration test and unittest both. For integration tests it will run smoothly as bucket creation will take some time > 1ms. For Unit test as `testStartTime` will be created per Test it's possible that difference in time would be in nano-sec and we will not be able to capture it as `fileModificationTime` is accurate till `milliSec` and not `nanoSec`.

It wasn't an issue earlier when we had `testStartTime` as a static as it will initialised before every test and it reduces the chances of flakiness.

I can still make it static and reduce the chances of flakiness of this unit tests. But that will be mitigation not a fix for flakiness.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
singhravidutt: We have changed bucket name from `gcsio-test` --> `dataproc-gcsio-test`.

So, earlier `d*` would only provide `directory1` whcich was created in test. But with bucket name cchange `d*` would provide two objects `directory1` and `dataproc-gcsio-test..`. Which is causing test to fail.

------

Why a change in bucket name is affecting the directories inside the bucket?

* Test class [`GoogleHadoopFileSystemIntegrationTest`](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java#L111) where we create a bucket using URI. We not only create the bucket but also endup creating an object with same name as bucketName. and this becomes an issue.

cloudbuild/cloudbuild.yaml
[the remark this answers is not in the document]
singhravidutt: This was working earlier as we use to have a [separate function](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTestBase.java#L61) to override the values of projectId. I got rid of that as I have to introduce another flavour of `loadConfig` with `clientType` support and it wasn't scaling well with all variations which were there.

Adding a env variable even for unit-test made sense instead of overriding them for a individual test. I understand it may lead to actual calls getting made to gcs but we can avoid that by providign a dummy value here.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
```diff
@@ -174,6 +184,9 @@ private void validateFileInfoInternal(
 
       Instant fileModificationTime = Instant.ofEpochMilli(fileInfo.getModificationTime());
 
+      // Causing flakiness for unittest
+      // Comparing just upto milliseconds as that's the least count we have for
+      // `fileModificationTime`
```
medb, line 189: nit: remove this comment?

cloudbuild/cloudbuild.yaml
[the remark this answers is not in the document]
medb: I think that this is suboptimal, as we end up requiring to set this var when we run unit tests in IDEs, etc. Is there an other way to fix it w/o requiring this var for init tests?

This made me thing that maybe we can parametrize tests with client type using env var instead of hacking this in code? I.e. we will need to run test twice with 2 different env vars to test 2 different client types, but we will not need to change almost any code, WDYT?

medb (changes requested):

cloudbuild/cloudbuild.yaml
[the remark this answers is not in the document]
singhravidutt: > I think that this is suboptimal, as we end up requiring to set this var when we run unit tests in IDEs, etc. Is there an other way to fix it w/o requiring this var for init tests?

Will investigate.

> This made me thing that maybe we can parametrize tests with client type using env var instead of hacking this in code? I.e. we will need to run test twice with 2 different env vars to test 2 different client types, but we will not need to change almost any code, WDYT?

1. Adding a filesystem Configuration variable as env variable seems a little odd.
2. Both the client don't have exact parity on the level test are written for eg. with HTTP client we can have interceptor support but same support is not there with java-storage (or even a GAPIC level). We are relying heavily on interceptors to trace requests during testing.
3. Overall test execution time will increase for sure as it will run those test too which wouldn't depend on `ClientType`. For 2.2.x we already have two flavour `hadoop2` and `hadoop3`. Introducing `ClientType` will make total runs to be 4.

I will still prefer `Client type` to be part of parametrization and applicable only to relevant test.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemIntegrationTest.java
singhravidutt: removed.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTestBase.java
```diff
@@ -53,27 +53,23 @@ public abstract class GoogleHadoopFileSystemTestBase extends HadoopFileSystemTes
...
-    if (serviceAccountJsonKeyFile != null) {
-      config.setEnum("fs.gs.auth.type", AuthenticationType.SERVICE_ACCOUNT_JSON_KEYFILE);
-      config.set("fs.gs.auth.service.account.json.keyfile", serviceAccountJsonKeyFile);
-    }
-    config.setBoolean("fs.gs.implicit.dir.repair.enable", true);
-    // Allow buckets to be deleted in test cleanup:
-    config.setBoolean("fs.gs.bucket.delete.enable", true);
+  protected static Configuration loadConfig(ClientType storageClientType) {
+    Configuration config = loadConfig();
+    config.setEnum(GCS_CLIENT_TYPE.toString(), storageClientType);
```
medb, line 61: Inline property name the same as rest of the properties are inlined.

medb (approved):

PR #969, opened 2023-02-23
Support for RequesterPays in java-storage client

singhravidutt:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -102,7 +102,12 @@ public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOpti
 
     GoogleCloudStorageClientWriteChannel channel =
         new GoogleCloudStorageClientWriteChannel(
-            storage, storageOptions, resourceIdWithGeneration, options, backgroundTasksThreadPool);
+            storage,
+            storageOptions,
+            resourceIdWithGeneration,
+            options,
+            requesterShouldPay(resourceIdWithGeneration.getBucketName()),
```
arunkumarchacko, line 109: why not do this during init and not do every time we create?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -134,7 +139,11 @@ private SeekableByteChannel open(
       GoogleCloudStorageReadOptions readOptions)
       throws IOException {
     return new GoogleCloudStorageClientReadChannel(
-        storage, itemInfo == null ? getItemInfo(resourceId) : itemInfo, readOptions);
+        storage,
+        itemInfo == null ? getItemInfo(resourceId) : itemInfo,
+        readOptions,
+        storageOptions,
+        requesterShouldPay(resourceId.getBucketName()));
```
arunkumarchacko, line 146: same here

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -184,6 +193,10 @@ private static Storage createStorage(
         .getService();
   }
 
+  private boolean requesterShouldPay(String bucketName) {
+    return ((GoogleCloudStorageImpl) getDelegate()).requesterShouldPay(bucketName);
```
arunkumarchacko, line 197: is this guaranteed to be of this type?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -494,6 +501,9 @@ private BlobSourceOption[] generateReadOptions(BlobId blobId) {
       if (blobId.getGeneration() != null) {
         readOptions.add(BlobSourceOption.generationMatch(blobId.getGeneration()));
       }
+      if (requesterPays) {
+        readOptions.add(BlobSourceOption.userProject(storageOptions.getRequesterPaysOptions().getProjectId()));
```
arunkumarchacko, line 505: is it ok to mutate this? If this object is getting reused, the next Object will also have value set, correct?

Mutating the value of the input parameter is an anti-pattern.

arunkumarchacko (commented):
What about testing? Why the existing tests did not fail for the JavaStrorageClient based Impl?

PR #974, opened 2023-03-06
Java8 ByteBuffer.flip returnType

singhravidutt:
Updating use of `BuyteBuffer.flip()` for Java8 https://github.com/plasma-umass/doppio/issues/497#issuecomment-334740243 explains the issue very well.


abmodi (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientWriteChannel.java
```diff
@@ -131,7 +132,7 @@ public Boolean call() throws Exception {
           }
           byteBuffer.put(data.toByteArray());
           // switch to read mode
-          byteBuffer.flip();
+          ((Buffer) byteBuffer).flip();
```
arunkumarchacko, line 135: Why we are doing this? What will happen if we do not do this?

Also, add a comment here.

singhravidutt: It is covered as part of PR description.

PR #981, opened 2023-03-20
disabled TD only for CloudBuild

singhravidutt:
Enabled traffic director by default and disabled it just for cloud-builds.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/TestConfiguration.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/TestConfiguration.java
@@ -22,6 +22,8 @@ public abstract class TestConfiguration {
 
   public static final String GCS_TEST_JSON_KEYFILE = "GCS_TEST_JSON_KEYFILE";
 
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/TestConfiguration.java
```diff
@@ -43,6 +45,16 @@ public String getPrivateKeyFile() {
     public String getServiceAccountJsonKeyFile() {
       return System.getenv(GCS_TEST_JSON_KEYFILE);
     }
+
+    /** By default, TD is enabled unless specified specifically. */
+    @Override
+    public boolean isTrafficDirector() {
+      String td = System.getenv(TRAFFIC_DIRECTOR_ENABLED);
+      if (td != null && td.equalsIgnoreCase("false")) {
```
arunkumarchacko, line 53: nit: jus return? 

ie. return td == null || td.equalsIgnoreCase("true")

arunkumarchacko (approved):

PR #994, opened 2023-05-02
support to read encoded content via java-storage library

singhravidutt:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -81,8 +85,17 @@ public GoogleCloudStorageClientReadChannel(
             itemInfo.getBucketName(), itemInfo.getObjectName(), itemInfo.getContentGeneration());
     this.readOptions = readOptions;
     this.storageOptions = storageOptions;
-    this.objectSize = itemInfo.getSize();
     this.contentReadChannel = new ContentReadChannel(readOptions, resourceId);
+    initMetadata(itemInfo.getContentEncoding(), itemInfo.getSize());
+  }
+
+  protected void initMetadata(@Nullable String encoding, long sizeFromMetadata) throws IOException {
+    gzipEncoded = nullToEmpty(encoding).contains(GZIP_ENCODING);
```
medb, line 93: This is weird check, maybe `Objects.equal(...)` is more readable?

medb (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
singhravidutt: It's inherited from https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/998eb64cdf8d528f6042195466c9aa90f2aff466/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L776

If I have to guess why `contains` instead of `equals` is used is because there could be multiple type of encoding used against an object.

PR #1018, opened 2023-06-23
upgrade java-storage version (#1017)

singhravidutt:


arunkumarchacko (approved):

pom.xml
```diff
@@ -374,6 +374,12 @@
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-bigquerystorage</artifactId>
         <version>${google.cloud-bigquerystorage.version}</version>
+        <exclusions>
```
arunkumarchacko, line 377: If we exclude this, will the Hive BQ connector still work?

Please add a comment and also a tracking bug to remove this.

singhravidutt: Added tracking bug.

We do have test case showcasing that all the bq workflow works even with 1.56 but it can't still upgrade the version till aug.

PR #1038, opened 2023-07-31
upgrade java-storage

singhravidutt:


gcs/CHANGES.md
```diff
@@ -122,6 +122,8 @@
 
 1.  Upgrade java-storage to 2.22.3
 
+1.  Upgrade java-storage to 2.25.0
```
arunkumarchacko, line 125: should we just update the existing entry since otherwise there will be two upgrade messages in the release notes?

singhravidutt: sure

arunkumarchacko (approved):

PR #1043, opened 2023-08-21
java-storage client resource cleanup

singhravidutt:


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
```diff
@@ -555,6 +564,7 @@ public void create_gcsItemInfo_metadataEquals() throws IOException {
                         /* finishWrite= */ true))
                 .build()
                 .toArray());
+    trackingGcs.delegate.close();
```
arunkumarchacko, line 567: this wont close if there is a test failure, correct?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -158,7 +158,12 @@ private SeekableByteChannel open(
   public void close() {
     try {
       try {
+        if (storage != null) {
+          storage.close();
+        }
         super.close();
+      } catch (Exception e) {
+        throw new RuntimeException("Error occurred while closing the storage client", e);
```
arunkumarchacko, line 166: The pattern we use at other places is to log and ignore the exception correct?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
```diff
@@ -628,6 +628,7 @@ public void testInitializeThrowsWhenCredentialsNotFound() throws URISyntaxExcept
 
     assertThat(thrown).hasMessageThat().startsWith("No enum constant ");
     assertThat(thrown).hasMessageThat().contains("AuthenticationType.INVALID_AUTH_TYPE");
+    ghfs.close();
```
arunkumarchacko, line 631: This wont close if there is test failure, correct?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -158,7 +158,12 @@ private SeekableByteChannel open(
   public void close() {
     try {
       try {
+        if (storage != null) {
+          storage.close();
```
arunkumarchacko, line 162: If this throws, then super.close wont get called, correct?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageImplTest.java
singhravidutt: yeah, but I am not too much worried. Test failure will any how will take priority over resource cleanup.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
singhravidutt: same as below.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
arunkumarchacko: Any reason why we are converting this to RuntimeException? Caller usually do not catch this and I think hence this does not look correct.

singhravidutt: I am specifically converting the exception here because as per our interface we don't throw from `close` with new client's integration they introduced resource cleanup in `close` as well as it throws too.
On second thought, we can absorb the exception here and log it too under `WARN`, as not cleaning up the resource should not cause whole workflow failure.

arunkumarchacko (approved):

PR #1047, opened 2023-09-01
Improve makeQualified

yigress:
makeQualified is in most code path. This improves its performance by removing unnecessary string operations.

1000 random lengths random paths
measure in Nanos
before:
min=1666.0, max=131334.0, median=3625.0
after:
min=333.0, max=57375.0, median=459.0

after the change, for majority use cases the time is weighing towards the min as there is minimum string operations.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -994,21 +994,24 @@ public Path makeQualified(Path path) {
...
-      sb.delete(0, 3);
+      i += 3;
     }
-
-    String strippedPath = sb.toString();
-
-    // Allow a Path of gs://someBucket to map to gs://someBucket/
-    if (strippedPath.equals("/..") || strippedPath.equals("")) {
-      strippedPath = "/";
+    if (i == upath.length() || upath.substring(i).equals("/..")) {
```
arunkumarchacko, line 1006: I feel that the new code is harder to understand. How much benefit (in terms of perforamnce) are we getting by having this? Looks like 3 micro seconds? Do we have any benchmarking data to justify this change?

yigress: agree the original code is easier to read. however, as the prefix "/../" check is for hadoop-2 style and it is not a common case, the original code does extra stringbuilder ops for this. my local benchmark shows 3ms for the common use cases. As we have seen deletion slow (for example flink checkpoint deletion can not finish within time for thousands of files deletion), optimize this is necessary.

medb (approved):

PR #1048, opened 2023-09-04
gRPC-core cleanup

singhravidutt:
Cleaning up the old grpc-core integration in favour of java-storage client integration.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -428,40 +410,22 @@ public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOpti
...
-            @Override
-            public Storage.Objects.Insert createRequest(InputStreamContent inputStream)
-                throws IOException {
-              return initializeRequest(
-                  super.createRequest(inputStream), resourceId.getBucketName());
-            }
-          };
-    }
+
+    channel =
```
arunkumarchacko, line 414: Please make assignment part of the declaration

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -341,13 +335,6 @@ public Boolean load(String bucketName) {
         MetricsSink.CLOUD_MONITORING == options.getMetricsSink()
             ? CloudMonitoringMetricsRecorder.create(options.getProjectId(), finalCredentials)
             : new NoOpMetricsRecorder();
-
-    // Create the gRPC stub if necessary;
-    if (options.isGrpcEnabled()) {
```
arunkumarchacko, line 346: Let use remove the java stroage flag and use gRPC flag to use Veneer client? grpc.enable flag will not have any effect, I guess?

arunkumarchacko (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
singhravidutt: updated

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
singhravidutt: I am keeping track of it. As other flag is already shared with customer in preview phase will not change it immediately. Will be updating it in follow up PR.

PR #1069, opened 2023-11-01
Adding GCS Specific Metrics

guljain:
(cherry picked from commit 48891bff6220a309810307753a0f413268d3aba2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ To add a dependency on one of the connectors using Maven, use the following:
     <dependency>
       <groupId>com.google.cloud.bigdataoss</groupId>
       <artifactId>gcs-connector</artifactId>
-      <version>hadoop2-2.2.18</version>
+      <version>hadoop2-2.2.19</version>
     </dependency>
     ```
 
--- a/bigquery/pom.xml
+++ b/bigquery/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigdataoss-parent</artifactId>
-    <version>2.2.19-SNAPSHOT</version>
+    <version>2.2.20-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
--- a/coverage/pom.xml
+++ b/coverage/pom.xml
@@ -7,11 +7,11 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigdataoss-parent</artifactId>
-    <version>2.2.19-SNAPSHOT</version>
+    <version>2.2.20-SNAPSHOT</version>
   </parent>
 
   <artifactId>coverage</artifactId>
-  <version>2.2.19-SNAPSHOT</version>
+  <version>2.2.20-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <profiles>
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+## 2.2.19 - 2024-01-02
+
 1. Upgrade java-storage to 2.29.0
 
 1. Upgrade guava to 32.1.2-jre
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigdataoss-parent</artifactId>
-    <version>2.2.19-SNAPSHOT</version>
+    <version>2.2.20-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
@@ -32,7 +32,7 @@
   </description>
 
   <artifactId>gcs-connector</artifactId>
-  <version>${hadoop.identifier}-2.2.19-SNAPSHOT</version>
+  <version>${hadoop.identifier}-2.2.20-SNAPSHOT</version>
 
   <profiles>
     <profile>
--- a/gcsio/pom.xml
+++ b/gcsio/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigdataoss-parent</artifactId>
-    <version>2.2.19-SNAPSHOT</version>
+    <version>2.2.20-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
@@ -32,7 +32,7 @@
   </description>
 
   <artifactId>gcsio</artifactId>
-  <version>2.2.19-SNAPSHOT</version>
+  <version>2.2.20-SNAPSHOT</version>
 
   <profiles>
     <profile>
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
 
   <groupId>com.google.cloud.bigdataoss</groupId>
   <artifactId>bigdataoss-parent</artifactId>
-  <version>2.2.19-SNAPSHOT</version>
+  <version>2.2.20-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <url>https://github.com/GoogleCloudDataproc/hadoop-connectors</url>
@@ -80,7 +80,7 @@
 
     <gpg.skip>true</gpg.skip>
 
-    <bigdataoss.version>2.2.19-SNAPSHOT</bigdataoss.version>
+    <bigdataoss.version>2.2.20-SNAPSHOT</bigdataoss.version>
     <bigquery-connector.version>1.2.1-SNAPSHOT</bigquery-connector.version>
     <apache.httpclient.version>4.5.13</apache.httpclient.version>
     <google.api.grpc.proto-google-iam-v1.version>1.1.7</google.api.grpc.proto-google-iam-v1.version>
--- a/util-hadoop/pom.xml
+++ b/util-hadoop/pom.xml
@@ -22,14 +22,14 @@
   <parent>
     <groupId>com.google.cloud.bigdataoss</groupId>
     <artifactId>bigdataoss-parent</artifactId>
-    <version>2.2.19-SNAPSHOT</version>
+    <version>2.2.20-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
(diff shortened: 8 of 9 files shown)


arunkumarchacko (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -120,9 +155,75 @@ private void resetMetrics(Map<String, AtomicLong> metrics) {
...
 
+  /**
+   * Updating the required gcs specific statistics based on httpresponse.
+   *
+   * @param statusCode
+   */
+  private void updateGcsIOSpecificStatistics(int statusCode) {
+
+    if (statusCode == 429) {
+      incrementRateLimitingCounter();
```
arunkumarchacko, line 166: nit: Can consider moving this inside the if in line 173. Once we do that we can return from the if statements. Not a big saving, but can reduce the number of if.

If this can get statuscode=200. Also can add another condition of <=200 and return. This will be most frequent operation, I guess.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -120,9 +155,75 @@ private void resetMetrics(Map<String, AtomicLong> metrics) {
...
+
+  /**
+   * Updating the required gcs specific statistics based on GoogleJsonResponseException.
+   *
+   * @param responseException contains statusCode based on which metrics are updated
+   */
+  @Subscribe
+  private void subsriberOnGoogleJsonResponseException(
+      GoogleJsonResponseException responseException) {
+    updateGcsIOSpecificStatistics(responseException.getStatusCode());
```
arunkumarchacko, line 191: Whould we check for responseException null? Also can getStatuCode be null?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -132,6 +133,7 @@ public synchronized int read() throws IOException {
       return -1;
     }
     if (numRead != 1) {
+      GoogleCloudStorageEventBus.postOnException(new Exception());
```
arunkumarchacko, line 136: Let us not create the Exception object. Why not add another method, which does not take the actual exception?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -344,12 +345,15 @@ public GoogleHadoopFileSystemBase() {
...
       storageStatistics = (GhfsStorageStatistics) globalStats;
+
     } else {
       logger.atWarning().log(
           "Encountered an error while registering to GlobalStorageStatistics. Some of the GCS connector metrics will not be reported to metrics sinks. globalStatsClassLoader=<%s>; classLoader=<%s>",
           globalStats.getClass().getClassLoader(), GhfsStorageStatistics.class.getClassLoader());
       storageStatistics = GhfsStorageStatistics.DUMMY_INSTANCE;
     }
+
+    GoogleCloudStorageEventBus.register(storageStatistics);
```
arunkumarchacko, line 356: Let us not rely on Eventbus to dedup the subscribers? Why not do the registration in GhfsStorageStatistics?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageStatistics.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.EnumSet;
+
+/** Statistics which are collected in GCS Client Side. */
+public enum GoogleCloudStorageStatistics {
+
+  /** Client-side Status Code statistics */
+  GCS_REQUEST_COUNT(
+      "gcs_total_request_count", "Counts the total number of gcs requests made", TYPE_COUNTER),
+
+  GCS_EXCEPTION_COUNT(
```
arunkumarchacko, line 33: Should have move the exception count out of GCS? i.e. let us have an exception count for entire connector? I agree having a bifurcation for GCS related exceptions will be useful. We would like to know connector level exception count as well.

util/src/main/java/com/google/cloud/hadoop/util/GoogleCloudStorageEventBus.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.eventbus.EventBus;
+import com.google.common.flogger.GoogleLogger;
+
+/** Event Bus class */
+public class GoogleCloudStorageEventBus {
+
+  /** Hold the instance of the event bus here */
+  private static EventBus eventBus = new EventBus();
+
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
```
arunkumarchacko, line 15: We are not using this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageStatistics.java
guljain: So, would it be better to bifurcate now into EXCEPTION_COUNT and GCS_EXCEPTION_COUNT ? Other thing we can do is to go with EXCEPTION_COUNT and later in future define another counter specific for GCS.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
guljain: StatusCode 200 is request succeeded, so maybe statusCode < 200 could have a return statement instead ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
guljain: Enclosed the subscriber methods in class and registered in GhfsStorageStatistics.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
guljain: Since the usage of this Subscriber method makes sure that responseException is never null, hence skipping the check. I do not think that getStatusCode() can return null but still added check.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -120,9 +156,73 @@ private void resetMetrics(Map<String, AtomicLong> metrics) {
...
+   * @param statusCode
+   */
+  private void updateGcsIOSpecificStatistics(int statusCode) {
+
+    if (statusCode < 200) {
+      return;
+    }
+
+    if (200 <= statusCode && statusCode < 300) {
+      incrementGcsSuccessCodeCount();
```
arunkumarchacko, line 171: return after this?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -331,7 +334,7 @@ public GoogleCloudStorageImpl(
    * @param storage {@link Storage} to use for I/O.
    */
   public GoogleCloudStorageImpl(GoogleCloudStorageOptions options, Storage storage) {
-    this(options, storage, /* credentials= */ null, /* downscopedAccessTokenFn= */ null);
```
arunkumarchacko, line 334: why type casting? Also why remove the param name comment?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -719,9 +728,12 @@ public void createEmptyObjects(
                     ioe.getClass().getSimpleName());
                 logger.atFine().withCause(ioe).log("Ignored exception while creating empty object");
               } else {
+                hasException.set(true);
```
arunkumarchacko, line 731: So we will count this only once?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -999,13 +1011,17 @@ protected InputStream openStream(long bytesToRead) throws IOException {
           }
           break;
         } catch (IOException footerException) {
+          hasException = true;
```
arunkumarchacko, line 1014: why not always increment?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1057,11 +1073,16 @@ protected InputStream openStream(long bytesToRead) throws IOException {
 
       return contentStream;
     } catch (IOException e) {
+      hasException = true;
```
arunkumarchacko, line 1076: Instead of having this variable, why not increment here?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
guljain: Yes. This is one way of doing it. I saw that we are already having "innerExceptions" storing exception so we may not need "hasException" variable even. Updated it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
guljain: Seems like we throw exception only after maximum retries is reached so it made more sense to increment EXCEPTION_COUNT in that case only otherwise we will be incrementing at every retry which is not what EXCEPTION_COUNT is supposed to do. Please let me know if this can be done in better way.

arunkumarchacko (approved):

PR #1076, opened 2023-11-23
Move list bucket methods to java-storage.

shilpi23pandey:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -184,6 +188,79 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+   * Shared helper for actually dispatching buckets list API calls and accumulating paginated
+   * results; these can then be used to either extract just their names, or to parse into full
+   * GoogleCloudStorageItemInfos.
+   */
+  private List<Bucket> listBucketsInternal() throws IOException {
+    logger.atFiner().log("listBucketsInternal()");
+    checkNotNull(storageOptions.getProjectId(), "projectId must not be null");
+    List<Bucket> allBuckets = new ArrayList<>();
+    try {
+      Page<Bucket> buckets =
```
arunkumarchacko, line 225: can there be more than one page or is it taken care of by the iterator?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -184,6 +188,79 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+          storage.list(
+              BucketListOption.pageSize(storageOptions.getMaxListItemsPerCall()),
```
arunkumarchacko, line 227: What is the max allowed?

shilpi23pandey: 1000 for bucket fields without acl and 200 with acl. We use non-acl fields so it is 1000 for this use case.

arunkumarchacko (approved):

PR #1087, opened 2023-12-19
Add java-storage implementation for copy objects method.

shilpi23pandey:
`java-storage` client internally uses gcs rewrite API for copy so we don't need the logic to decide weather to issue a copy or rewrite as opposed to the APIary implementation. https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.Storage#com_google_cloud_storage_Storage_copy_com_google_cloud_storage_Storage_CopyRequest_

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
@@ -18,6 +18,8 @@ package com.google.cloud.hadoop.gcsio;
 
 import static com.google.cloud.hadoop.gcsio.GoogleCloudStorageExceptions.createFileNotFoundException;
 import static com.google.cloud.hadoop.gcsio.GoogleCloudStorageImpl.decodeMetadata;
+import static com.google.cloud.hadoop.gcsio.GoogleCloudStorageImpl.encodeMetadata;
+import static com.google.cloud.hadoop.gcsio.GoogleCloudStorageImpl.validateCopyArguments;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
@@ -47,6 +49,7 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.BucketInfo.LifecycleRule.LifecycleAction;
 import com.google.cloud.storage.BucketInfo.LifecycleRule.LifecycleCondition;
+import com.google.cloud.storage.CopyWriter;
 import com.google.cloud.storage.ParallelCompositeUploadBlobWriteSessionConfig.BufferAllocationStrategy;
 import com.google.cloud.storage.ParallelCompositeUploadBlobWriteSessionConfig.ExecutorSupplier;
 import com.google.cloud.storage.ParallelCompositeUploadBlobWriteSessionConfig.PartCleanupStrategy;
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -209,6 +212,105 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+    executor.queue(
+        () -> {
+          try {
+            String srcString = StringPaths.fromComponents(srcBucketName, srcObjectName);
+            String dstString = StringPaths.fromComponents(dstBucketName, dstObjectName);
+
+            CopyWriter copyWriter = storage.copy(copyRequestBuilder.build());
+            while (!copyWriter.isDone()) {
+              copyWriter.copyChunk();
+              logger.atFiner().log(
```
arunkumarchacko, line 292: let us make this finest?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -209,6 +212,105 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+          BlobId.of(dstBucketName, dstObjectName),
+          BlobTargetOption.generationMatch(dstContentGeneration));
+    } else {
+      copyRequestBuilder.setTarget(BlobId.of(dstBucketName, dstObjectName));
+    }
+
+    if (storageOptions.getMaxRewriteChunkSize() > 0) {
+      copyRequestBuilder.setMegabytesCopiedPerChunk(
+          // Convert raw byte size into Mib.
+          storageOptions.getMaxRewriteChunkSize() / (1024 * 1024));
```
arunkumarchacko, line 281: If this is less than 1MB, it will be 0? Is that ok?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -209,6 +212,105 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+        ConcurrentHashMap.newKeySet();
+
+    BatchExecutor executor = new BatchExecutor(storageOptions.getBatchThreads());
+
+    try {
+      for (Map.Entry<StorageResourceId, StorageResourceId> entry :
+          sourceToDestinationObjectsMap.entrySet()) {
+        StorageResourceId srcObject = entry.getKey();
+        StorageResourceId dstObject = entry.getValue();
+        copyInternal(
```
arunkumarchacko, line 241: Why are we not use Rewrite if copyWithRewrite config is enabled?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: We do have checks which enforces that `getMaxRewriteChunkSize` should be multiple of MiB so I don't think it will ever be zero. https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/babcabd1f922d65852c7f9e288b2949844f30524/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java#L236

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: Veneer does not expose separate methods for rewrite and copy, it internally always uses **rewrite** and decides number of calls based on the object size. The APIary copy also uses rewrite API but with a single RPC call https://cloud.google.com/storage/docs/json_api/v1/objects/copy

arunkumarchacko (approved):

PR #1094, opened 2023-12-27
Add java-storage implementation for createEmptyObjects implementaion.

shilpi23pandey:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -219,6 +223,178 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+            .build(),
+        blobTargetOptions.toArray(BlobTargetOption[]::new));
+  }
+
+  /**
+   * Helper to check whether an empty object already exists with the expected metadata specified in
+   * {@code options}, to be used to determine whether it's safe to ignore an exception that was
+   * thrown when trying to create the object, {@code exceptionOnCreate}.
+   */
+  private boolean canIgnoreExceptionForEmptyObject(
```
arunkumarchacko, line 349: We are doing a lot of code duplication. Is there a way to avoid it?

arunkumarchacko (approved):

arunkumarchacko (commented):
What about tests? Code coverage is way too low.

shilpi23pandey:
> What about tests? Code coverage is way too low.
Some ITs relied on the common implementation so added that code in the GoogleCloudStorasgeClient. Also added unit tests for different cases.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -219,6 +223,178 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
...
+      StorageException exceptionOnCreate, StorageResourceId resourceId, CreateObjectOptions options)
+      throws IOException {
+    if (errorExtractor.getErrorType(exceptionOnCreate) == ErrorType.RESOURCE_EXHAUSTED
```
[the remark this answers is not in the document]
shilpi23pandey: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: I believe, we can avoid this duplication by refactoring the GoogleCloudStorageImpl at bit. However, given that we might deprecate GoogleCloudStorageImpl in future. I dont see merit in investing in this re-factoring at the moment. Let me know if you have any other thoughts.

arunkumarchacko (approved):

PR #1095, opened 2023-12-27
Add unit tests for GoogleCloudStorageClientImpl.

shilpi23pandey:
1. Adds unit tests for
* listBucket
* deleteBucket
* deleteObjects
* updateObjects
* compose
* copy

2. Add redundant APIs for compose and copy so that GoogleCloudStorageClient implementation is invoked for all APIs.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
@@ -308,6 +308,14 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
       copyRequestBuilder.setTarget(BlobId.of(dstBucketName, dstObjectName));
     }
 
+    if (storageOptions.getEncryptionKey() != null) {
+      copyRequestBuilder.setSourceOptions(
+          BlobSourceOption.decryptionKey(storageOptions.getEncryptionKey().value()));
+      copyRequestBuilder.setTarget(
+          copyRequestBuilder.build().getTarget().getBlobId(),
+          BlobTargetOption.encryptionKey(storageOptions.getEncryptionKey().value()));
+    }
+
     if (storageOptions.getMaxRewriteChunkSize() > 0) {
       copyRequestBuilder.setMegabytesCopiedPerChunk(
           // Convert raw byte size into Mib.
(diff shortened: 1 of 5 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -219,6 +219,33 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
     }
   }
 
+  /**
+   * See {@link GoogleCloudStorage#copy(String, List, String, List)} for details about expected
+   * behavior.
+   */
+  @Override
+  public void copy(
```
arunkumarchacko, line 227: Are we dong more than adding UTs in this PR?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageTest.java
```diff
@@ -1637,7 +1637,7 @@ public void googleCloudStorageItemInfo_toString() throws IOException {
     assertThat(itemInfo2.toString()).contains("exists: no");
   }
 
-  static <K, V> void assertMapsEqual(
+  public static <K, V> void assertMapsEqual(
```
arunkumarchacko, line 1640: Why is this required? I could not see this getting used?

shilpi23pandey: This is used here - https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1095/files#diff-603a2524bd5e4d31fea4c3596b45056b63dfa96f351d6d6816d4a06f0e09daddR556

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: There are some APIs which share some common code between `GoogleCloudStorageImpl` and `GoogleCloudStorageClientImpl`. For example, this copy method implementation is similar for both the clients -they in turn call the https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1095/files#diff-23a518b24b3819c111aacec5ec168906787ef291585a52e128c1ce776b8ed0fbR246 which contains client specific implementation. To reduce code duplication, this code was not  added earlier. However, while writing the UTs realised that with the delegator in place, this method will always call the GoogleCloudStorageImpl copy. So added these implementations here as well.

arunkumarchacko: Do you want to update the PR description accordingly.

arunkumarchacko (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shilpi23pandey: Done

PR #1097, opened 2024-01-03
update interceptor logging for PCU

singhravidutt:
Fixed below 2 issues:

1. PCU had issue in logging `null` uploadIds. 
    * PCUs use direct upload for part files and hence don't have concept of uploadIds. Updated code to have a default uploadId as empty string instead of `null` value.

1. In Direct uploads incocationIds were gettign logged as `NOT-FOUND`.
    * Earlier we were dependent on `streamCreated` to update the Ids for tracing but there are cases when actual stream creation is delayed and sometimes even happened after `onMessage` is called . Updated the logic of updatign invocationid in tracers by updating it as soon as we get the request to start the tracer i.e. `SimpleForwardingClientCall.start`.

_____

Added integration test to very the tracing for PCU calls.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcTracingInterceptor.java
```diff
@@ -275,41 +275,69 @@ private class WriteObjectStreamTracer extends TrackingStreamTracer {
...
+                    response.getResource().getGeneration()));
+      }
+      logRecords.put(GoogleCloudStorageTracingFields.UPLOAD_ID.name, this.streamUploadId);
+      logger.atInfo().log("%s", toJson(logRecords.build()));
     }
 
-    private void updateUploadId(@Nonnull String uploadId) {
-      if (streamUploadId == null) {
+    private void updateStreamUploadId(@Nonnull String uploadId) {
+      if (Strings.isNullOrEmpty(streamUploadId) && !Strings.isNullOrEmpty(uploadId)) {
```
arunkumarchacko, line 340: I guess streamUploadId and uploadId cannot be null?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcTracingInterceptor.java
```diff
@@ -275,41 +275,69 @@ private class WriteObjectStreamTracer extends TrackingStreamTracer {
...
-                  .put(
-                      GoogleCloudStorageTracingFields.FINALIZE_WRITE.name, request.getFinishWrite())
-                  .put(
-                      GoogleCloudStorageTracingFields.CONTENT_LENGTH.name,
-                      request.getChecksummedData().getContent().size())
-                  .build()));
+
+      logRecords
+          .put(GoogleCloudStorageTracingFields.WRITE_OFFSET.name, request.getWriteOffset())
+          .put(GoogleCloudStorageTracingFields.FINALIZE_WRITE.name, request.getFinishWrite())
```
arunkumarchacko, line 304: FInish write is same as finalizewrite?

arunkumarchacko (approved):
The code is way too complex. Can you please add any background information relevant to this so that readers can figure out WHY/HOW this is working.

PR #1102, opened 2024-01-16
Make downscoping work with PCU

arunkumarchacko:


frankyn:
Hi @ejona86, can you help review this PR for interceptor usage?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+
+      @Override
+      public void sendMessage(ReqT message) {
+        setAuthHeader(message);
+        if (headers != null) {
+          delegate().start(responseListener, headers);
+        }
+
+        if (flowControlRequests != 0) {
+          super.request(flowControlRequests);
```
ejona86, line 112: This is assuming the caller does the correct ordering of calls, and if it doesn't you'd probably talk to me to find the bug and I'd have little hope. I know you only care about unary requests, but for easier maintenance it should be a bit more defensive, especially as things change over time.

```java
if (headers != null) {
  setAuthHeader(message);
  super.start(responseListener, headers);
  headers = null;
  if (flowControlRequests != 0) {
    super.request(flowControlRequests);
    // If you are paranoid, you could also:
    flowControlRequests = 0;
  }
}
super.sendMessage(message);
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+    this.downscopingFunction = downscopedAccessToken;
+  }
+
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    logger.atFinest().log(String.format("interceptCall(): method=%s", method.getFullMethodName()));
+    return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+      private int flowControlRequests;
+      private String methodName;
```
ejona86, line 82:
```suggestion
      private final String methodName = method.genFullMethodName();
```

`start()` isn't a constructor, so it is a bit confusing initializing it there.

ejona86 (commented):
The interceptor LGTM. It is known not to handle all cases, but after my suggested change to sendMessage(), things should generally break loudly if used in a way it doesn't support.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+        if (headers != null) {
+          this.flowControlRequests += numMessages;
+        } else {
+          super.request(numMessages);
+        }
+      }
+
+      @Override
+      public void sendMessage(ReqT message) {
+        setAuthHeader(message);
```
singhravidutt, line 106: is stream's header required to be updated for every request message? IIUC, it would only be functional if stream is not created yet.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+          ReadObjectRequest readObjectRequest = (ReadObjectRequest) message;
+          String bucketName = getBucketName(readObjectRequest.getBucket());
+          String objectName = readObjectRequest.getObject();
+
+          return ImmutableList.of(
+              AccessBoundary.create(bucketName, objectName, Action.READ_OBJECTS));
+        }
+
+        if (GOOGLE_STORAGE_V_2_STORAGE_WRITE_OBJECT.equals(methodName)) {
+          WriteObjectSpec writeObjectSpec = ((WriteObjectRequest) message).getWriteObjectSpec();
```
singhravidutt, line 151: for resumable upload specific write requests Resource info will not be there any only uploadId would be present.

Do we have integration test for it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
arunkumarchacko: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    logger.atFinest().log(String.format("interceptCall(): method=%s", method.getFullMethodName()));
+    return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+      private int flowControlRequests;
+      private String methodName;
+      private Metadata headers;
+      private Listener<RespT> responseListener;
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
```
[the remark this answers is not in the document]
arunkumarchacko: This is the recommended way from the SDK team to set the Auth header from Interceptor.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
arunkumarchacko: Auth header is updated only once since this.headers will be set to null.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
arunkumarchacko: Updated. Thank you.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,226 @@
...
+        } else {
+          super.request(numMessages);
+        }
+      }
+
+      @Override
+      public void sendMessage(ReqT message) {
+        setAuthHeader(message);
+        if (headers != null) {
+          delegate().start(responseListener, headers);
```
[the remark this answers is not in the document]
arunkumarchacko: Changed to calle super.start based on the comment from Eric.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
arunkumarchacko: Test this out. I am not seeing the behavior you are describing. We do not have integration tests for Downscoped tokens.

singhravidutt: [`WriteObjectRequest`](https://github.com/googleapis/java-storage/blob/v2.22.1/proto-google-cloud-storage-v2/src/main/java/com/google/storage/v2/WriteObjectRequest.java#L71) captures what writeObjectSpec will capturing i.e. either uploadId or resource

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
```diff
@@ -0,0 +1,224 @@
...
+          String bucketName = getBucketName(readObjectRequest.getBucket());
+          String objectName = readObjectRequest.getObject();
+
+          return ImmutableList.of(
+              AccessBoundary.create(bucketName, objectName, Action.READ_OBJECTS));
+        }
+
+        if (GOOGLE_STORAGE_V_2_STORAGE_WRITE_OBJECT.equals(methodName)) {
+          WriteObjectSpec writeObjectSpec = ((WriteObjectRequest) message).getWriteObjectSpec();
+          com.google.storage.v2.Object obj = writeObjectSpec.getResource();
```
singhravidutt, line 150: @frankyn can you confirm whether we will have Resource attached in WriteObjectSpec for ResumableUpload specific writeObject. By looking into the code and verification via integration test I am able to reach the conclusion that,
*  for Direct Uplaod request (i.e. part file up-load in PCU) WriteObjectRequest have `FirstMessageCase` as 2 (i.e. WRITE_OBJECT_SPEC type) and that contains Resource info
* for resumable upload WriteObjectRequest have `FirstMessageCase` as 1 (i.e. UPLOAD_ID) and do not contain Resource info.

IIUC, object and Bucket extracted here would have empty string value (`""`).

frankyn: @singhravidutt, I think you're referring to the following code block:

```
if (flowControlRequests != 0) {
            super.request(flowControlRequests);
            flowControlRequests = 0;
}
```

This flowControl code was provided by @ejona86; Eric, can you clarify how this variable is used to make sure headers in the initial message are modified? IIUC only the initial message requires the Auth header and not subsequent messages.

frankyn: After our chat, my current hypothesis as to why interceptor works without bucket or object names is that when `upload_id` is provided in resumable upload request in gRPC, auth header is not required similar to [JSON/XML APIs](https://cloud.google.com/storage/docs/performing-resumable-uploads#initiate-session).

That said, we need to confirm with @brandony, does gRPC work similar to JSON/XML that uploading to a resumable session only requires upload_id once the session is created?

image

singhravidutt: given that @arunkumarchacko is able to test it out even for resumable upload. Going ahead with the current changes.

singhravidutt (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
arunkumarchacko: Discussed offline.

BrandonY:
You're correct. No need for an auth header: the upload ID is the credential.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcDownscopingInterceptor.java
BrandonY: You're correct. The upload ID is the credential; no need for an auth header.

PR #1125, opened 2024-03-07
Cleanup for statistics

guljain:
Cleaning up dead code

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
@@ -39,40 +39,22 @@ import org.apache.hadoop.fs.statistics.StreamStatisticNames;
 @InterfaceStability.Unstable
 public enum GhfsStatistic {
   /** HTTP request counters */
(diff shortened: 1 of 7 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
```diff
@@ -39,40 +39,22 @@
...
-      "action_http_post_request_failures", "POST request failure.", TYPE_COUNTER),
-  ACTION_HTTP_DELETE_REQUEST_FAILURE(
-      "action_http_delete_request_failures", "DELETE request failure.", TYPE_COUNTER),
 
   /** FileSystem-level statistics */
   DIRECTORIES_CREATED(
       "directories_created",
       "Total number of directories created through the object store.",
       TYPE_COUNTER),
-  DIRECTORIES_DELETED(
```
arunkumarchacko, line 72: why removing this?

guljain: Directories Deleted isn't used in master, only functions were defined but never actually incremented.

arunkumarchacko (approved):

PR #1131, opened 2024-03-18
Modify logging

guljain:
Avoid logging if the first request for a statistic exceeds threshold.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsGlobalStorageStatistics.java
```diff
@@ -203,7 +203,10 @@ private void updateMinMaxStats(
     String maxKey = getMaxKey(statistic.getSymbol());
     AtomicLong maxVal = maximums.get(maxKey);
     if (maxDurationMs > maxVal.get()) {
-      if (maxDurationMs > LATENCY_LOGGING_THRESHOLD_MS) {
+
+      // Log is avoided if the first request exceedes threshold
+      if (maxDurationMs > LATENCY_LOGGING_THRESHOLD_MS
+          && opsCount.get(statistic.getSymbol()).get() > 0) {
```
arunkumarchacko, line 209: This can never be null, correct?

arunkumarchacko (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsGlobalStorageStatistics.java
guljain: No, we always add all the counters with initial value as 0.

PR #1132, opened 2024-04-01
Modifying FsBenchmark

guljain:
Enhancing fsBenchmark to add p50, p95, p99 latencies along with throughput for random-read


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -241,10 +246,14 @@ private void benchmarkWrite(
...
                         fileBytesWrite += writeSize;
                         writeCallBytes.accept(writeSize);
+
+                        writeFileTime.add(System.nanoTime() - writeCallStart);
                         writeCallTimeNs.accept(System.nanoTime() - writeCallStart);
                       } while (fileBytesWrite < totalSize);
 
                       writeFileBytes.accept(fileBytesWrite);
+
+                      writeFileTime.add(System.nanoTime() - writeStart);
```
arunkumarchacko, line 256: Wont this add twice?

arunkumarchacko, line 256: What is the difference b/w this and writeFileTimeNs?

guljain: Can you elaborate on what can be added twice ?

guljain: From writeFileTimeNs, we cannot calculate p50, p95 etc as from object of class LongSummaryStatistics, we cant get list of values stored imo.

guljain: modified

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -201,6 +202,10 @@ private void benchmarkWrite(
     Set<LongSummaryStatistics> writeCallBytesList = newSetFromMap(new ConcurrentHashMap<>());
     Set<LongSummaryStatistics> writeCallTimeNsList = newSetFromMap(new ConcurrentHashMap<>());
 
+    // for all threads
+    ArrayList<Long> writeFileTime = new ArrayList<>();
+    ArrayList<Long> writeCallTime = new ArrayList<>();
```
arunkumarchacko, line 207: since items are added concurrently, having some form of concurrent list might be better. I understand that currently implementation has the issue.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/FsBenchmark.java
```diff
@@ -241,10 +246,14 @@ private void benchmarkWrite(
                         output.write(writeBuffer);
                         fileBytesWrite += writeSize;
                         writeCallBytes.accept(writeSize);
+
+                        writeCallTime.add(System.nanoTime() - writeCallStart);
```
arunkumarchacko, line 250: This can be huge, correct? Not sure if this is the right thing to do since this can interfere with the benchmark and give a wrong result.

PR #1133, opened 2024-04-02
Remove piped write in when using Java Storage Client

arunkumarchacko:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientWriteChannel.java
```diff
@@ -196,31 +121,23 @@ public void close() throws IOException {
...
   private int writeInternal(ByteBuffer byteBuffer) throws IOException {
     int bytesWritten = writableByteChannel.write(byteBuffer);
     logger.atFinest().log(
         "%d bytes were written out of provided buffer of capacity %d",
         bytesWritten, byteBuffer.limit());
     return bytesWritten;
   }
+
+  @Override
+  public int write(ByteBuffer src) throws IOException {
```
singhravidutt, line 140: How do we guarantee that all bytes requested are actually written?

singhravidutt, line 140: call to storage `writer` is blocking, any other subsequent and may effect the runtime. Have we taken this into account?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientWriteChannelTest.java
```diff
@@ -111,121 +99,32 @@ public static void cleanUp() {
   @Test
   public void writeMultipleChunksSuccess() throws IOException {
     int numberOfChunks = 10;
-    writeChannel.initialize();
     ByteString data =
         GoogleCloudStorageTestHelper.createTestData(
             MAX_WRITE_CHUNK_BYTES.getNumber() * numberOfChunks);
     writeChannel.write(data.asReadOnlyByteBuffer());
     writeChannel.close();
     // Fake writer only writes half the buffer at a time
```
singhravidutt, line 107: `fakeWriteChannel` implementation only writes half a buffer at a time, we should not revert it with just one write call instead fix the code to validate that whole byte buffer is written properly.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientWriteChannelTest.java
singhravidutt, line 1: We should add test for Exception being thrown from java-storage write channel.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImplCreateTest.java
```diff
@@ -215,82 +205,6 @@ public void close() {
     assertThat(thrown).hasCauseThat().isEqualTo(fakeException);
   }
 
-  /**
-   * Test handling when the parent thread waiting for the write operation to finish via the close
-   * call is interrupted, that the actual write is cancelled and interrupted as well.
-   */
-  @Test
-  public void testCreateObjectApiInterruptedException() throws Exception {
```
singhravidutt, line 223: these test cases are used against apairy writer as well as java storage writer. We should be careful while removing test cases from here.

arunkumarchacko: Good point. Missed that.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientWriteChannel.java
arunkumarchacko: This should be fine. the write() interface is coming from WritableByteChannel and the caller will be handling this.

arunkumarchacko: Even the pipe based implementation will block after the buffer fills up. Since the caller is expected to be much faster than network, most of the time the buffering is not likely to help. In fact based on our benchmarks, the non-pipe based implementation is faster and consumed less memory.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImplCreateTest.java
```diff
@@ -248,7 +243,7 @@ public void testCreateObjectApiInterruptedException() throws Exception {
         .thenReturn(
             new FakeWriteChannel() {
               @Override
-              public int write(ByteBuffer src) {
```
singhravidutt, line 251: 1. Is it really need to be handed over to `close` function. 
1. Given we don't have pip anymore for javastorage writer, this test seems to be an overkill for javastorage. Should be bypass it?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageClientImplIntegrationTest.java
```diff
@@ -343,8 +343,9 @@ public void uploadViaPCUInvalidPartFileNamePrefix() throws IOException {
     byte[] bytesToWrite = new byte[partFileCount * bufferCapacity];
     GoogleCloudStorageTestHelper.fillBytes(bytesToWrite);
     WritableByteChannel writeChannel = gcs.create(resourceId);
-    writeChannel.write(ByteBuffer.wrap(bytesToWrite));
```
singhravidutt, line 346: Why it wasn't showing up in `write` earlier? Is it because actual remote wasn't made till `close` but with pipe removal it will?

singhravidutt (approved):
Some minor comments, overall LGTM.

PR #1137, opened 2024-04-17
Version Update for LTS

guljain:


pom.xml
```diff
@@ -458,6 +458,26 @@
         <artifactId>grpc-xds</artifactId>
         <version>${grpc.version}</version>
       </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
```
arunkumarchacko, line 462: why we had to add this?

arunkumarchacko (approved):

PR #1157, opened 2024-05-15
Throwing read exception instead of absorbing 

guljain:
Throwing exception instead of absorbing to avoid getting incorrect exception.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
@@ -134,6 +134,7 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
             }
             response = numRead;
           } catch (IOException e) {
+            System.out.println("huraayyyyyyyyyyyyyyyyyyy");
             streamStatistics.readException();
             throw e;
           }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
@@ -26,6 +26,7 @@ import static org.apache.hadoop.fs.statistics.impl.IOStatisticsBinding.trackDura
 
 import com.google.cloud.hadoop.gcsio.FileInfo;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystem;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.flogger.GoogleLogger;
 import java.io.IOException;
 import java.net.URI;
@@ -84,7 +85,8 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
     return new GoogleHadoopFSInputStream(ghfs, fileInfo.getPath(), channel, statistics);
   }
 
-  private GoogleHadoopFSInputStream(
+  @VisibleForTesting
+  GoogleHadoopFSInputStream(
       GoogleHadoopFileSystem ghfs,
       URI gcsPath,
       SeekableByteChannel channel,
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
@@ -39,6 +39,8 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.statistics.IOStatistics;
 import org.apache.hadoop.fs.statistics.IOStatisticsSource;
 
+@VisibleForTesting
+/* This class should only be called via static function. Creating instance via constructor is not recommended */
 class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSource {
 
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
@@ -85,7 +87,6 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
     return new GoogleHadoopFSInputStream(ghfs, fileInfo.getPath(), channel, statistics);
   }
 
-  @VisibleForTesting
   GoogleHadoopFSInputStream(
       GoogleHadoopFileSystem ghfs,
       URI gcsPath,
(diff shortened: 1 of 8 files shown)


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -200,6 +201,25 @@ public void testGetDefaultPortIndicatesPortsAreNotUsed() throws Exception {
...
 
+  @Test
+  public void read_non_existing_file_throws_exception() throws Exception {
+    URI path = new URI("gs://test-non-existent/read-throws-exception");
+    Configuration configuration = new Configuration();
+
+    // try to read a non-existing file
+    configuration.setBoolean("fs.gs.inputstream.fast.fail.on.not.found.enable", false);
+    GoogleHadoopFileSystem ghfs =
+        GoogleHadoopFileSystemIntegrationHelper.createGhfs(path, configuration);
```
singhravidutt, line 212: This will try creating an actual connection with gcs, which would fail eventually and hence you are seeing `IOException` and test passes. But I believe that's not the intent of this unit test. In this test suite we create InMemory ghfs, try using that.

One suggestion would be to create an inMemeory file of size 1B and try reading from offset 100 this should fail with an exception.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -140,6 +142,62 @@ public void eagerInitialization_fails_withInvalidCredentialsConfiguration() {
...
+          }
+
+          @Override
+          public void close() throws IOException {
+            throw new IOException("close not implemented");
+          }
+
+          @Override
+          public int read(ByteBuffer buffer) throws IOException {
+            throw new IOException("read throws exception");
```
singhravidutt, line 166: nit: capture message in a variable and compare the cause while asserting.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -140,6 +142,62 @@ public void eagerInitialization_fails_withInvalidCredentialsConfiguration() {
...
+          @Override
+          public SeekableByteChannel truncate(long l) throws IOException {
+            throw new IOException("truncate not implemented");
+          }
+        };
+
+    GoogleHadoopFSInputStream inputStream =
+        new GoogleHadoopFSInputStream(myGhfs, bucketPath, channel, null);
+    assertThrows(
+        "read throws exception", IOException.class, () -> inputStream.read(new byte[1], 0, 1));
```
singhravidutt, line 198: should we verify single byte read api as well?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -84,7 +87,7 @@ static GoogleHadoopFSInputStream create(
     return new GoogleHadoopFSInputStream(ghfs, fileInfo.getPath(), channel, statistics);
   }
 
-  private GoogleHadoopFSInputStream(
```
arunkumarchacko, line 87: instead of making this public, why not use the create method and mock GoogleHadoopFileSystem?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -38,6 +39,8 @@
 import org.apache.hadoop.fs.statistics.IOStatistics;
 import org.apache.hadoop.fs.statistics.IOStatisticsSource;
 
+@VisibleForTesting
```
singhravidutt, line 42: I don't think we should annotate the class but the constructor along with comment (as in your previous implementation).

singhravidutt (commented):
LGTM, some minor comments.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
singhravidutt: There were total of three options 

1. mocking the transport layer.
2. mocking gcsio layer i.e GoogleCloudStorageFileSystem layer
3. mocking the constructor of GHFSInputStream and provide the mocked channel ( implemented in here).

Option 1 will require separate mocking for http and grpc layers. Given it's too much work and as unit which we are testing is just the exception thrown in channel. It not ideal to go via this route.

Option 2, just mocking GCSFileSystem will not be enough we have to also mock the underneath channel returned form GCSFileSystem layer which is an extra effort, but still doable.

Option 3, mocking the channel passed in GHFSInputStream. It is straight forward and seems the best option as for writing unit test in gcs layer we should be treating gcsio layer as blackbox. In doing so we have to expose the constructor of `GoogleHadoopFSInputStream`. It doesn't seem too much concerning to me as the only place we create the GHFSInputStream is in GHFS.create call.

Having said that if this is a concern we can move to Option 2.

guljain: Updated to Option 2

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
guljain: added test

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
guljain: removed

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
guljain: PTAL

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -266,4 +321,68 @@ public void testInitializeCompatibleWithHadoopCredentialProvider() {}
 
   @Override
   public void testRenameHnBucket() {}
+
+  private class InMemoryGoogleCloudStorageForRead extends InMemoryGoogleCloudStorage {
```
singhravidutt, line 325: nit: InMemoryGoogleCloudStorageForRead -> CustomInMemoryCloudStorage. 

Also, provide the description about what it is suppose to do.

singhravidutt (approved):

arunkumarchacko (approved):

PR #1159, opened 2024-05-22
Add readVectored api support in gcsio

singhravidutt:
Added an interface `GoogleCloudStorageReadableByteChannel` whcih extends `SeekableByteChannel` and provides a way to add support for other functions which are not part of nio channel's SeekableByteChannel i.e. readVectored.

Also, updated GoogleCloudStortage's `open` api's return type,`SeekableByteChannel` -> `GoogleCloudStorageReadableByteChannel`.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
@@ -83,7 +83,6 @@ import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.grpc.ClientInterceptor;
 import java.io.IOException;
-import java.nio.channels.SeekableByteChannel;
 import java.nio.channels.WritableByteChannel;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Paths;
@@ -1021,14 +1020,14 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
   }
 
   @Override
-  public SeekableByteChannel open(
+  public GoogleCloudStorageReadableByteChannel open(
       StorageResourceId resourceId, GoogleCloudStorageReadOptions readOptions) throws IOException {
     logger.atFiner().log("open(%s, %s)", resourceId, readOptions);
     return open(resourceId, /* itemInfo= */ null, readOptions);
   }
 
   @Override
-  public SeekableByteChannel open(
+  public GoogleCloudStorageReadableByteChannel open(
       GoogleCloudStorageItemInfo itemInfo, GoogleCloudStorageReadOptions readOptions)
       throws IOException {
     logger.atFiner().log("open(%s, %s)", itemInfo, readOptions);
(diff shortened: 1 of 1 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -196,13 +197,13 @@ public synchronized void createEmptyObjects(
   }
 
   @Override
-  public SeekableByteChannel open(
+  public GoogleCloudStorageReadableByteChannel open(
```
abmodi, line 200: Since we are changing signatures of public interface, how are we planning to release it?

singhravidutt: `GoogleCloudStorageReadableByteChannel` still extends `SeekableByteChannel` and it will not be a breaking change for existing users. They although will not be able to utilize the new APIs unless they start consuming `GoogleCloudStorageReadableByteChannel`.

singhravidutt: only other alternate is to add new API something like `openWithVectored`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadableByteChannel.java
```diff
@@ -0,0 +1,60 @@
...
+   * request and throw IOException
+   *
+   * <p>While a readVectored() operation is in progress, normal read api calls may block.
+   *
+   * @param ranges the byte ranges to read
+   * @param allocate the function to allocate ByteBuffer
+   * @throws IOException any IOE.
+   * @throws IllegalArgumentException if any of ranges are invalid, or they overlap.
+   */
+  default void readVectored(
```
arunkumarchacko, line 55: Why is this part of the Channel interface and not the stream interface? Why not have GoogleHadoopFsInputStream implemention PositionedReadable interface and have the vectoredIO implementation there?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -631,7 +630,7 @@ public SeekableByteChannel open(
    * behavior.
    */
   @Override
-  public SeekableByteChannel open(
+  public GoogleCloudStorageReadableByteChannel open(
```
arunkumarchacko, line 633: What is the problem if we still return SeekableBytChannel?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadableByteChannel.java
singhravidutt: that is one way to go about it but there will be performance implication and also have a strong depedency on how gcsInputStream i.e. ReadChannel is implemented.

> Why not have GoogleHadoopFsInputStream implemention PositionedReadable interface and have the vectoredIO implementation there?

Currently we hold a lot of optimization logic in gcsio layer such as, itemInfo call on open, footer caching etc, which all worked well for positioned read but will be an overhead for vectoredRead and also cause performance issue. One e.g. of it would be that we make an itemInfo call in gsio layer for `open`. If we try implementing vectoredRead in GoogleHadoopFsInputStream we have to make atleast one more itemInfo call.

> Why not have GoogleHadoopFsInputStream implemention PositionedReadable interface

It already does. GoogleHadoopFsInputStream extends FSInputStream which implements PositionedReadable.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
singhravidutt: then returned channel will not have new api we want to add i.e. readVectored

PR #1168, opened 2024-06-05
Refactoring UT for HN Bucket

guljain:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -163,7 +163,7 @@ private void addFolderResourceInBlockingQueue(FolderInfo folderResource) {
    *
    * @param folderResource of the folder that is now deleted
    */
-  private void successfullDeletionOfFolderResource(FolderInfo folderResource) {
+  void successfullDeletionOfFolderResource(FolderInfo folderResource) {
```
arunkumarchacko, line 166: If we are increasing the visibility for testing, why not make this protected instead of private?

guljain: Changed access modifiers

arunkumarchacko (approved):

PR #1175, opened 2024-06-12
Support zero-copy readChannel

singhravidutt:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -252,9 +252,15 @@ public int readContent(ByteBuffer dst) throws IOException {
           }
           int bytesRead = byteChannel.read(dst);
 
+          /*
+          As we are using the zero copy implementation of byteChannel, it can return even zero bytes,
+          while reading,
+          we should not treat it as an error scenario anymore.
+          */
           if (bytesRead == 0) {
```
arunkumarchacko, line 260: can this cause infinite loop?

arunkumarchacko (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
singhravidutt: Well, underneath channel will expire after a timeout.

PR #1182, opened 2024-06-17
Adding total time metrics

guljain:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStatistic.java
```diff
@@ -88,7 +89,8 @@ public enum GhfsStatistic {
       "Count of exceptions raised during input stream reads",
       TYPE_COUNTER),
   STREAM_READ_OPERATIONS("stream_read_operations", "Calls of read()", TYPE_DURATION),
-
+  STREAM_READ_OPERATIONS_TIME(
```
arunkumarchacko, line 92: Instead make the current metric to TYPE_DURATION_TOTAL?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -465,6 +502,10 @@ private String getMeanKey(String symbol) {
     return symbol + "_mean";
   }
 
+  private String getTimeKey(String symbol) {
+    return symbol + "_time";
```
arunkumarchacko, line 506: should we call this duration?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsStorageStatistics.java
```diff
@@ -86,6 +88,13 @@ public GhfsStorageStatistics() {
         maximums.put(getMaxKey(symbol), new AtomicLong(0));
         means.put(getMeanKey(symbol), new MeanStatistic());
       }
+
+      if (opType.getType() == StatisticTypeEnum.TYPE_DURATION_TOTAL) {
```
arunkumarchacko, line 92: nit: may be the three lines can be move to the previous if block and get rid of the the three line code duplication.

arunkumarchacko (approved):

PR #1207, opened 2024-07-07
Track latency, retry and backoff time of GCS JSON API requests

arunkumarchacko:


util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -80,14 +80,17 @@ public void initialize(HttpRequest request) throws IOException {
       credentials.initialize(request);
     }
 
+    RequestTracker tracker = getRequestTracker(request);
```
singhravidutt, line 83: A Retired request will get initialized too and will initialize a separate tracker for it, correct?  

How we are able to maintain the retryCount in RequestTracker then?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -151,30 +158,38 @@ private static class UnsuccessfulResponseHandler implements HttpUnsuccessfulResp
...
         // If credentials decides it can handle it, the return code or message indicated something
         // specific to authentication, and no backoff is desired.
         return true;
       }
 
+      long backOffStartTime = System.currentTimeMillis();
       if (delegate.handleResponse(request, response, supportsRetry)) {
+        tracker.trackBackOffCompleted(backOffStartTime, request);
         // Otherwise, we defer to the judgement of our internal backoff handler.
+        tracker.trackRetryStarted();
```
singhravidutt, line 187: I would recommend overriding `HttpBackOffUnsuccessfulResponseHandler` handleResponse and adding this logic there instead of polluting `UnsuccessfulResponseHandler`'s `handleResponse` which has nothing to do with the retries.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -151,30 +158,38 @@ private static class UnsuccessfulResponseHandler implements HttpUnsuccessfulResp
...
       logResponseCode(request, response);
+      tracker.trackUnsuccessfulResponseHandler(response);
 
       if (credentials != null && credentials.handleResponse(request, response, supportsRetry)) {
         // If credentials decides it can handle it, the return code or message indicated something
         // specific to authentication, and no backoff is desired.
         return true;
       }
 
+      long backOffStartTime = System.currentTimeMillis();
```
singhravidutt, line 183: What about using using [getCurrentIntervalMillis](https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.util.ExponentialBackOff#com_google_api_client_util_ExponentialBackOff_getCurrentIntervalMillis__)?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -236,7 +253,20 @@ public boolean handleIOException(HttpRequest httpRequest, boolean supportsRetry)
       // We sadly don't get anything helpful to see if this is something we want to log.
       // As a result we'll turn down the logging level to debug.
       logger.atFine().log("Encountered an IOException when accessing URL %s", httpRequest.getUrl());
-      return delegate.handleIOException(httpRequest, supportsRetry);
+      tracker.trackIOException(httpRequest);
```
singhravidutt, line 256: Similar comments as of `UnsuccessfulResponseHandler` handleResponse?

util/src/main/java/com/google/cloud/hadoop/util/RequestTracker.java
```diff
@@ -0,0 +1,121 @@
...
+    // will also get called.
+    if (stopWatch.isRunning()) {
+      postToEventQueue(GcsJsonApiEvent.getResponseEvent(response, stopWatch.elapsed().toMillis()));
+      stopTracking(response.getRequest());
+    }
+
+    if (retryCount != 0) {
+      // Change to minute
+      logger.atInfo().atMostEvery(10, TimeUnit.SECONDS).log(
+          "Operation completed after retries with code '%s'. %s", response.getStatusCode(), this);
```
singhravidutt, line 52: mind capturing the retry count here and total time spend in the request?

util/src/test/java/com/google/cloud/hadoop/util/GcsJsonApiEventTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class GcsJsonApiEventTest {
+  private static final String URL = "http://fake-url.com";
+
+  @Test
+  public void testGetResponseEvent() throws IOException {
```
singhravidutt, line 36: what about adding test for all event types?

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
```diff
@@ -126,6 +145,62 @@ private void errorCodeResponse_succeedsAfterRetries(int statusCode) throws Excep
         .contains(InvocationIdInterceptor.GCCL_INVOCATION_ID_PREFIX);
     assertThat(res).isNotNull();
     assertThat(res.getStatusCode()).isEqualTo(HttpStatusCodes.STATUS_CODE_OK);
+
+    System.out.println(requestTracker.getEvents());
```
singhravidutt, line 149: cleanup?

util/src/main/java/com/google/cloud/hadoop/util/RequestTracker.java
```diff
@@ -0,0 +1,121 @@
...
+      logger.atInfo().atMostEvery(10, TimeUnit.SECONDS).log(
+          "Retry skipped after %s retries. context=%s", retryCount, this);
+    }
+  }
+
+  protected void postToEventQueue(GcsJsonApiEvent event) {
+    GoogleCloudStorageEventBus.postGcsJsonApiEvent(event);
+  }
+
+  protected RequestTracker init(HttpRequest request) {
```
singhravidutt, line 89: why can't we make this part of constructor?

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
```diff
@@ -166,4 +248,22 @@ private static RetryHttpInitializer createRetryHttpInitializer(Credentials crede
...
+
+  private class TestRetryHttpInitializer extends RetryHttpInitializer {
+    private boolean init;
+
+    public TestRetryHttpInitializer(Credentials credentials, RetryHttpInitializerOptions build) {
+      super(credentials, build);
+    }
+
+    @Override
+    protected RequestTracker getRequestTracker(HttpRequest request) {
```
singhravidutt, line 260: Correct me If I am wrong, With this override what we are able to use the same RequestTracker instance against all the gcs requests made in test case i.e. getMetadata, getObject, listObject etc, all the retires of the request are also captured via same tracer but in actual production scenario it will not be the case, every request will have a separate tracker.

If yes, will it even be a valid test then?

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
```diff
@@ -166,4 +248,22 @@ private static RetryHttpInitializer createRetryHttpInitializer(Credentials crede
             .setReadTimeout(Duration.ofSeconds(5))
             .build());
   }
+
+  private class TestRetryHttpInitializer extends RetryHttpInitializer {
+    private boolean init;
```
singhravidutt, line 253: initialized?

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
```diff
@@ -151,30 +158,38 @@ private static class UnsuccessfulResponseHandler implements HttpUnsuccessfulResp
...
 
       if (credentials != null && credentials.handleResponse(request, response, supportsRetry)) {
         // If credentials decides it can handle it, the return code or message indicated something
         // specific to authentication, and no backoff is desired.
         return true;
       }
 
+      long backOffStartTime = System.currentTimeMillis();
       if (delegate.handleResponse(request, response, supportsRetry)) {
+        tracker.trackBackOffCompleted(backOffStartTime, request);
```
singhravidutt, line 185: why not just pass the timeElapsed vs startTime?

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
singhravidutt: Yeah, same tracker will be used across all retries of same request, but if there are multiple requests, say two different request to access object metadata of object1 and object2. `TestRetryHttpInitializer` configured here will not be able to provide different tracker but provide the same tracker which will yield weird outcomes at verification step.

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Override
+  protected void postToEventQueue(GcsJsonApiEvent event) {
+    this.events.add(event);
+  }
+
+  void verifyEvents(List<ExpectedEventDetails> expectedEvents) {
+    List<GcsJsonApiEvent> actualEvents = getEvents();
+    assertThat(actualEvents.size()).isEqualTo(expectedEvents.size());
+    assertThat(actualEvents.size()).isAtLeast(1);
```
singhravidutt, line 41: add a comment about why?

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
```diff
@@ -0,0 +1,123 @@
...
+      if (eventType == GcsJsonApiEvent.EventType.BACKOFF) {
+        verifyNotEmpty(actual, GcsJsonApiEvent.BACKOFFTIME);
+      }
+
+      for (String key : expected.properties.keySet()) {
+        if (key.equals(GcsJsonApiEvent.BACKOFFTIME)) {
+          long backOffTime = (long) actual.getProperty(key);
+          int expectedBackoffTime = (int) expected.properties.get(GcsJsonApiEvent.BACKOFFTIME);
+          assertThat(backOffTime).isAtLeast(expectedBackoffTime);
+          assertThat(backOffTime).isLessThan(expectedBackoffTime + 10);
```
singhravidutt, line 68: why it varies as range? 
How we came up with `10`?
will it cause flakiness? 
what if some adds a test where many more retires are done, this verification will still be valid?

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
```diff
@@ -0,0 +1,123 @@
...
+
+      if (eventType == GcsJsonApiEvent.EventType.RESPONSE) {
+        verifyNotEmpty(actual, GcsJsonApiEvent.DURATION);
+      }
+
+      if (eventType == GcsJsonApiEvent.EventType.BACKOFF) {
+        verifyNotEmpty(actual, GcsJsonApiEvent.BACKOFFTIME);
+      }
+
+      for (String key : expected.properties.keySet()) {
```
singhravidutt, line 63: what about validating the size of properties?

util/src/main/java/com/google/cloud/hadoop/util/RequestTracker.java
```diff
@@ -0,0 +1,121 @@
...
+
+    return this;
+  }
+
+  private void stopTracking(HttpRequest httpRequest) {
+    if (stopWatch.isRunning()) {
+      stopWatch.stop();
+
+      if (stopWatch.elapsed().toMillis() > LOGGING_THRESHOLD) {
+        logger.atInfo().atMostEvery(10, TimeUnit.SECONDS).log(
```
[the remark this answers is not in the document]
singhravidutt: How it is actionable? We are not logging invocationIds here correct?

util/src/main/java/com/google/cloud/hadoop/util/GcsJsonApiEvent.java
```diff
@@ -0,0 +1,103 @@
...
+    return method;
+  }
+
+  Object getProperty(String key) {
+    return properties == null ? null : properties.get(key);
+  }
+
+  private GcsJsonApiEvent(HttpRequest request, EventType eventType, int capacity) {
+    this(request, eventType);
+    this.properties = new HashMap<>(capacity);
```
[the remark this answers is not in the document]
arunkumarchacko: If you provide the capacity, it will set the capacity to the next power of 2. It will not resize. What you think it will resize?

util/src/main/java/com/google/cloud/hadoop/util/RequestTracker.java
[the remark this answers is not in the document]
arunkumarchacko: Logging InvocationId will be coming in the next change.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
arunkumarchacko: getCurrentIntervalMillis is not the backOffTime, it is the next backoff interval.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
arunkumarchacko: ok. I get what you are saying. Agree, that would have worked as well. The current implementation is fine too.

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
arunkumarchacko: will sync offline.

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
arunkumarchacko: This is not required. Removed. Thanks.

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
```diff
@@ -0,0 +1,123 @@
...
+      assertThat(actual.getContext().toString()).isEqualTo(expected.context.toString());
+
+      GcsJsonApiEvent.EventType eventType = actual.getEventType();
+      if (eventType == GcsJsonApiEvent.EventType.RESPONSE) {
+        verifyNotEmpty(actual, GcsJsonApiEvent.DURATION);
+      }
+
+      if (eventType == GcsJsonApiEvent.EventType.RESPONSE) {
+        verifyNotEmpty(actual, GcsJsonApiEvent.DURATION);
+      }
```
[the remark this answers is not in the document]
arunkumarchacko: Done

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
arunkumarchacko: The property map is encapsulated in the class. Do not want to expose it just for testing.

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
arunkumarchacko: These are unit tests, not actual backoff is happening. The buffer is the potential overhead. I do not expect it to be more than 10 seconds.

util/src/main/java/com/google/cloud/hadoop/util/GcsJsonApiEvent.java
```diff
@@ -0,0 +1,103 @@
...
+
+  static GcsJsonApiEvent getBackoffEvent(
+      HttpRequest request, @Nonnegative long backOffTime, @Nonnegative int retryCount) {
+    return new GcsJsonApiEvent(request, EventType.BACKOFF, 2)
+        .set(BACKOFFTIME, backOffTime)
+        .set(RETRYCOUNT, retryCount);
+  }
+
+  @VisibleForTesting
+  GcsJsonApiEvent(@Nonnull HttpRequest request, EventType eventType) {
```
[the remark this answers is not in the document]
arunkumarchacko: gRPC support will come in a separate change.

util/src/test/java/com/google/cloud/hadoop/util/TestRequestTracker.java
singhravidutt: > These are unit tests, not actual backoff is happening.

Well, that is not always true, is it? One can mock the transport in unit tests but still use RetryHttpIniitalizer which comes with ExponentialBackOff strategy and hence the actual backoff will will happen.

util/src/main/java/com/google/cloud/hadoop/util/RetryHttpInitializer.java
singhravidutt: if you query it before the actual backOff being requested it will provide you the value you are looking for.

util/src/test/java/com/google/cloud/hadoop/util/RetryHttpInitializerTest.java
```diff
@@ -166,4 +246,23 @@ private static RetryHttpInitializer createRetryHttpInitializer(Credentials crede
             .setReadTimeout(Duration.ofSeconds(5))
             .build());
   }
+
+  // Helper class which help provide a custom test implementation of RequestTracker
```
singhravidutt, line 250: I would recommend adding a note on limitation of `TestRetryHttpInitializer`, which is, "it can only used against test with only one request because RequestTracker is initialized along with HttpInitializer".

arunkumarchacko: This is an inner class and is an implementation detail. I think it is obvious from the code that this applies for one request per TestRetryHttpInitializer.

singhravidutt (approved):

PR #1209, opened 2024-07-08
Fix bytesRead reported

singhravidutt:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -207,13 +207,9 @@ public synchronized int read(long position, byte[] buf, int offset, int length)
       throws IOException {
     long startTimeNs = System.nanoTime();
 
+    // Note: don't update byteRead Metric here. As, it will be reported by super.read
```
arunkumarchacko, line 210: This override does not seem to be doing anything. Why we need this override?

singhravidutt: I agree, don't see any value add of this as of now. We don't have it in 3.0.

arunkumarchacko (approved):

PR #1220, opened 2024-07-09
control grpc write path via flag (#1188)

singhravidutt:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -497,6 +497,11 @@ public class GoogleHadoopFileSystemConfiguration {
       new HadoopConfigurationProperty<>(
           "fs.gs.client.type", GoogleCloudStorageFileSystemOptions.DEFAULT.getClientType());
 
+  /** Configuration key to configure client to use for GCS access. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_GRPC_WRITE_ENABLE =
```
arunkumarchacko, line 501: should we add this to the Configuration file?

singhravidutt: This is a cherry-pick PR. Any required changes need to go through separate PR.

arunkumarchacko (approved):

PR #1227, opened 2024-07-24
2.2.x Fixing bug : Avoid registering subscriber class multiple times

guljain:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemBase.java
```diff
@@ -361,7 +361,8 @@ public GoogleHadoopFileSystemBase() {
       storageStatistics = GhfsStorageStatistics.DUMMY_INSTANCE;
     }
 
-    GoogleCloudStorageEventBus.register(new GoogleCloudStorageEventSubscriber(storageStatistics));
+    GoogleCloudStorageEventBus.register(
+        GoogleCloudStorageEventSubscriber.getInstance(storageStatistics));
```
arunkumarchacko, line 365: why not use the previous logic? i.e. just use storageStatistics

arunkumarchacko (approved):

PR #1244, opened 2024-09-05
Add open() API with FileStatus

arunkumarchacko:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -589,6 +589,24 @@ public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOExceptio
         });
   }
 
+  public FSDataInputStream open(FileStatus status) throws IOException {
```
singhravidutt, line 592: How we are going to consume this given we don't have it in FIleSystem?

singhravidutt, line 592: Why FileStatus and not directly use GoogleHadoopFileStatus given it's a custom api?

arunkumarchacko: discussed offline

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -589,6 +589,24 @@ public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOExceptio
...
+    logger.atFine().log("openWithStatus(%s)", status);
+
+    GoogleHadoopFileStatus fileStatus = (GoogleHadoopFileStatus) status;
```
singhravidutt, line 595: Wouldn't it be casting issue?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
singhravidutt, line 592: Is using `open` intentional? What about api metric counters given we already have an API with `open`  and I do see we use dame statistics i.e. `GhfsStatistic.INVOCATION_OPEN`

arunkumarchacko: yes. I do not think adding a new metric for this is required. Too many metrics are not good.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -589,6 +589,24 @@ public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOExceptio
...
+    return trackDurationWithTracing(
+        instrumentation,
+        globalStorageStatistics,
+        GhfsStatistic.INVOCATION_OPEN,
+        status.getPath(),
+        this.traceFactory,
+        () -> {
+          checkOpen();
+          return new FSDataInputStream(
+              GoogleHadoopFSInputStream.create(this, fileStatus.getFileInfo(), statistics));
```
singhravidutt, line 606: IIUC, there is no restriction on the bucket of provided FileStatus to the bucket with which ghfs is initialized. Is it an intended behaviour? Isn't it deviating from all other APIs?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
singhravidutt: But it's a different api all together. Not capturing metric for it is fine, but we should definitely not corrupt the existing metric.

arunkumarchacko: This can be called from outside connector and it may not have GoogleHadoopFileStatus dependency.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
arunkumarchacko: Caller will have to handle the exception.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
arunkumarchacko: Discussed offline

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
arunkumarchacko: Thank you for the suggestion. Updated.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
singhravidutt: I would disagree. This is a custom API we are creating if someone is calling it they have to have gcs-connector library to use it and hence they would have `GoogleHadoopFileStatus`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
singhravidutt: It would be better to validate and  throw IllegalArgument instead of throwing type cast exception from API?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
singhravidutt: Discussed offline, it's fine to use `FileStatus` given the way we are planning to use it. `GoogleHadoopFileStatus` will make it more complex to use.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
arunkumarchacko: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -589,6 +589,33 @@ public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOExceptio
         });
   }
 
+  public FSDataInputStream open(FileStatus status) throws IOException {
+    logger.atFine().log("openWithStatus(%s)", status);
+
+    if (!GoogleHadoopFileStatus.class.isAssignableFrom(status.getClass())) {
```
singhravidutt, line 595: `isInstanceOf` would have sufficed, no?

singhravidutt (approved):

PR #1250, opened 2024-09-13
Lock the random access if a forward or backward seek is detected

arunmahadevan:
The optimized `AUTO_RANDOM` mode currently requires a switch to `sequential` before it can switch to random.  The proposed change locks the access pattern to `random` as soon as a forward or backward seek is detected.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/FileAccessPatternManager.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/FileAccessPatternManager.java
@@ -65,12 +65,10 @@ class FileAccessPatternManager {
       return;
     }
     if (readOptions.getFadvise() == Fadvise.AUTO_RANDOM) {
-      if (randomAccess) {
+      if (isBackwardOrForwardSeekNotRequested()) {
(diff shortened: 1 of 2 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/FileAccessPatternManager.java
```diff
@@ -151,18 +149,21 @@ private boolean isRandomAccessPattern(long currentPosition) {
...
     return randomAccess
-        && !isBackwardSeekRequested
-        && !isForwardSeekRequested
+        && isBackwardOrForwardSeekNotRequested()
         && consecutiveSequentialCount >= readOptions.getFadviseRequestTrackCount()
         && readOptions.getFadvise() == Fadvise.AUTO_RANDOM;
   }
 
   private boolean shouldDetectRandomAccess() {
-    return !randomAccess
```
singhravidutt, line 163: If already in randomAccess mode there is no need to detect random access.

I do agree that the current logic will not update backwardSeek and forwardSeek if occurredwhen random access was set but I would suggest updating those flags well within fn:`updateAccessPattern` .

PR #1253, opened 2024-09-17
Add getFileStatusWithHint API

arunkumarchacko:


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTestBase.java
```diff
@@ -539,4 +542,94 @@ public void provideCoverageForUnmodifiedMethods() throws IOException {
...
+  }
+
+  private void waitForUpdation(GoogleCloudStorageStatistics stat, int value)
+      throws InterruptedException {
+    for (int i = 0; i < 10; i++) {
+      if (getStatisticValue(stat) >= value) {
+        return;
+      }
+
+      Thread.sleep(1);
```
singhravidutt, line 584: this will still result in test flakiness?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -985,18 +996,26 @@ public FileInfo getFileInfoObject(URI path) throws IOException {
...
+      throws IOException {
     if (resourceId.isRoot() || resourceId.isBucket()) {
       return gcs.getItemInfo(resourceId);
     }
-    StorageResourceId dirId = resourceId.toDirectoryId();
 
+    StorageResourceId dirId = resourceId.toDirectoryId();
     Future<List<GoogleCloudStorageItemInfo>> listDirFuture =
-        (options.isStatusParallelEnabled() ? cachedExecutor : lazyExecutor)
+        (options.isStatusParallelEnabled() && pathTypeHint != PathTypeHint.FILE
```
singhravidutt, line 1016: Shouldn't we log the cases where provided hint is wrong?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTestBase.java
arunkumarchacko: I assume the other thread should catch up in 10 ms. If not, it is safe to increase the busy wait loop count.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
arunkumarchacko: Done.

singhravidutt (approved):

PR #1275, opened 2024-11-14
upgrade java-storage version

singhravidutt:


gcs/CHANGES.md
```diff
@@ -1,8 +1,8 @@
 # Release Notes
 
 ## Next
-1. Upgrade google-cloud-storage to 2.43.1
-1. Upgrade grpc to 1.68.0
+1. Upgrade google-cloud-storage to 2.44.1
+1. Upgrade grpc to 1.67.1
```
arunkumarchacko, line 5: you mean downgrade?

arunkumarchacko (approved):

gcs/CHANGES.md
singhravidutt: 1.68.0 was a mistake. They accidentally skipped 1.67.0 as well a s introduced a bug. They released 1.67.1 after that and fixed the bug as well. So, it may look like a downgrade but it's a upgrade.

PR #1282, opened 2024-11-25
Remove gRPC server address

arunkumarchacko:


Deependra-Patel (commented):
Wouldn't it be a breaking change if some customer is already using this property?

Deependra-Patel:
Let's add context why are we removing it

arunkumarchacko:
> Wouldn't it be a breaking change if some customer is already using this property?

This will not break any customer since this config is not getting used.
Also, this is not a documented configuration

Deependra-Patel (approved):

PR #1294, opened 2025-01-10
Make {Read,Write}Options-parsing methods public

clairemcginty:
👋 I'm an Apache Beam user who recently contributed [this](https://github.com/apache/beam/pull/32769) PR to allow users to pass in custom StorageReadOptions to a Beam Pipeline. However, I don't have an easy way to construct these options from my `Configuration` objects. I'm proposing to make this method public in gcs-connector so I can easy transform a `Configuration` into a `GoogleCloudStorageReadOptions` instance.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -653,7 +653,7 @@ private static GoogleCloudStorageReadOptions getReadChannelOptions(Configuration
         .build();
   }
 
-  private static AsyncWriteChannelOptions getWriteChannelOptions(Configuration config) {
+  public static AsyncWriteChannelOptions getWriteChannelOptions(Configuration config) {
```
clairemcginty, line 656: though this method isn't strictly needed based on my Beam PR, I thought it made sense to make it public as well to have symmetry with the `getReadChannelOptions` method.

clairemcginty:
hi, I just wanted to bump this PR! @arunkumarchacko, would you have time to take a look? thanks in advance!

clairemcginty:
cc @animesh-g - can you take a look at this PR? 🙏

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -347,6 +349,82 @@ public void grpcConfiguration() {
         .isEqualTo(Duration.ofMillis(20));
   }
 
+  @Test
```
clairemcginty, line 352: the tests use all non-default options for each property, so we're not inadvertently testing the fallback behavior 👍

clairemcginty:
> Apologies for replying late. I am fine with making these public but this PR needs two changes,
> 
> * Since these methods are now a public API, they should have a javadoc comment.
> * Explicit unit tests are required to validate the behavior of these two methods.

thanks! just updated the PR with the requested changes.

PR #1295, opened 2025-01-16
Close should be resilient to being called again

mswintermeyer:
Multiple other fields in close are resilient if close is called multiple times. Make the final field resilient as well.


mswintermeyer:
Hey @Deependra-Patel I see you reviewing some other PRs. What's the process like to get PRs reviewed?

mswintermeyer:
Fixed with https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1536 instead

PR #1305, opened 2025-02-20
Reduce logging frequency

arunkumarchacko:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsGlobalStorageStatistics.java
```diff
@@ -86,7 +86,7 @@ public class GhfsGlobalStorageStatistics extends StorageStatistics {
   private final Map<String, AtomicLong> maximums = new HashMap<>();
   private final Map<String, MeanStatistic> means = new HashMap<>();
   private final Map<String, AtomicDouble> total = new HashMap<>();
-  private final Stopwatch stopwatch = Stopwatch.createStarted();
+  private static final Stopwatch uptimeStopWatch = Stopwatch.createStarted();
```
Deependra-Patel, line 89: Nit: UPTIME_STOPWATCH? since it is static now

abmodi (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GhfsGlobalStorageStatistics.java
arunkumarchacko: Done

Deependra-Patel (approved):

PR #1312, opened 2025-03-18
Upgrade GCS Java SDK version

dheerajsngh:
Upgrade GCS Java SDK version from 2.43 to 2.49.
This is required for Move api which is available only after 2.48

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,7 @@
     <google.auth.version>1.22.0</google.auth.version>
     <google.auto-value.version>1.10.4</google.auto-value.version>
     <google.cloud-core.verion>2.44.1</google.cloud-core.verion>
-    <google.cloud-storage.bom.version>2.43.1</google.cloud-storage.bom.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -89,7 +89,7 @@
     <google.auth.version>1.22.0</google.auth.version>
     <google.auto-value.version>1.10.4</google.auto-value.version>
     <google.cloud-core.verion>2.44.1</google.cloud-core.verion>
-    <google.cloud-storage.bom.version>2.43.1</google.cloud-storage.bom.version>
+    <google.cloud-storage.bom.version>2.49.1</google.cloud-storage.bom.version>
```
animesh-g, line 92: There us no such release as 2.49.1 in SDK. We should use 2.49.0 . https://screenshot.googleplex.com/3nFYtqJxrmWEoMN

animesh-g (changes requested):
Please make sure it is building properly in local setup.

PR #1317, opened 2025-03-28
Release 3.1.0

arunkumarchacko:


README.md
```diff
@@ -9,9 +9,7 @@ open-source software and Google Cloud Platform.
 
 ## Google Cloud Storage connector for Apache Hadoop (HCFS)
 
-[![Maven Central](https://img.shields.io/maven-central/v/com.google.cloud.bigdataoss/gcs-connector/hadoop1.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:com.google.cloud.bigdataoss%20AND%20a:gcs-connector%20AND%20v:hadoop1-*)
```
singhravidutt, line 12: Historically we use to keep all the previous versions as well.

Also, should it be a regex pattern like 3.1.* to alway fetch the latest instead of pointing a static 3.1.0?

arunkumarchacko: Regex is not working. If you click on the link it is not pointing to the correct jar. 

I do not think we should keep the old ones. Even if we do we should keep that previous. However the problem with it is that it might point to a stale release since if that branch is updated, this branch will not get reflected. If regex was working, we could have done it. But for some reason regex is not working.

PR #1324, opened 2025-04-07
Staging uber

ShreyasSinha:
Adding conscrypt jar to the shaded connector jar for Uber branch.

### DIFF
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -224,6 +224,14 @@
               <filters>
                 <!-- minimizeJar breaks all the dynamic load hacks; add them explicitly -->
                 <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>META-INF/*.SF</exclude>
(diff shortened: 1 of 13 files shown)


gcs/pom.xml
```diff
@@ -223,6 +223,14 @@
               </transformers>
               <filters>
                 <!-- minimizeJar breaks all the dynamic load hacks; add them explicitly -->
+                <filter>
```
ShreyasSinha, line 226: Give conscrypt is a signed jar, it was giving a "Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes" when trying to create an uber shaded jar.
Thus removing all SF, DSA etc file to create unsigned jar. Refer: https://stackoverflow.com/questions/999489/invalid-signature-file-when-attempting-to-run-a-jar/6743609#6743609

BenWhitehead (approved):

singhravidutt (approved):

PR #1334, opened 2025-04-26
[Do not review] Using Apiary client for isHnBucket checks

animesh-g:
getStorageLayout call from Apiary client results in faster responses than StorageControlClient so using the same. Here are the results from our experiments,

Using Apiary call,
Before initializing connection - 1202 ms 
After initializing connection - 66 ms

Using Storage control client,
Without initializing connection - 2969 ms 
After initializing connection - 83 ms

There is around **20%** improvement in post-initialization calls and **60%** improvement in the pre-initialization calls. This will also helps us reducing the tax for non-HNS users when we enable the HNS flag by default in connector. One can find code changes for benchmarking these APIs [here](https://github.com/animesh-g/hadoop-connectors/pull/4/commits/5db42b44749244c1af155e5d885334e3570b1423).

### DIFF
--- a/gcs/CHANGES.md
+++ b/gcs/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+1. Add getFileStatusWithHint() API
+
 1. Upgrade google-cloud-storage to 2.43.1
 
 1. Upgrade grpc to 1.68.0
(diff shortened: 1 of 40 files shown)


ShreyasSinha (approved):

animesh-g:
Closing this PR as other one is approved and merged.

PR #1337, opened 2025-04-29
Create performance cache in all flows when `fs.gs.performance.cache.enable` flag is set

cpriti-os:
Creates performance cache in all constructors of GoogleCloudStorageFileSystem and modifieds UT to test the code.

Did not add for the third constructor since it uses an already configured GoogleCloudStorage.


cpriti-os:
Not required.

PR #1349, opened 2025-05-12
Using Apiary client for isHnBucket checks 

animesh-g:
getStorageLayout call from Apiary client results in faster responses than StorageControlClient so using the same. Here are the results from our experiments,

Using Apiary call,
Before initializing connection - 1202 ms
After initializing connection - 66 ms

Using Storage control client,
Without initializing connection - 2969 ms
After initializing connection - 83 ms

There is around 20% improvement in post-initialization calls and 60% improvement in the pre-initialization calls. This will also helps us reducing the tax for non-HNS users when we enable the HNS flag by default in connector. One can find code changes for benchmarking these APIs [here](https://github.com/animesh-g/hadoop-connectors/pull/4/commits/5db42b44749244c1af155e5d885334e3570b1423).

Older PR from a fork - https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1334

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
@@ -46,6 +46,7 @@ import com.google.api.gax.core.FixedCredentialsProvider;
 import com.google.api.services.storage.Storage;
 import com.google.api.services.storage.StorageRequest;
 import com.google.api.services.storage.model.Bucket;
+import com.google.api.services.storage.model.BucketStorageLayout;
 import com.google.api.services.storage.model.Buckets;
 import com.google.api.services.storage.model.ComposeRequest;
 import com.google.api.services.storage.model.Objects;
@@ -1824,7 +1825,7 @@ public class GoogleCloudStorageImpl implements GoogleCloudStorage {
     checkNotNull(listedFolder, "Must provide a non-null container for listedFolder.");
 
     ListFoldersPagedResponse listFolderRespose =
-        storageControlClient.listFolders(listFoldersRequest);
+        lazyGetStorageControlClient().listFolders(listFoldersRequest);
     try (ITraceOperation op = TraceOperation.addToExistingTrace("gcs.folders.list")) {
       Iterator<Folder> itemsIterator = listFolderRespose.getPage().getValues().iterator();
       while (itemsIterator.hasNext()) {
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2461,7 +2453,7 @@ public void renameHnFolder(URI src, URI dst) throws IOException {
 
     try (ITraceOperation to = TraceOperation.addToExistingTrace("renameHnFolder")) {
       logger.atFine().log("Renaming HN folder (%s -> %s)", src, dst);
-      this.storageControlClient.renameFolderOperationCallable().call(request);
+      lazyGetStorageControlClient().renameFolderOperationCallable().call(request);
```
singhravidutt, line 2456: what about this API? have we benchmarked it as well?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2414,20 +2415,11 @@ public boolean isHnBucket(URI src) throws IOException {
       return isEnabled;
     }
 
-    String prefix = src.getPath().substring(1);
```
singhravidutt, line 2417: How come there are no UnitTest failure caused by this change?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -2240,6 +2243,38 @@ public void testListBucketInfo() throws IOException {
...
+        new BucketStorageLayout()
+            .setHierarchicalNamespace(new HierarchicalNamespace().setEnabled(true));
+    MockHttpTransport transport = mockTransport(jsonDataResponse(layout));
+    GoogleCloudStorage gcs =
+        mockedGcsImpl(GCS_OPTIONS, transport, trackingRequestInitializerWithRetries);
+
+    URI bucketUri = new URI("gs://hns-enabled-bucket");
+    boolean result = gcs.isHnBucket(bucketUri);
+
+    assertThat(result).isTrue();
```
singhravidutt, line 2259: shall we also verifying if mockedTransport was is even called?

singhravidutt (commented):
LGTM, just wondering why flipping the client didn't cause any Unit test failure.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
animesh-g: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
animesh-g: There was no unit test written for this method. I have added some in this PR.

PR #1362, opened 2025-05-23
Prefix invocationId to logs to track connector request flow

cpriti-os:
Prefix invocationId to logs to track connector request flow across different components.
- Create a thread local variable invocationId using uuid which is inherited by child threads.
-  Set this thread local variable  at the start of connector API calls.
- Same invocation Id is used in trace logs and HTTP requests
- Added a custom log formatter to root logger which prefixes the invocationId to all appropriate logs.
- Unit tested and manually verified.


animesh-g:
I had a few questions on this PR on how it works in the multithreading environment with static invocation ID. Let's discuss offline on the same.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -760,13 +770,14 @@ private <B> B trackDurationWithTracing(
       ITraceFactory traceFactory,
       CallableRaisingIOE<B> operation)
       throws IOException {
-
+    InvocationIdContext.setInvocationId();
```
animesh-g, line 783: Q: Wouldn't it be redundant in above methods to setInvocationId directly as well as via trackDurationWithTracing? Why don't just add it here only?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -1107,6 +1130,7 @@ public Map<String, byte[]> getXAttrs(Path path) throws IOException {
 
   @Override
   public Map<String, byte[]> getXAttrs(Path path, List<String> names) throws IOException {
+    InvocationIdContext.getInvocationId();
```
animesh-g: Why are few functions calling setInvocationId and some getInvocationId ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -176,6 +179,8 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
   static final String GETFILESTATUS_FILETYPE_HINT = "fs.gs.getfilestatus.filetype.hint";
 
   static {
+    // Add formatter to root logger.
+    LoggingFormatter.addFormatter(Logger.getLogger(""));
```
animesh-g, line 186: Why is it in static block? Why can't we move this implementation to the LoggingFormatter class itself?

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Generates and sets a new unique invocation ID for the current thread. The ID is prefixed with
+   * {@code GCCL_INVOCATION_ID_PREFIX}.
+   */
+  public static void setInvocationId() {
+    String uuid = UUID.randomUUID().toString().substring(0, 8);
+    INVOCATION_ID.set(GCCL_INVOCATION_ID_PREFIX + uuid);
+  }
```
animesh-g, line 40: Should we check first if any invocation ID is already created in this thread before creating a new one?

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Checks if the given message is in JSON format.
+   *
+   * @param message the log message to check
+   * @return true if the message is in JSON format, false otherwise
+   */
+  private boolean isJson(String message) {
+    // Check if the message is a trace log.
+    return message != null && (message.startsWith("{") || message.startsWith("["));
```
animesh-g: Shouldn't we use JSON library to check if it is valid json or not?

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
```diff
@@ -0,0 +1,53 @@
...
+   * Formats the given log record by prefixing it with the invocation ID.
+   *
+   * @param record the log record to be formatted
+   * @return the formatted log message
+   */
+  @Override
+  public String format(LogRecord record) {
+    String invocationId = InvocationIdContext.getInvocationId();
+    String message = record.getMessage();
+    if (invocationId.isEmpty() || isJson(message)) {
```
animesh-g: Can you explain it a little? Why aren't we adding invocationId if message is in JSON format?

animesh-g (changes requested):

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
cpriti-os: The first and most important reason is: if we prefix invocation id to a json string, we'll lose the json format (Any existing tool depending on the format of certain logs being json would break). Secondly, JSON logs we are interested in are traces and they already have a field for invocation id.

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
cpriti-os: Creating again is by design, basically the idea is if we are using a single thread and we are calling connector APIs sequentially, all of them should have a different invocation id. If the API doesn't change, but multiple-threads are spawned, they would have the same invocationID by inheritance.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
cpriti-os: The configuration is in the static block to ensure that the logging formatter is set up as soon as the GoogleHadoopFileSystem class is loaded, before any instance is created or any static method is called. This guarantees that all log messages from this package use the custom formatter from the very beginning.
Wanted to keep the utility functions for formatting separate from configuring formatting, therefore did not place this in LoggingFormatter. Let me know if you recommend otherwise.

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
cpriti-os: Now using the gson package to validate JSON. Initially tried to avoid introducing a new dependency, but since gson is already included, this is definitely a better solution. Thanks!

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
cpriti-os: Agreed. Removed it from APIs that call this function, left it as is for APIs that do not call trackDurationWithTracing.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
cpriti-os: Thanks for catching this. All calls here should be setInvocationId, fixed it.

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
animesh-g: Sounds good!

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -65,8 +66,11 @@ public void intercept(HttpRequest request) throws IOException {
     final String signatureKey = "Signature="; // For V2 and V4 signedURLs
     final String builtURL = request.getUrl().build();
     if (!builtURL.contains(signatureKey)) {
-      UUID invocationId = UUID.randomUUID();
-      String invocationEntry = GCCL_INVOCATION_ID_PREFIX + invocationId;
+      // Reuse existing invocationId if present.
+      String invocationEntry = InvocationIdContext.getInvocationId();
+      if (invocationEntry.isEmpty()) {
+        invocationEntry = GCCL_INVOCATION_ID_PREFIX + UUID.randomUUID();
```
animesh-g: Shouldn't we set the variable again to InvocationIdContext? This value is written here but is never read again.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -581,6 +586,7 @@ public String getScheme() {
 
   @Override
   public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOException {
+    InvocationIdContext.setInvocationId();
```
animesh-g: We are calling setInvocationId() here as well as from trackDurationWithTracing. I think it needs to be removed here as discussed in the earlier comment.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -599,6 +605,7 @@ public FSDataInputStream open(Path hadoopPath, int bufferSize) throws IOExceptio
 
   /** This is an experimental API and can change without notice. */
   public FSDataInputStream open(FileStatus status) throws IOException {
+    InvocationIdContext.setInvocationId();
```
animesh-g: Same as above. Remove setInvocationId from here as it is already being set in trackDurationWithTracing .

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /** Clears the invocation ID for the current thread. */
+  public static void clear() {
```
animesh-g: Is it being called anywhere except tests? If it is only used for testing, prefer using `VisibleForTesting` instead of making it public.

animesh-g (changes requested):
Mostly looks good. Added some comments to handle minor issues.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
cpriti-os: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
cpriti-os: Done

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
cpriti-os: Done

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
cpriti-os: Done

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
```diff
@@ -39,7 +40,8 @@ public static void setInvocationId() {
   }
 
   /** Clears the invocation ID for the current thread. */
-  public static void clear() {
+  @VisibleForTesting
+  static void clear() {
```
animesh-g, line 44: nit: it is better to explicitly mark this private.

animesh-g (approved):

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
cpriti-os: This function needs to be of default visibility so that test classes in the same package can access it. Making it private would not allow tests to access it. Reopen this if you disagree.

util/src/test/java/com/google/cloud/hadoop/util/InvocationIdContextTest.java
```diff
@@ -0,0 +1,78 @@
...
+        invocationId.substring(InvocationIdInterceptor.GCCL_INVOCATION_ID_PREFIX.length());
+    assertEquals(8, uuidPart.length());
+  }
+
+  @Test
+  public void testClearInvocationId() {
+    // Set an invocation ID, clear it, and verify it is reset to an empty string
+    InvocationIdContext.setInvocationId();
+    InvocationIdContext.clear();
+    assertEquals("", InvocationIdContext.getInvocationId());
```
Dhriti07, line 46: Nit: Split into Arrange Act Assert

cpriti-os: Done.

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Formats the given log record by prefixing it with the invocation ID.
+   *
+   * @param record the log record to be formatted
+   * @return the formatted log message
+   */
+  @Override
+  public String format(LogRecord record) {
+    String invocationId = InvocationIdContext.getInvocationId();
+    String message = record.getMessage();
```
ajayky-os: Since we are not using `Optional` throughout the project, consider it optional suggestion:

Wrapping message and loggerName in Optional can simplify the code and avoid null checks.
Ex:
```
Optional<String> message = Optional.ofNullable(record.getMessage);
Optional<String> loggerName ...;

if (loggerName.map(name -> name.startsWith(GCS_C...)).orElse(false) 
     || message.map(this::praseJson)
     || invocationId.isEmpty()) {
...
}

go/java-practices/null#avoid

File: util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java [hunk 14]
```diff
@@ -0,0 +1,74 @@
...
+    // Prefixing JSON logs would break their structure; some JSON logs like traces already
+    // include the invocation ID.
+    if (loggerName == null
+        || !loggerName.startsWith(GCS_CONNECTOR_LOGGER_PREFIX)
+        || invocationId.isEmpty()
+        || isJson(message)) {
+      return String.format("%s%n", message);
+    }
+    // Prefix the invocation ID to the log message
+    return String.format("[%s]: %s%n", invocationId, record.getMessage());
```
ajayky-os: nit: reuse `message` variable.

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
```diff
@@ -0,0 +1,47 @@
...
+        @Override
+        protected String initialValue() {
+          return "";
+        }
+      };
+
+  /**
+   * Retrieves the current thread's invocation ID.
+   *
+   * @return the invocation ID for the current thread, or an empty string if not set
```
ajayky-os, line 27: Is there a usecase where we need it empty ? I see we are setting it if it is empty.

Can we move this logic inside the getter `getInvocationId()` itself ? Ie: if it is empty then set it and then return ? This way it will simplify the usage of getInvocationId() everywhere. `setInvocationId` can be made private then.

util/src/test/java/com/google/cloud/hadoop/util/InvocationIdContextTest.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test
+  public void testInitialValue() {
+    // Test that the initial value of INVOCATION_ID is an empty string
+    assertEquals("", InvocationIdContext.getInvocationId());
+  }
+
+  @Test
+  public void testSetInvocationId() {
```
ajayky-os: go/unit-testing-practices?polyglot=java#naming

util/src/test/java/com/google/cloud/hadoop/util/LoggingFormatterTest.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    // Clean up the ThreadLocal after each test
+    InvocationIdContext.clear();
+  }
+
+  @Test
+  public void testFormat_withInvocationId_andMatchingLogger() {
```
ajayky-os: We can remove the `test` prefix from the spec name.. it is understood from the `@Test` annotation that it is test spec.

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
cpriti-os: Thanks, done!

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
cpriti-os: Thanks for the suggestion, tried making the changes.

util/src/main/java/com/google/cloud/hadoop/util/InvocationIdContext.java
cpriti-os: Yes, we would want it to be empty for logs that are not part of a connector API calls like configurations being printed, initialization logs etc. Also, there are specific points like at the start of a connector API call where would want to set it with a new value even if an invocation id exists currently.

util/src/test/java/com/google/cloud/hadoop/util/InvocationIdContextTest.java
cpriti-os: Done. Thanks

util/src/test/java/com/google/cloud/hadoop/util/LoggingFormatterTest.java
cpriti-os: Done.

util/src/main/java/com/google/cloud/hadoop/util/interceptors/InvocationIdInterceptor.java
```diff
@@ -65,8 +65,10 @@ public void intercept(HttpRequest request) throws IOException {
     final String signatureKey = "Signature="; // For V2 and V4 signedURLs
     final String builtURL = request.getUrl().build();
     if (!builtURL.contains(signatureKey)) {
-      UUID invocationId = UUID.randomUUID();
-      String invocationEntry = GCCL_INVOCATION_ID_PREFIX + invocationId;
+      if (InvocationIdContext.getInvocationId().isEmpty()) {
```
ajayky-os, line 70: Can multiple threads invoke this block simultaneously ? If yes then this does not look thread safe.

cpriti-os: The invocation ID is managed using InheritableThreadLocal in InvocationIdContext, ensuring each thread has its own value. There is no shared mutable state in this interceptor itself, so it is safe for concurrent use as per my understanding. Let me know if you disagree and I can dig deeper to understand if there is any thread safety concern.

PR #1363, opened 2025-05-28
Fix copy for `gRPC` path 

dheerajsngh:
Root Cause : In `gRPC` copy we have precontiion check for [generation matching](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/7f5f6bbb12eff18714dca88b511613819d832965/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java#L581) while in JSON [`copy`](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/7f5f6bbb12eff18714dca88b511613819d832965/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L1139) there is no such check. It was happeining because we were not fetching `generationId` during list leading to being set to null and giving error during read itself. 

Changes for the fix
- Added `generation` in OBJECT_FIELDS to fetch `generation` as well during list operation
- Updated the `GoogleCloudStorageClientGrpcTracingInterceptor` to not put null `uploadId` in list

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -159,7 +159,7 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
 
   // Request only object fields that are used in Hadoop FileStatus:
   // https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations
-  private static final String OBJECT_FIELDS = "bucket,name,size,updated";
+  private static final String OBJECT_FIELDS = "bucket,name,size,updated,generation";
   private static final ListFileOptions LIST_OPTIONS =
       ListFileOptions.DEFAULT.toBuilder().setFields(OBJECT_FIELDS).build();
   private static final String XATTR_KEY_PREFIX = "GHFS_XATTR_";
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcTracingInterceptor.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientGrpcTracingInterceptor.java
@@ -266,7 +266,8 @@ public class GoogleCloudStorageClientGrpcTracingInterceptor implements ClientInt
 
   private class WriteObjectStreamTracer extends TrackingStreamTracer {
 
-    private String streamUploadId = null;
+    private static final String STREAM_UPLOAD_ID_NOT_FOUND = "STREAM_UPLOAD_ID_NOT_FOUND";
+    private String streamUploadId = STREAM_UPLOAD_ID_NOT_FOUND;
 
     WriteObjectStreamTracer(String rpcMethod) {
       super(rpcMethod);
@@ -301,7 +302,7 @@ public class GoogleCloudStorageClientGrpcTracingInterceptor implements ClientInt
           "%s",
           toJson(
               getResponseTrackingInfo()
-                  .put(GoogleCloudStorageTracingFields.UPLOAD_ID.name, streamUploadId)
+                  .put(GoogleCloudStorageTracingFields.UPLOAD_ID.name, this.streamUploadId)
                   .put(
                       GoogleCloudStorageTracingFields.PERSISTED_SIZE.name,
                       response.getPersistedSize())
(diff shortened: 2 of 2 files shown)


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemJavaStorageClientIntegrationTest.java
```diff
@@ -62,10 +62,6 @@ public void unauthenticatedAccessToPublicBuckets_fsGsProperties() {}
   @Test
   public void unauthenticatedAccessToPublicBuckets_googleCloudProperties() {}
 
-  @Ignore
```
animesh-g: Why is this change part of this PR?

animesh-g (changes requested):
Changes look fine but tests only cover the gRPC read channel. Please add tests for JSON read channel as well as gRPC read channel. 

Also PR title doesn't accurately reflect what is being in this change.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemJavaStorageClientIntegrationTest.java
dheerajsngh: Removed. 
Will send a seperate PR for it.

dheerajsngh:
* Added Unit test for JSON
* Updated PR title

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannelTest.java
```diff
@@ -197,6 +199,59 @@ public void readFullObject() throws IOException {
     verifyNoMoreInteractions(fakeReadChannel);
   }
 
+  @Test
+  public void read_withPositiveGeneration_usesGenerationMatchPrecondition() throws IOException {
+    long generation = 12345L;
```
animesh-g, line 204: nit: We can use final here. Same for other methods.

animesh-g (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannelTest.java
```diff
@@ -584,6 +584,61 @@ public void read_gzipEncoded_shouldReadAllBytes() throws IOException {
     assertThat(readChannel.size()).isEqualTo(testData.length);
   }
 
+  @Test
```
ajayky-os, line 587: Unit tests tend to follow a very regular 3-block structure (arrange, act and assert). In simple cases, blank lines can be used to make the boundaries between the blocks clear within a single test. However, when blocks are complex enough to need blank lines within themselves, you can use section comments like `Given [a description of a situation]`, `When [a feature is exercised]`, `"Then [something is the case]` at the start of the block to make the boundaries clear.

For example:

```
// Given [a description of a situation]
... Create local variables to be passed to the SUT
... Create more local variables to be passed to the SUT
<blank line here to separate local variable declarations and stubbings>
... Stub calls
... Stub more calls

// When [a feature is exercised]
... call SUT

// Then [something is the case]
... Build expected value
... Assert something
... Assert something else grouped with the one above
<blank line here to separate assertions and mock verifications>
... Verify a mock interaction
... Verify another mock interaction
```

go/unit-test-practices#structure

The problem is that the blank line here results in more than 3 blocks. Either the blank line should be removed or section comments should be used. Same for all the other specs.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1145,7 +1145,7 @@ protected Storage.Objects.Get createDataRequest() throws IOException {
     Storage.Objects.Get getData =
         storageRequestFactory.objectsGetData(
             resourceId.getBucketName(), resourceId.getObjectName());
-    if (resourceId.hasGenerationId()) {
+    if (resourceId.hasGenerationId() && resourceId.getGenerationId() > 0) {
```
ajayky-os, line 1148: check if line 1163 also needs to be updated.

dheerajsngh: Updated. 
Thanks

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannelTest.java
dheerajsngh: Updated.
Thanks

PR #1373, opened 2025-06-04
Add generation id check for both source and dst during rename

dheerajsngh:
Add generation id check for both source and dst during rename.


animesh-g (commented):
I assume this PR should be closed and master branch PR should be backported here.

dheerajsngh:
Will be writing a proposal to add generation id from gcs layer and pass it on to the gcsio layer. 
Closing this PR.

PR #1379, opened 2025-06-13
Computing and matching checksum on the write channel

animesh-g:
In this PR ,

* Checksum Verification for Writes: Introduces a new feature to calculate a client-side CRC32C checksum during file writes and compare it against the checksum returned by the GCS server upon successful upload. This helps verify data integrity during the transfer.
* Configuration Option: Adds a new Hadoop configuration property, `fs.gs.write.rolling.checksum.enable`, to enable or disable this write checksum verification feature. It is disabled by default.

Testing done -
On Hadoop CLI,
- File write, overwrite and append operation
- Both small(few KBs) and large(500 MB)

On Dataproc Cluster(Trino),
- Ran [CTAS](https://docs.aws.amazon.com/athena/latest/ug/ctas-examples.html) Queries...
- Ran [TPCDS](https://www.tpc.org/tpcds/) queries

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
@@ -26,6 +26,8 @@ import com.google.cloud.hadoop.util.AsyncWriteChannelOptions;
 import com.google.cloud.hadoop.util.ClientRequestHelper;
 import com.google.cloud.hadoop.util.GoogleCloudStorageEventBus;
 import com.google.cloud.hadoop.util.LoggingMediaHttpUploaderProgressListener;
+import com.google.common.io.BaseEncoding;
+import com.google.common.primitives.Ints;
 import java.io.IOException;
 import java.io.InputStream;
 import java.time.Duration;
@@ -117,9 +119,26 @@ public class GoogleCloudStorageWriteChannel extends AbstractGoogleAsyncWriteChan
   }
 
   @Override
-  public void handleResponse(StorageObject response) {
+  public void handleResponse(StorageObject response) throws IOException {
     completedItemInfo = GoogleCloudStorageImpl.createItemInfoForStorageObject(resourceId, response);
-    serverProvidedCrc32c = response.getCrc32c();
+    if (channelOptions.isRollingChecksumEnabled()) {
+      verifyChecksums(response.getCrc32c());
+    }
+  }
+
+  private void verifyChecksums(String serverProvidedCrc32c) throws IOException {
+    String srcCrc =
+        BaseEncoding.base64().encode(Ints.toByteArray(cumulativeCrc32cHasher.hash().asInt()));
+    if (!srcCrc.equals(serverProvidedCrc32c)) {
+      throw new IOException(
+          String.format(
+              "Data integrity check failed for resource '%s'. Client-calculated CRC32C (%s) did not match server-provided CRC32C (%s).",
+              getResourceString(), srcCrc, serverProvidedCrc32c));
+    } else {
+      logger.atFine().log(
+          "Data integrity check passed for resource '%s'. Client-calculated CRC32C (%s) matched the server-provided CRC32C (%s).",
+          getResourceString(), srcCrc, serverProvidedCrc32c);
+    }
   }
 
   /**
--- a/util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
@@ -21,8 +21,6 @@ import static com.google.common.base.Preconditions.checkState;
 import com.google.common.flogger.GoogleLogger;
 import com.google.common.hash.Hasher;
 import com.google.common.hash.Hashing;
-import com.google.common.io.BaseEncoding;
-import com.google.common.primitives.Ints;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -64,12 +62,10 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
 
   private ByteBuffer uploadCache = null;
 
-  private final Hasher cumulativeCrc32cHasher;
+  protected final Hasher cumulativeCrc32cHasher;
 
   // To avoid duplicate calculations in case of reuploadFromCache.
   private boolean reuploadFromCacheInitiated = false;
-
-  protected String serverProvidedCrc32c = "";
 
   /** Construct a new channel using the given ExecutorService to run background uploads. */
   public AbstractGoogleAsyncWriteChannel(
@@ -86,11 +82,11 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
    * Handle the API response.
    *
    * <p>This method is invoked after the upload has completed on the same thread that invokes
-   * close().
+   * close(). It can throw IOException if checksum matching is enabled and mismatched.
    *
    * @param response The API response object.
    */
-  public void handleResponse(T response) {}
+  public void handleResponse(T response) throws IOException {}
 
   /** Returns true if direct media uploads are enabled. */
   public boolean isDirectUploadEnabled() {
@@ -132,12 +128,8 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
     try {
       int originalPosition = buffer.position();
       int writtenBytes = pipeSink.write(buffer);
-      if (this.channelOptions.isRollingChecksumEnabled() && !this.reuploadFromCacheInitiated) {
-        ByteBuffer duplicateBuffer = buffer.duplicate();
-        duplicateBuffer.position(originalPosition);
-        // Only calculate hash for written bytes.
-        duplicateBuffer.limit(writtenBytes);
-        cumulativeCrc32cHasher.putBytes(duplicateBuffer);
+      if (channelOptions.isRollingChecksumEnabled() && !reuploadFromCacheInitiated) {
+        addBytesToCumulativeChecksum(buffer, writtenBytes, originalPosition);
       }
       return writtenBytes;
     } catch (IOException e) {
@@ -146,6 +138,15 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
               "Failed to write %d bytes in '%s'", buffer.remaining(), getResourceString()),
           e);
     }
+  }
+
+  private void addBytesToCumulativeChecksum(
+      ByteBuffer src, int writtenBytes, int originalPosition) {
+    ByteBuffer duplicateBuffer = src.duplicate();
+    duplicateBuffer.position(originalPosition);
+    // Only calculate hash for written bytes.
+    duplicateBuffer.limit(writtenBytes);
+    cumulativeCrc32cHasher.putBytes(duplicateBuffer);
   }
 
   /**
@@ -184,28 +185,10 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
     } finally {
       closeInternal();
     }
-    if (this.channelOptions.isRollingChecksumEnabled()) {
-      compareChecksums();
-    }
-  }
-
-  private void compareChecksums() throws IOException {
-    String srcCrc =
-        BaseEncoding.base64().encode(Ints.toByteArray(cumulativeCrc32cHasher.hash().asInt()));
-    if (!srcCrc.equals(this.serverProvidedCrc32c)) {
-      throw new IOException(
-          String.format(
-              "Data integrity check failed for resource '%s'. Client-calculated CRC32C (%s) did not match server-provided CRC32C (%s).",
-              getResourceString(), srcCrc, this.serverProvidedCrc32c));
-    } else {
-      logger.atFine().log(
-          "Data integrity check passed for resource '%s'. Client-calculated CRC32C (%s) matched the server-provided CRC32C (%s).",
-          getResourceString(), srcCrc, this.serverProvidedCrc32c);
-    }
   }
 
   private void reuploadFromCache() throws IOException {
-    this.reuploadFromCacheInitiated = true;
+    reuploadFromCacheInitiated = true;
     closeInternal();
     initialized = false;
 


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -65,13 +77,60 @@ public GoogleCloudStorageWriteChannel(
...
+      buffer.limit(intExact);
+      consumed = intExact;
+    }
+    cumulativeCrc32c.putBytes(buffer);
+    src.position(src.position() + consumed);
+    return consumed;
+  }
+
+  @Override
+  public void close() throws IOException {
```
ajayky-os: In AbstractGoogleAsyncWriteChannel write() method, it is not guaranteed that the buffer is actually successfully written to GCS on which we are computing cumulativeCrc32c. On looking at close() implementation in AbstractGoogleAsyncWriteChannel it seems it tries to flush the buffer, in some case for this it uses reuploadFromCache(). It may not change the rolling checksum computation, but I am wondering if it can change the actualCrc which is turn result in false alarm ?

breach10ck (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
animesh-g: The current implementation relies on the Apiary layer to initiate retries. The hash calculation is performed within the write methods invoked by HadoopFS, ensuring a single execution per buffer. So, I think hash would be correct in all cases.

Furthermore, any failures or errors encountered during the upload API call will trigger an exception, preventing the close method from being invoked. This mechanism makes the probability of false alarms highly improbable.

Someone in dataproc team would know better on how Spark(and others) behave in this case, maybe they can confirm this.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -129,6 +129,7 @@ public class GoogleHadoopFileSystemConfigurationTest {
           put("fs.gs.write.parallel.composite.upload.part.file.name.prefix", "");
           put("fs.gs.fadvise.request.track.count", 3);
           put("fs.gs.operation.move.enable", false);
+          put("fs.gs.operation.checksum.write.enable", false);
```
singhravidutt: it's not an operation, what about naming it as `fs.gs.write.rolling.checksum.enable`?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -632,7 +641,8 @@ static GoogleCloudStorageOptions.Builder getGcsOptionsBuilder(Configuration conf
         .setOperationTraceLogEnabled(GCS_OPERATION_TRACE_LOG_ENABLE.get(config, config::getBoolean))
         .setTrafficDirectorEnabled(GCS_GRPC_TRAFFICDIRECTOR_ENABLE.get(config, config::getBoolean))
         .setWriteChannelOptions(getWriteChannelOptions(config))
-        .setMoveOperationEnabled(GCS_OPERATION_MOVE_ENABLE.get(config, config::getBoolean));
+        .setMoveOperationEnabled(GCS_OPERATION_MOVE_ENABLE.get(config, config::getBoolean))
+        .setChecksumWriteEnabled(GCS_CHECKSUM_WRITE_ENABLE.get(config, config::getBoolean));
```
singhravidutt: Better to make it part of WriteCHannelOptions.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -440,7 +440,8 @@ public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOpti
             storageOptions.getWriteChannelOptions(),
             resourceId,
             options,
-            writeConditions) {
+            writeConditions,
+            storageOptions) {
```
singhravidutt: Don't have to pass storageOption if checkSum config is part of write options.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -106,6 +106,15 @@ public class GoogleHadoopFileSystemConfiguration {
   public static final HadoopConfigurationProperty<Boolean> GCS_HIERARCHICAL_NAMESPACE_ENABLE =
       new HadoopConfigurationProperty<>("fs.gs.hierarchical.namespace.folders.enable", false);
 
+  /**
+   * Configuration key for checksum writes
+   *
+   * <p>If this is enabled, write channel will calculate rolling checksum and compare it from server
+   * response.
+   */
+  public static final HadoopConfigurationProperty<Boolean> GCS_CHECKSUM_WRITE_ENABLE =
```
cnauroth: For future extensibility, would it make sense for this to be multi-valued (e.g. enum)? For example, if GCS in the future adds a new kind of checksum, we'd be able to support it with a new value for the same configuration key instead of needing to add a second boolean property.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -65,13 +77,61 @@ public GoogleCloudStorageWriteChannel(
...
     this.writeConditions = writeConditions;
+    this.cumulativeCrc32c = Hashing.crc32c().newHasher();
+    this.actualCrc32c = "";
+    this.storageOptions = options;
+  }
+
+  @Override
+  public synchronized int write(ByteBuffer src) throws IOException {
+    ByteBuffer dup = src.duplicate();
+    int written = super.write(src);
```
singhravidutt: why to have that coupling? Why not move hash calculation logic to base class?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannelTest.java
```diff
@@ -49,7 +49,8 @@ public void createRequest_shouldSetKmsKeyName() throws IOException {
             AsyncWriteChannelOptions.DEFAULT,
             new StorageResourceId(BUCKET_NAME, OBJECT_NAME),
             CreateObjectOptions.DEFAULT_NO_OVERWRITE.toBuilder().setKmsKeyName(kmsKeyName).build(),
-            ObjectWriteConditions.NONE);
+            ObjectWriteConditions.NONE,
```
singhravidutt: Would need to write other test cases where we can inject the wrong checksum by mocked transport layer.

animesh-g: Added test cases.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
animesh-g: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
animesh-g: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
animesh-g: Done

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
animesh-g: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
animesh-g: Currently GCS supports both crc32c and md5. We have gone with crc32c intentionally as it is faster and less compute heavy. Also md5 is not available in case of compose(). Collisions are not much of problems for checking data integrity. I don't see the need for supporting multiple types of checksums. It can be evaluated as per the need in future.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
animesh-g: Have updated the implementation to avoid this issue.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -295,6 +295,17 @@ public class GoogleHadoopFileSystemConfiguration {
...
 
+  /**
+   * Configuration key for rolling checksum on writes.
+   *
+   * <p>If this is enabled, write channel will calculate rolling crc32c checksum and compare it from
+   * server response.
+   */
+  public static final HadoopConfigurationProperty<Boolean> GCS_WRITE_ROLLING_CHECKSUM_ENABLE =
+      new HadoopConfigurationProperty<>(
+          "fs.gs.outputstream.rolling.checksum.enable",
```
singhravidutt: mit: make it consistent with config name.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -128,6 +145,20 @@ public synchronized int write(ByteBuffer buffer) throws IOException {
     }
   }
 
+  private long hash(ByteBuffer src, long written) {
```
singhravidutt: why long?

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -128,6 +145,20 @@ public synchronized int write(ByteBuffer buffer) throws IOException {
...
+    ByteBuffer buffer = src.slice();
+    int remaining = buffer.remaining();
+    int consumed = remaining;
+    if (written < remaining) {
+      int intExact = Math.toIntExact(written);
+      buffer.limit(intExact);
+      consumed = intExact;
+    }
+    cumulativeCrc32c.putBytes(buffer);
```
singhravidutt: It's a bit complex for me to follow, 
Shouldn't we be just be adjusting the duplicate buffer's limit to pos()+writtenBytes and provide it to the hasher to read?

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -117,9 +128,15 @@ public synchronized int write(ByteBuffer buffer) throws IOException {
     } else {
       uploadCache = null;
     }
-
     try {
-      return pipeSink.write(buffer);
+      int originalPos = buffer.position();
+      int written = pipeSink.write(buffer);
```
singhravidutt: writtenBytes?

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -117,9 +128,15 @@ public synchronized int write(ByteBuffer buffer) throws IOException {
...
+      if (this.channelOptions.isRollingChecksumEnabled() && !this.isChecksumComputed) {
```
singhravidutt: what's the purpose of `isChecksumComputed`?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -172,7 +172,9 @@ public StorageObject call() throws Exception {
       // Try-with-resource will close this end of the pipe so that
       // the writer at the other end will not hang indefinitely.
       try (InputStream ignore = pipeSource) {
-        return uploadObject.execute();
+        StorageObject response = uploadObject.execute();
+        serverProvidedCrc32c = response.getCrc32c();
```
singhravidutt: why we are extracting values from response and making them a class variable? Wouldn't it be better to let the `handleResponse` method to take care of it?

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -164,6 +195,23 @@ public void close() throws IOException {
...
       closeInternal();
     }
+    if (this.channelOptions.isRollingChecksumEnabled() && this.isChecksumComputed) {
+      compareChecksums();
+    }
+  }
+
+  private void compareChecksums() throws IOException {
+    String srcCrc = BaseEncoding.base64().encode(Ints.toByteArray(cumulativeCrc32c.hash().asInt()));
+    if (!srcCrc.equals(this.serverProvidedCrc32c)) {
```
singhravidutt: Could there be a case where crc hash is not returned in response?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -278,6 +282,84 @@ public void testCreateObjectWithGenerationId() throws Exception {
...
+    try (WritableByteChannel writeChannel =
+        gcs.create(new StorageResourceId(BUCKET_NAME, OBJECT_NAME, 1))) {
+      assertThat(writeChannel.isOpen()).isTrue();
+      writeChannel.write(ByteBuffer.wrap(testData));
+      IOException thrown = assertThrows(IOException.class, writeChannel::close);
+      assertThat(thrown)
+          .hasMessageThat()
+          .isEqualTo(
+              String.format(
+                  "Data integrity check failed for resource 'gs://foo-bucket/bar-object'."
```
singhravidutt: nit: get the path from object, instead of hardcoding it.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: CRC32C is always available for all objects. Empty CRC could also indicate a server side corruption issue. See GCS documentation https://cloud.google.com/storage/docs/xml-api/reference-headers#xgooghash

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
animesh-g: Good catch. That seems to be the better place.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: To avoid duplicate calculation while reuploadCache. Added a comment for the same.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: Done

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: I have cleaned this method up. Return type was there for some debugging. We just need to check change buffer limit before putting the bytes to the buffer.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: Removed.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
animesh-g: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
```diff
@@ -119,6 +119,7 @@ Storage.Objects.Insert createRequest(InputStreamContent inputStream) throws IOEx
   @Override
   public void handleResponse(StorageObject response) {
     completedItemInfo = GoogleCloudStorageImpl.createItemInfoForStorageObject(resourceId, response);
+    serverProvidedCrc32c = response.getCrc32c();
```
singhravidutt: No, what I meant to say was why to even make it a class member and not just use it during hashComparison which should be initiated at handleResponse

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -164,9 +184,28 @@ public void close() throws IOException {
     } finally {
       closeInternal();
     }
+    if (this.channelOptions.isRollingChecksumEnabled()) {
+      compareChecksums();
```
singhravidutt: Can this be initiated from `handleResponse`?

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -117,9 +129,17 @@ public synchronized int write(ByteBuffer buffer) throws IOException {
     } else {
       uploadCache = null;
     }
-
     try {
-      return pipeSink.write(buffer);
+      int originalPosition = buffer.position();
+      int writtenBytes = pipeSink.write(buffer);
+      if (this.channelOptions.isRollingChecksumEnabled() && !this.reuploadFromCacheInitiated) {
```
singhravidutt: can we put hash calculation in separately function.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageWriteChannel.java
animesh-g: I didn't do it because it changes the signature of handleResponse as it will start throwing exception. Also it was in overridden in derived class. But thinking more about it, keeping it in handleResponse seems reasonable so made the changes.

util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
animesh-g: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
cnauroth: Sounds good, will resolve this comment.

singhravidutt:
follow up bug fixed in: https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1385

PR #1385, opened 2025-06-18
[Bug fix] Correcting checksum calculation when received buffer is already moved

animesh-g:
Identified an issue where checksum calculations could be giving false positives. However Hadoop always seems to be writing at buffer starting a 0, so we didn't encounter this problem in our testing(for Hadoop CLI, Spark and Trino). I believe this fix is more critical for clients directly using gcsio e.g. Apache Beam.

testing done,
- Write and append from Hadoop CLI
- Added a unit test for simulating this scenario

I will backport this change to release branches as well.

note: reuploadFromCache could be broken for long time due to similar issue because it resets the buffer at position 0 via buffer.flip() without checking if original write was triggered with a moved buffer. It needs to be handled separately.

### DIFF
--- a/util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
@@ -144,7 +144,7 @@ public abstract class AbstractGoogleAsyncWriteChannel<T> implements WritableByte
       ByteBuffer src, int writtenBytes, int originalPosition) {
     ByteBuffer duplicateBuffer = src.duplicate();
     duplicateBuffer.position(originalPosition);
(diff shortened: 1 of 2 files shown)


util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java
```diff
@@ -145,7 +145,7 @@ private void addBytesToCumulativeChecksum(
     ByteBuffer duplicateBuffer = src.duplicate();
     duplicateBuffer.position(originalPosition);
     // Only calculate hash for written bytes.
```
singhravidutt: may wanna update comment regarding why position+writtenBytes and not just writtenBytes

singhravidutt (approved):

PR #1396, opened 2025-06-29
Avoid NullPointerException when DeleteFolderOperation is timed out

animesh-g:
In the current Implementation of `DeleteFolderOperation`, client side timeouts can cause `NullPointerException` which is often not caught by underlying Applications(e.g. Spark) and causes it to crash. We should be throwing a transient error in these scenarios i.e. `InterruptedException` in this case which eventually causes `IOexception` in the Hadoop operation.

It helps underlying frameworks/applications to retry in case of these failures instead of crashing with a fatal error.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
@@ -116,8 +116,9 @@ class DeleteFolderOperation {
     try {
       FolderInfo folderInfo = folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
       if (folderInfo == null) {
-        // Throwing an exception here because client side timeouts can cause
-        throw new InterruptedException("Timed out while getting an folder from blocking queue");
+        // Throwing an InterruptedException here because client side timeouts can cause folderInfo
+        // to be null.
+        throw new InterruptedException("Timed out while getting an folder from blocking queue.");
       }
       return folderInfo;
     } catch (Exception e) {
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
@@ -85,6 +85,11 @@ class DeleteFolderOperation {
       // this will avoid infinite loop when all folders are deleted
       while (folderSize != 0 && encounteredNoExceptions()) {
         FolderInfo folderToDelete = getElementFromBlockingQueue();
+        if (folderToDelete == null) {
+          // Throwing an IllegalStateException here because client side timeouts can cause
+          // folderToDelete to be null.
+          throw new IllegalStateException("Timed out while getting a folder from blocking queue.");
+        }
         folderSize--;
 
         // Queue the deletion request
@@ -120,16 +125,10 @@ class DeleteFolderOperation {
   }
 
   /** Gets the head from the blocking queue */
-  public FolderInfo getElementFromBlockingQueue()
+  private FolderInfo getElementFromBlockingQueue()
       throws InterruptedException, IllegalStateException {
     try {
-      FolderInfo folderInfo = folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
-      if (folderInfo == null) {
-        // Throwing an IllegalStateException here because client side timeouts can cause folderInfo
-        // to be null.
-        throw new IllegalStateException("Timed out while getting a folder from blocking queue.");
-      }
-      return folderInfo;
+      return folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
     } catch (InterruptedException | IllegalStateException e) {
       logger.atSevere().log(
           "Encountered exception while getting an element from queue in HN enabled bucket : %s", e);
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
@@ -125,11 +125,10 @@ class DeleteFolderOperation {
   }
 
   /** Gets the head from the blocking queue */
-  private FolderInfo getElementFromBlockingQueue()
-      throws InterruptedException, IllegalStateException {
+  private FolderInfo getElementFromBlockingQueue() throws InterruptedException {
     try {
       return folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
-    } catch (InterruptedException | IllegalStateException e) {
+    } catch (InterruptedException e) {
       logger.atSevere().log(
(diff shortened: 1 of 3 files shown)


gcsio/src/test/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperationTest.java
```diff
@@ -87,6 +101,32 @@ public void checkDeletionOrderForHnBucketSkewedFolders() throws InterruptedExcep
...
+        .hasMessageThat()
+        .isEqualTo("Timed out while getting an folder from blocking queue.");
+  }
+
+  private void setMockFolderDeleteBlockingQueue(DeleteFolderOperation deleteFolderOperation)
+      throws Exception {
+    Field queueField = DeleteFolderOperation.class.getDeclaredField("folderDeleteBlockingQueue");
+    queueField.setAccessible(true);
+    queueField.set(deleteFolderOperation, mockFolderDeleteBlockingQueue);
+  }
```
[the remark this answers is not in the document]
animesh-g: Reflection seems much cleaner in this case because it doesn't add a new private constructor to test the application logic. Also [this method](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/fdb061137aaa57101773071da05d84535d9d0077/util/src/test/java/com/google/cloud/hadoop/util/ThreadTraceTest.java#L58) in `ThreadTraceTest` also uses similar strategy.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -114,7 +114,13 @@ public KeySetView<IOException, Boolean> getAllExceptions() {
   /** Gets the head from the blocking queue */
   public FolderInfo getElementFromBlockingQueue() throws InterruptedException {
     try {
-      return folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
+      FolderInfo folderInfo = folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
+      if (folderInfo == null) {
+        // Throwing an InterruptedException here because client side timeouts can cause folderInfo
+        // to be null.
```
dheerajsngh: Can we also add a log to indiacte why we are throwing the error?

animesh-g: We are logging the same in catch block.

singhravidutt: Not sure if this is the right place to catch the null, shouldn't it be at caller side where we extract the items from queue. There we would know if queue should be empty or we anticipate items to be there.

E.g. if queue is empty because of genuine reason then calling a poll on it will always return `null`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -114,7 +114,13 @@ public KeySetView<IOException, Boolean> getAllExceptions() {
...
+        throw new InterruptedException("Timed out while getting an folder from blocking queue.");
```
singhravidutt: doesn't make sense to throw `InterruptedException`. process is not interrupted but in an Illegal state.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -114,7 +114,13 @@ public KeySetView<IOException, Boolean> getAllExceptions() {
   /** Gets the head from the blocking queue */
   public FolderInfo getElementFromBlockingQueue() throws InterruptedException {
     try {
-      return folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
+      FolderInfo folderInfo = folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
+      if (folderInfo == null) {
```
singhravidutt: Does it make sense to capture it here. What datapoints we have here to make sure that it should not be null?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
animesh-g: Done as discussed offline.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
animesh-g: BlockingQueue.poll only returns `null` only when it is timed out as it is doesn't allow NULL to be queued. So it is better throw the exception early instead from its caller. I am not sure what datapoints are you looking for. The NPE that we faced with a customer was also from 1 minute timed out as confirmed from operation traces.

animesh-g:
@singhravidutt Please take a look.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
singhravidutt: So, caller would know if there would be any elements in the queue or not. 
BlockingQueue returning null is a genuine case. Caller would decide if that is favourable or not for eg. if caller known that it need to delete 10 folder. At certain point in time it deleted 5. Further call to get folders from queue resulted ion null response. Now, caller should make a call that it's an Illegal state, if it deleted just 5 instead of 10 folders and yet got nothing from blocking queue.

blockingqueue returning null after waiting for 1 minute just denotes that there is nothing in the queue.

animesh-g: Updated the code. I am fine with either way of implementation because caller insures that getElementFromBlockingQueue is only called with folders remaining to delete. Also this method is was unnecessarily public so made it private as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -77,17 +77,30 @@ class DeleteFolderOperation {
...
+        }
+        folderSize--;
 
-      // Queue the deletion request
-      queueSingleFolderDelete(folderToDelete, /* attempt */ 1);
+        // Queue the deletion request
+        queueSingleFolderDelete(folderToDelete, /* attempt */ 1);
+      }
+    } catch (InterruptedException | IllegalStateException e) {
+      Thread.currentThread().interrupt();
```
singhravidutt, line 99: Why this needs to be flipped in case `IllegalStateException` ?

https://www.baeldung.com/java-interrupted-exception#4-the-interrupt-status-flag

animesh-g: Done. Thanks for pointing out.

singhravidutt (approved):

PR #1397, opened 2025-06-30
Added JMH Benchmark for Delete Operation 

yash-jainn:
[go/gcs-jmh-delete](https://docs.google.com/document/d/1YnSbiVO02Atux4Ei6XMzx0mghzYsT1-an7_cwHqCX8U/edit?resourcekey=0-H2OrlX7JkhOzRYomXFlYmA&tab=t.4vpk4h3utehz)

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/benchmarking/GoogleHadoopFileSystemJMHBenchmarking.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/benchmarking/GoogleHadoopFileSystemJMHBenchmarking.java
@@ -2,6 +2,7 @@ package com.google.cloud.hadoop.fs.gcs.benchmarking;
 
 import com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem;
 import com.google.cloud.hadoop.fs.gcs.benchmarking.JMHBenchmarks.GCSListStatusBenchmark;
+import com.google.cloud.hadoop.fs.gcs.benchmarking.JMHBenchmarks.GCSRenameBenchmark;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


shubham-up-47 (approved):
Looks good to me.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/benchmarking/GoogleHadoopFileSystemJMHBenchmarking.java
```diff
@@ -69,4 +70,18 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+    // Run the single-shot benchmark, which will delete the actual file/folder.
+    runJMHBenchmarkAndLog("DELETE", () -> GCSDeleteBenchmark.runBenchmark(hadoopPath, recursive));
+
+    // IMPORTANT: Because the benchmark itself performs the final delete, we DO NOT call
+    // super.delete() here. We simply return true to indicate the user's command succeeded.
+    logger.atInfo().log(
+        "Benchmark complete. The path '%s' was deleted as part of the benchmark run.", hadoopPath);
+    // This line is only reached if the benchmarked delete operation was successful.
+    // A failure would have thrown an IOException, exiting this method prematurely.
+    return true;
```
dheerajsngh, line 85: who consumes this boolean value? why are we throwing boolean value? why can't we simply return?

dheerajsngh: Got it. The original delete also returns boolean, so this should be okay. Resolving.

dheerajsngh (approved):
Looks good for now. In future, we can think about how how we can increase #iterations.

PR #1404, opened 2025-07-08
Fix: Handle IOException in composeObjects to prevent NPE

cpriti-os:
This PR addresses a NullPointerException that occurred when the GCS connector attempts to compose objects with mismatched storage classes.

The root cause was an unhandled `IOException` (specifically `GoogleJsonResponseException`) thrown by the `composeObjects` method, which resulted in the `composedObject` variable remaining null. Subsequent access to this null object led to the NPE.

Changes in this PR:

1.  Wrapped the `composeObjects` call in `GoogleHadoopSyncableOutputStream.java` with a try-catch block to handle potential `IOException`s. The exception is logged and re-thrown to ensure failures are propagated.
2.  Added a null check for the `composedObject` after the `composeObjects` call to gracefully handle cases where the operation might return null without throwing.
3. Error propagation and cleanup: Propagate the IOException to ensure failures are communicated upstream.
Mark the stream as closed at the appropriate point. Ensure that any temporary files created for the compose operation are cleaned up, preventing resource leaks
4.  Introduced new unit tests in `GoogleHadoopSyncableOutputStreamTest.java`:
    *   `testCommitCurrentFile_throwsIOExceptionOnComposeFailure`: Ensures that `IOException` from `composeObjects` is correctly propagated.
    *   `testCommitCurrentFile_handlesNullComposedObjectGracefully`: Verifies that a null result from `composeObjects` does not lead to an NPE.

These changes ensure that compose failures are handled more gracefully, preventing the NPE and providing better error indication.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopSyncableOutputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopSyncableOutputStream.java
@@ -360,20 +360,33 @@ public class GoogleHadoopSyncableOutputStream extends OutputStream implements Sy
                 "Destination bucket in path '%s' doesn't match temp file bucket in path '%s'",
                 finalGcsPath, curGcsPath));
       }
-      GoogleCloudStorageItemInfo composedObject =
-          ghfs.getGcsFs()
-              .getGcs()
-              .composeObjects(
-                  ImmutableList.of(destResourceId, tempResourceId),
-                  destResourceId,
-                  GoogleCloudStorageFileSystem.objectOptionsFromFileOptions(fileOptions));
-      curDestGenerationId = composedObject.getContentGeneration();
-      deletionFutures.add(
-          cleanupThreadpool.submit(
-              () -> {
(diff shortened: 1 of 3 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopSyncableOutputStream.java
```diff
@@ -360,20 +391,34 @@ private void commitCurrentFile() throws IOException {
...
+                .getGcs()
+                .composeObjects(
+                    ImmutableList.of(destResourceId, tempResourceId),
+                    destResourceId,
+                    GoogleCloudStorageFileSystem.objectOptionsFromFileOptions(fileOptions));
+
+        if (composedObject != null) {
+          curDestGenerationId = composedObject.getContentGeneration();
+        } else {
+          logger.atWarning().log("Composed object is null for destination: %s", finalGcsPath);
```
animesh-g: Should we throw exception here? Without composedObject, we shouldn't try to commit.

cpriti-os: Agreed, threw an exception

PR #1408, opened 2025-07-10
Added changes for read-vectored bidi at the gcsio layer.

ShreyasSinha:
Fastbyte Design doc: [go/fastbyte-gcs-connector](http://goto.google.com/fastbyte-gcs-connector)

Dataproc projects yet to be allowlisted for Fastbyte. I'll add Integeration Test in next cl.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
@@ -18,15 +18,21 @@ package com.google.cloud.hadoop.gcsio;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.api.core.BetaApi;
 import com.google.api.services.storage.model.Folder;
+import com.google.cloud.storage.BlobId;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.ByteBuffer;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.channels.WritableByteChannel;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.function.IntFunction;
 
 /**
  * Interface for exposing the Google Cloud Storage API behavior in a way more amenable to writing
@@ -426,6 +432,23 @@ public interface GoogleCloudStorage {
       ListFolderOptions listFolderOptions,
       String pageToken)
       throws IOException;
+
+  /**
+   * Reads data from Google Cloud Storage using vectored I/O operations.
+   *
+   * @param {@link VectoredIORange} List of ranges to be read.
+   * @param allocate Function to allocate ByteBuffer for reading.
+   * @param blobId Id of the blob to be read.
+   * @return {@link VectoredIOResult} Result of the vectored I/O operation.
+   * @throws IOException on IO error
+   * @throws ExecutionException if the get computation threw an exception
+   * @throws InterruptedException if the current thread was interrupted while waiting for get
+   * @throws TimeoutException if the wait timed out
+   */
+  @BetaApi
+  VectoredIOResult readVectored(
+      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, BlobId blobId)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException;
 
   /**
    * Gets information about an object or a bucket.
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
@@ -20,7 +20,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.api.core.BetaApi;
 import com.google.api.services.storage.model.Folder;
-import com.google.cloud.storage.BlobId;
 import java.io.IOException;
 import java.net.URI;
 import java.nio.ByteBuffer;
@@ -30,8 +29,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
 import java.util.function.IntFunction;
 
 /**
@@ -438,17 +435,13 @@ public interface GoogleCloudStorage {
    *
    * @param {@link VectoredIORange} List of ranges to be read.
    * @param allocate Function to allocate ByteBuffer for reading.
-   * @param blobId Id of the blob to be read.
-   * @return {@link VectoredIOResult} Result of the vectored I/O operation.
+   * @return {@link VectoredIOMetrics} Result of the vectored I/O operation.
    * @throws IOException on IO error
-   * @throws ExecutionException if the get computation threw an exception
-   * @throws InterruptedException if the current thread was interrupted while waiting for get
-   * @throws TimeoutException if the wait timed out
    */
   @BetaApi
-  VectoredIOResult readVectored(
-      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, BlobId blobId)
-      throws IOException, ExecutionException, InterruptedException, TimeoutException;
+  VectoredIOMetrics readVectored(
+      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, URI gcsPath)
+      throws IOException;
 
   /**
    * Gets information about an object or a bucket.
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
@@ -546,9 +546,14 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
   }
 
   @Override
-  public VectoredIOResult readVectored(
-      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, BlobId blobId)
-      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+  public VectoredIOMetrics readVectored(
+      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, URI gcsPath)
+      throws IOException {
+    StorageResourceId resourceId =
+        StorageResourceId.fromUriPath(gcsPath, /* allowEmptyObjectName= */ false);
+    BlobId blobId =
+        BlobId.of(
+            resourceId.getBucketName(), resourceId.getObjectName(), resourceId.getGenerationId());
     logger.atFiner().log("readVectored() called for BlobId=%s", blobId.toString());
     long clientInitializationDurationStartTime = System.currentTimeMillis();
     AtomicInteger totalBytesRead = new AtomicInteger();
@@ -584,11 +589,13 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
       ApiFutures.allAsList(transformFutures).get(30, TimeUnit.SECONDS);
       long rangedReadTime = System.currentTimeMillis() - rangedReadStartTime;
       logger.atFiner().log("Ranged read successful in %d", rangedReadTime);
-      return VectoredIOResult.builder()
+      return VectoredIOMetrics.builder()
           .setReadBytes(totalBytesRead.get())
           .setReadDuration(rangedReadTime)
           .setClientInitializationDuration(clientInitializationDuration)
           .build();
+    } catch (IOException | ExecutionException | InterruptedException | TimeoutException e) {
+      throw new IOException(e);
     }
   }
 
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
@@ -54,6 +54,7 @@ import com.google.cloud.storage.Storage.BlobTargetOption;
 import com.google.cloud.storage.Storage.MoveBlobRequest;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
+import com.google.cloud.storage.ZeroCopySupport.DisposableByteString;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
@@ -63,6 +64,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.flogger.GoogleLogger;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.protobuf.ByteString;
 import io.grpc.ClientInterceptor;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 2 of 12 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -427,6 +433,23 @@ ListPage<FolderInfo> listFolderInfoForPrefixPage(
...
+   * @param blobId Id of the blob to be read.
+   * @return {@link VectoredIOResult} Result of the vectored I/O operation.
+   * @throws IOException on IO error
+   * @throws ExecutionException if the get computation threw an exception
+   * @throws InterruptedException if the current thread was interrupted while waiting for get
+   * @throws TimeoutException if the wait timed out
+   */
+  @BetaApi
+  VectoredIOResult readVectored(
+      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, BlobId blobId)
```
singhravidutt: BlobId is SDK's model do we really want it to be part of gcsio interface?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
```diff
@@ -427,6 +433,23 @@ ListPage<FolderInfo> listFolderInfoForPrefixPage(
...
+      throws IOException, ExecutionException, InterruptedException, TimeoutException;
```
singhravidutt: Not sure if it make sense to expose all concurrent exception to clients, can't we make decision about them in library in itself. Meaning, what should client do differently for `ExecutionException` (and others) and `IOException` ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/VectoredIOResult.java
```diff
@@ -0,0 +1,48 @@
+package com.google.cloud.hadoop.gcsio;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
+
+@AutoValue
+/** Result of the complete vectoredIO read operation. */
+public abstract class VectoredIOResult {
```
singhravidutt: this is not a result. As in, it's not having the data of requested range but bunch of metrics we collect

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -514,6 +545,62 @@ private static ExecutorSupplier getPCUExecutorSupplier(ExecutorService pCUExecut
         : ExecutorSupplier.useExecutor(pCUExecutorService);
   }
 
+  @Override
+  public VectoredIOResult readVectored(
```
singhravidutt: Looks like we have made it as a sync API, it defeats the purpose of gathering data as Future in VectoredIORange.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -514,6 +545,62 @@ private static ExecutorSupplier getPCUExecutorSupplier(ExecutorService pCUExecut
...
+      List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate, BlobId blobId)
+      throws IOException, ExecutionException, InterruptedException, TimeoutException {
+    logger.atFiner().log("readVectored() called for BlobId=%s", blobId.toString());
+    long clientInitializationDurationStartTime = System.currentTimeMillis();
+    AtomicInteger totalBytesRead = new AtomicInteger();
+    try (BlobReadSession blobReadSession =
+        storage.blobReadSession(blobId).get(30, TimeUnit.SECONDS)) {
```
singhravidutt: nit: why this is a future?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -514,6 +545,62 @@ private static ExecutorSupplier getPCUExecutorSupplier(ExecutorService pCUExecut
...
+      long clientInitializationDuration =
+          System.currentTimeMillis() - clientInitializationDurationStartTime;
+      logger.atFiner().log("Client Initialization successful in %d", clientInitializationDuration);
+      Map<VectoredIORange, ApiFuture<byte[]>> futures =
+          ranges.stream()
+              .collect(
+                  Collectors.toMap(
+                      Function.identity(),
+                      range ->
+                          blobReadSession.readAs(
```
singhravidutt: From https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/ReadProjectionConfigs.java#L99

It seems like sdk will use a memory space to hold on to the data and then we are copying it back to the memory space provided by the client via allocate. Am I correct?

What about us the using `ReadAsChannel` and pass on the array from the allocate?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
ShreyasSinha: Good catch. I agree doesn't make sense to have SDK's model in interface, replaced it with gcspath.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorage.java
ShreyasSinha: The idea is to allow consumer (if required) to have different behaviour for different set of exceptions. But I agree, I can't think of anything specific as of now the client would want to do with these apart from retrying for just few cases. Also I re-checked the whole interface, we are exposing IOException. 
I made the change to merge all exception into IOException.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/VectoredIOResult.java
ShreyasSinha: I agree its better to rename it VectoredIOMetrics.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
ShreyasSinha: Its an unfortunate naming convention, it doesn't hold the result. Changed it to VectoredIOMetrics instead.
The result is part of the input range which is populated async.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
ShreyasSinha: Its a SDK implementation for fastbyte bidi api.
It asynchronously sets up a new session corresponding to an object which can be used to read multiple times from a single object without performing  initialization, authentication checks and metadata lookups repeatedly.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
ShreyasSinha: I think you linked incorrectly, I am using asFutureBytes not asFutureByteString.

BenWhitehead: @singhravidutt Is correct, the bytes will accumulate in netty's pool until the future resolves, at which point they will be released after creating a single contiguous `byte[]` on the heap. Until the future resolves, no bytes will be held on the gc managed heap of the process.

Using a channel would reduce the accumulation in netty's pool (one in progress message + one pending message) would be the maximum buffer depth, but you would need a thread to read the bytes through the channel.

If you want to avoid the `byte[]` allocation, you can use the `asFutureByteString()` config. I'll add another comment with a diff showing how I'd write this.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/FakeBlobReadSession.java
```diff
@@ -0,0 +1,43 @@
...
+  public <Projection> Projection readAs(ReadProjectionConfig<Projection> readProjectionConfig) {
+    assertThat(readProjectionConfig).isInstanceOf(ReadAsFutureBytes.class);
+    RangeSpec range = ((ReadAsFutureBytes) readProjectionConfig).getRange();
+    return (Projection)
+        ApiFutures.immediateFuture(getSubString(range).getBytes(StandardCharsets.UTF_8));
+  }
+
+  private String getSubString(RangeSpec range) {
+    return TEST_STRING.substring(
+        (int) range.begin(), (int) (range.begin() + range.maxLength().getAsLong()));
```
BenWhitehead: I know this is test code, but I recommend using `Math.toIntExact` to convert the longs to ints. That way you are protected if overflow would happen.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -514,6 +545,69 @@ private static ExecutorSupplier getPCUExecutorSupplier(ExecutorService pCUExecut
...
+                  ApiFutures.transform(
+                      future,
+                      result -> {
+                        totalBytesRead.addAndGet(populateFileRangeFuture(result, allocate, range));
+                        return null;
+                      },
+                      boundedThreadPool)));
+      long rangedReadStartTime = System.currentTimeMillis();
+      // We need to wait for the futures before exiting the try-with-resources in order to avoid
+      // parent stream closed exception.
```
BenWhitehead: The following is how i'd write this interaction.

A few items to call out:
1. Starting the range read timer before submitting any of the ranges. The bytes will be submitted to gcs as soon as `readAs` completes, not when `future.get` is invoked.
2. Avoiding the map allocation. In your version, the keys of the map is only used in the second forEach. Instead, in my version the range is closed over by the transform created in the `map`
3. Transform the future returned from `readAs` immediately, we don't actually need that future once we define the transform.
4. by having the transform in the map, we can close over the range and don't need a map and another iteration of all the elements
5. Switch to using `DisposableByteString` to avoid any bytes being put on the heap by this code, or the sdk.

```suggestion
    // start opening the BlobReadSession
    ApiFuture<BlobReadSession> sessionFuture = storage.blobReadSession(blobId);
    try (BlobReadSession blobReadSession = sessionFuture.get(30, TimeUnit.SECONDS)) {
      long clientInitializationDuration =
          System.currentTimeMillis() - clientInitializationDurationStartTime;
      logger.atFiner().log("Client Initialization successful in %d", clientInitializationDuration);
      long rangedReadStartTime = System.currentTimeMillis();
      List<ApiFuture<?>> transformFutures = ranges.stream()
          .map(range -> {
            ApiFuture<DisposableByteString> futureBytes = blobReadSession.readAs(
                ReadProjectionConfigs.asFutureByteString()
                    .withRangeSpec(RangeSpec.of(range.getOffset(), range.getLength())));

            return ApiFutures.transform(futureBytes, disposableByteString -> {
              // wrap in try-with to ensure it is closed and the netty memory will be cleared
              try (DisposableByteString dbs = disposableByteString) {
                ByteString byteString = dbs.byteString();
                int size = byteString.size();
                ByteBuffer buf = ByteBuffer.allocate(size);
                // iterate over the sub-buffers that make up the ByteString to avoid any copying
                // of bytes into the heap before we copy them into buf
                for (ByteBuffer b : byteString.asReadOnlyByteBufferList()) {
                  buf.put(b);
                }
                buf.flip();
                totalBytesRead.addAndGet(size);
                range.getData().complete(buf);
                return null;
              } catch (IOException e) {
                // TODO: does something more connector appropriate here
                throw new RuntimeException(e);
              }
            }, boundedThreadPool);
          })
          .collect(Collectors.toList());
```

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientTest.java
```diff
@@ -0,0 +1,111 @@
...
+    VectoredIOMetrics result =
+        gcsClientImpl.readVectored(
+            ranges, allocator, new URI("gs://" + TEST_BUCKET_NAME + "/" + TEST_OBJECT_NAME));
+
+    assertEquals(getReadVectoredData(ranges.get(0)), FakeBlobReadSession.SUBSTRING_20_10);
+    assertEquals(getReadVectoredData(ranges.get(1)), FakeBlobReadSession.SUBSTRING_50_7);
+    assertEquals(getReadVectoredData(ranges.get(2)), FakeBlobReadSession.SUBSTRING_65_17);
+
+    assertThat(result.getClientInitializationDuration()).isLessThan(100L);
+    assertThat(result.getReadDuration()).isLessThan(100L);
```
BenWhitehead: If these times are important to your test -- or any tests for that matter -- I recommend providing a clock or function to the instance under test where you can control the advancement of the time in your test instead of relying on wall time to cooperate.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientTest.java
```diff
@@ -0,0 +1,111 @@
...
+  private static final String TEST_OBJECT_NAME = "foo-object";
+
+  private static MockHttpTransport transport = mockTransport();
+
+  @Test
+  public void readVectored_successfulRead()
+      throws IOException, ExecutionException, InterruptedException, TimeoutException,
+          URISyntaxException {
+    GoogleCloudStorageClientImpl gcsClientImpl = getMockedGcsClientImpl(false);
+    IntFunction<ByteBuffer> allocator = (length) -> ByteBuffer.allocateDirect(length);
```
BenWhitehead: nit
```suggestion
    IntFunction<ByteBuffer> allocator = ByteBuffer::allocateDirect;
```

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientTest.java
```diff
@@ -0,0 +1,111 @@
...
+    assertEquals(getReadVectoredData(ranges.get(2)), FakeBlobReadSession.SUBSTRING_65_17);
+
+    assertThat(result.getClientInitializationDuration()).isLessThan(100L);
+    assertThat(result.getReadDuration()).isLessThan(100L);
+  }
+
+  @Test
+  public void readVectored_returnsClientDuration() throws IOException, URISyntaxException {
+    GoogleCloudStorageClientImpl gcsClientImpl = getMockedGcsClientImpl(true);
+    IntFunction<ByteBuffer> allocator = (length) -> ByteBuffer.allocateDirect(length);
```
BenWhitehead: nit
```suggestion
    IntFunction<ByteBuffer> allocator = ByteBuffer::allocateDirect
```

BenWhitehead (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientTest.java
ShreyasSinha: Makes sense. Need some broader changes rather than simply mocking the SUT, will do in future cls.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/FakeBlobReadSession.java
ShreyasSinha: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
ShreyasSinha: Refactored accordingly.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
ShreyasSinha: I think there are a couple of things being discussed here:

1) Channel vs ByteString:
I think both of these are efficient in terms of space as they will just use the netty's pool as compared to byte[] which is using additional heap space as well.

2) ByteString vs Byte[]:
My ultimate aim was to use ByteString only.
I initially used byte[] to benchmark its memory usage against ByteString, as a prior discussion with Ben suggested there would be a negligible performance difference. Now that I'm confident we have addressed potential lifecycle management concerns with ByteString, particularly around unintentionally retaining Netty memory, I have updated the implementation to use it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
singhravidutt: No, I meant to say that the way we have coded this API it's a sync APi and Futures we have added in VectoredIORanges are of not much use.

PR #1413, opened 2025-07-11
Fixing DeleteFolder stalling issue where high CPU usage can cause timeouts

animesh-g:
RCA - 
In case of HNS, Folders are deleted in Bottom up BFS(breadth first search) manner and it gets timed out after a minute if it doesn't find any more folder to delete in the queue. If a folder has high amount of subfolders and/or thread scheduling is slow due to high CPU usage , it may take longer than a minute to delete all child folders before moving to the parent one. We found one of such failures in the [CI pipeline](https://screenshot.googleplex.com/6YpYaDrt4zqESsW) as well as some users have reported the same. Initially it was attributed to the network congestion but that is ruled out after above CI pipeline failed in a unit test with mock implementation of network request. 

It is a follow up after https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1396 fix.

Summary of changes -
- Add a new method in `BatchExecutor` to detect idle threads
- Stop throwing exception if poll expires on the blocking queue. Instead of throw exception when ThreadPoolExecuter reaches an Idle state with some folders still pending to delete.
- A minor correction is spelling of `addToBatchExecutorQueue` method.
- Always shutdown ThreadPoolExecuter even in case of exceptions to avoid resource leaks
- Changed computeChildrenForFolderResource method to private

See [bug](https://buganizer.corp.google.com/issues/425906826) for more details.

Note: Once a recursive delete API is available in GCS for HNS bucket, we can replace this implementation with a much more simpler one.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperation.java
```diff
@@ -134,7 +139,8 @@ public KeySetView<IOException, Boolean> getAllExceptions() {
   /** Gets the head from the blocking queue */
   private FolderInfo getElementFromBlockingQueue() throws InterruptedException {
     try {
-      return folderDeleteBlockingQueue.poll(1, TimeUnit.MINUTES);
+      // Average latency of DeleteFolder is around 200 ms.
+      return folderDeleteBlockingQueue.poll(200, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
animesh-g: Explained in the comment.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/BatchExecutor.java
```diff
@@ -102,6 +102,22 @@ private static <T> void execute(Callable<T> task, FutureCallback<T> callback) th
...
+   * stalling issues in the jobs.
+   */
+  public boolean isIdle() {
+    if (requestsExecutor instanceof ThreadPoolExecutor) {
+      ThreadPoolExecutor executor = (ThreadPoolExecutor) requestsExecutor;
+      // Thread's idle if no threads are active and the work queue is empty.
+      return executor.getActiveCount() == 0 && executor.getQueue().isEmpty();
+    }
+    // The executor is the instance of a DirectExecutorService, it is always idle because there are
+    // no child
```
ajayky-os: nit: unnecessary line break in comment

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/BatchExecutorTest.java
```diff
@@ -54,6 +57,48 @@ public void queue_throwsException_afterShutdownCalled() throws IOException {
         .startsWith("requestExecutor should not be terminated to queue request");
   }
 
+  @Test
+  public void isIdle_forThreadPoolExecutor_returnsCorrectState() throws Exception {
```
ajayky-os: Unit tests tend to follow a very regular 3-block structure (arrange, act and assert). In simple cases, blank lines can be used to make the boundaries between the blocks clear within a single test. However, when blocks are complex enough to need blank lines within themselves, you can use section comments like `Given [a description of a situation]`, `When [a feature is exercised]`, `"Then [something is the case]` at the start of the block to make the boundaries clear.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/BatchExecutorTest.java
```diff
@@ -54,6 +57,48 @@ public void queue_throwsException_afterShutdownCalled() throws IOException {
...
+    BatchExecutor batchExecutor = new BatchExecutor(/* numThreads= */ 1);
+
+    assertThat(batchExecutor.isIdle()).isTrue();
```
ajayky-os, line 64: This behaviour of default `isIdle()` returns true can be in seperate spec. This will also help in keeping the arrange, act, assert block seperate in this spec.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/DeleteFolderOperationTest.java
```diff
@@ -118,7 +119,30 @@ public void checkExceptionTypeWhenPollTimesOutForPerformDeleteOperation() throws
...
+        new FolderInfo(
+            FolderInfo.createFolderInfoObject(BUCKET_NAME, "test-folder/test-folder-sleep"));
+    foldersToDelete.add(testFolder1);
+    foldersToDelete.add(testFolder2);
+
+    CustomDeleteFolderOperationTest deleteFolderOperation =
+        new CustomDeleteFolderOperationTest(
+            foldersToDelete, GoogleCloudStorageOptions.DEFAULT, null);
+
+    deleteFolderOperation.performDeleteOperation();
```
ajayky-os, line 142: delete blank line before this statement, this will keep the act block grouped together.

breach10ck (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/BatchExecutorTest.java
animesh-g: Added some comments for clarification.

PR #1422, opened 2025-07-17
Added fastbyte through enhanced channel implementation

ShreyasSinha:
Fastbyte Design doc: [go/fastbyte-gcs-connector](http://goto.google.com/fastbyte-gcs-connector)

Dataproc projects yet to be allowlisted for Fastbyte. I'll add Integeration Test in next cl.

It is guarnteed that close will be called at https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java#L291 which will ultimately close BlobReadSession.


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+            taskQueue,
+            new ThreadFactoryBuilder()
+                .setNameFormat("vectoredRead-range-pool-%d")
+                .setDaemon(true)
+                .build());
+  }
+
+  private void initializeBlobReadSession() throws IOException {
+    try {
+      this.blobReadSession = storage.blobReadSession(blobId).get(30, TimeUnit.SECONDS);
```
BenWhitehead: If you already have a read timeout type of config, it'd probably make sense to use it here instead of hard coding 30s.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+        new ThreadPoolExecutor(
+            readOptions.getBidiThreadCount(),
+            readOptions.getBidiThreadCount(),
+            0L,
+            TimeUnit.MILLISECONDS,
+            taskQueue,
+            new ThreadFactoryBuilder()
+                .setNameFormat("vectoredRead-range-pool-%d")
+                .setDaemon(true)
+                .build());
```
BenWhitehead: This will create a 16 thread thread pool for each channel created. That could be a lot of threads, and a lot of thread churn if the channels are short lived.

It probably makes sense to declare this pool some place else, and pass it in here so the thread can be shared and live longer then the individual channels.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.function.IntFunction;
+
+public class GoogleCloudStorageBidiReadChannel implements SeekableByteChannel, ReadVectored {
+
+  private final Storage storage;
+  private final StorageResourceId resourceId;
+  private final BlobId blobId;
+  private final GoogleCloudStorageReadOptions readOptions;
+  private BlobReadSession blobReadSession;
```
BenWhitehead: Keeping the session open for the life of the channel makes a lot of sense lifecycle wise. And closing the session when the channel is closed ensures we area able to clean up and not leak memory.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+            0L,
+            TimeUnit.MILLISECONDS,
+            taskQueue,
+            new ThreadFactoryBuilder()
+                .setNameFormat("vectoredRead-range-pool-%d")
+                .setDaemon(true)
+                .build());
+  }
+
+  private void initializeBlobReadSession() throws IOException {
```
BenWhitehead: nit: make this method state, pass in storage and blobId to it, and return the resolved session which can then be assigned to a final field in the constructor.

Having constructors call into their own instances methods can create complex interactions and I prefer to avoid.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      this.blobReadSession = storage.blobReadSession(blobId).get(30, TimeUnit.SECONDS);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new IOException(e);
+    }
+  }
+
+  @Override
+  public int read(ByteBuffer dst) throws IOException {
```
BenWhitehead, line 61: [ReadProjectionConfigs#asSeekableChannel()](https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.ReadProjectionConfigs#com_google_cloud_storage_ReadProjectionConfigs_asSeekableChannel__) could be used to create a SeekableByteChannel attached to the session in order to provide an implementation to these read related methods.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,161 @@
...
+                  ReadProjectionConfigs.asFutureByteString()
+                      .withRangeSpec(RangeSpec.of(range.getOffset(), range.getLength())));
+          ApiFutures.transform(
+              futureBytes,
+              disposableByteString -> {
+                try {
+                  return processBytesAndCompleteRange(disposableByteString, range, allocate);
+                } catch (IOException e) {
+                  throw new RuntimeException(e);
+                }
```
BenWhitehead: probably want to change this to

```suggestion
                } catch (Throwable t) {
                  range.getData().completeExceptionally(t)
                  return null;
                }
```

Propogating `Throwable` will ensure that if something like an out of memory error were to happen the future would be notified and it wouldn't just be lost in the background.

ShreyasSinha: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Ack. The remaining read methods will be implemented in follow up cl. Already added todo for each.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Ack.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Done, move it to StorageClientImpl.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,151 @@
...
+    this.blobId =
+        BlobId.of(
+            resourceId.getBucketName(), resourceId.getObjectName(), resourceId.getGenerationId());
+    this.readOptions = readOptions;
+    this.blobReadSession =
+        initializeBlobReadSession(storage, blobId, readOptions.getBidiClientTimeout());
+    this.boundedThreadPool = boundedThreadPool;
+  }
+
+  private BlobReadSession initializeBlobReadSession(
```
BenWhitehead:
```suggestion
  private static BlobReadSession initializeBlobReadSession(
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,151 @@
...
+public class GoogleCloudStorageBidiReadChannel implements SeekableByteChannel, ReadVectored {
+
+  private final Storage storage;
+  private final StorageResourceId resourceId;
+  private final BlobId blobId;
+  private final GoogleCloudStorageReadOptions readOptions;
+  private final BlobReadSession blobReadSession;
+  private ExecutorService boundedThreadPool;
+
+  private final BlockingQueue<Runnable> taskQueue = new LinkedBlockingQueue<>();
```
BenWhitehead: can probably be removed now that you moved the thread pool init out of here

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,151 @@
...
+              disposableByteString -> {
+                try {
+                  return processBytesAndCompleteRange(disposableByteString, range, allocate);
+                } catch (Throwable t) {
+                  range.getData().completeExceptionally(t);
+                  return null;
+                }
+              },
+              boundedThreadPool);
+        });
```
BenWhitehead: You'll want to change this from `transform` instead to `addCallback` so that you can observe if the future fails. `transform` will only apply to a successful resolution, and is generally useful when you want to apply a modification to a future you intend to return for awaiting elsewhere.

```suggestion
          ApiFutures.addCallback(
              futureBytes,
              new ApiFutureCallback<>() {
                @Override
                public void onFailure(Throwable t) {
                  range.getData().completeExceptionally(t);
                }

                @Override
                public void onSuccess(DisposableByteString disposableByteString) {
                  try {
                    processBytesAndCompleteRange(disposableByteString, range, allocate);
                  } catch (Throwable t) {
                    range.getData().completeExceptionally(t);
                  }
                }
              },
              boundedThreadPool);
```

BenWhitehead (commented):
After applying the recommended changes you will need to get someone more familiar with the connector to review this for that side of things.

Having the read session have a wide lifecycle makes sense and makes the lifecycle management clearer.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: acknowledged

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Oops, missed it. Thanks

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,155 @@
...
+                  } catch (Throwable t) {
+                    range.getData().completeExceptionally(t);
+                  }
+                }
+              },
+              boundedThreadPool);
+        });
+  }
+
+  private Void processBytesAndCompleteRange(
```
animesh-g: nit: any particular reason for boxing the primitive type?

ShreyasSinha: Its reminiscence of the earlier implementation when it was used inside a generic. I agree no need for this to still be Void Class and can be replaced with keyword void.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -577,6 +577,25 @@ public class GoogleHadoopFileSystemConfiguration {
           "fs.gs.operation.move.enable",
           GoogleCloudStorageOptions.DEFAULT.isMoveOperationEnabled());
 
+  /** Flag to enable the bidirectional Rapid Storage API. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_OPERATION_BIDI_API_ENABLE =
+      new HadoopConfigurationProperty<>(
+          "fs.gs.operation.bidi.enable", GoogleCloudStorageReadOptions.DEFAULT.isBidiEnabled());
```
ajayky-os: The prefix in this case `fs.gs.operation.bidi` is different from other configurations ie: `fs.gs.bidi`.

IMO we should have a consistent prefix for a feature. Bidi is not an operation but a feature so we can remove the operation prefix.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,154 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.IntFunction;
+
+public class GoogleCloudStorageBidiReadChannel implements SeekableByteChannel, ReadVectored {
```
ajayky-os: Did we evaluate inheriting the existing read channel implementations so that we don't have to implement the other methods like read, seek, position, etc ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public void readVectored(List<VectoredIORange> ranges, IntFunction<ByteBuffer> allocate)
+      throws IOException {
+    ranges.forEach(
+        range -> {
+          ApiFuture<DisposableByteString> futureBytes =
+              blobReadSession.readAs(
+                  ReadProjectionConfigs.asFutureByteString()
```
ajayky-os, line 113: In client side implementation of vectored read, we are also merging the nearby ranges. Is this implementation faster  and does not require such optimization or handled in veneer?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
```diff
@@ -0,0 +1,122 @@
...
+  public void readVectored_successfulRead()
+      throws IOException, ExecutionException, InterruptedException, TimeoutException,
+          URISyntaxException {
+    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel();
+    IntFunction<ByteBuffer> allocator = ByteBuffer::allocateDirect;
+    List<VectoredIORange> ranges = getListOfVectoredIORange();
+
+    bidiReadChannel.readVectored(ranges, allocator);
+
+    assertEquals(getReadVectoredData(ranges.get(0)), FakeBlobReadSession.SUBSTRING_20_10);
```
ajayky-os, line 62: The relation of `FakeBlobReadSession.SUBSTRIN_x_y` is not clear from the spec ie: where it is coming from and what is the integration/relationship with ranges and getReadVectoredData, ideally all the relationship should be clear by reading the body of the test spec.

ShreyasSinha: Added comments to to make the test case more readable.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Given we are able to re-use the same stream it might not be beneficial to merge ranges.

Abhishek Modi had the same opinion here
https://docs.google.com/document/d/1aRmWcipn6YAOo-wCRowRPPwupZloX-Z5MRFVIhwb-Vg/edit?disco=AAABiAnRyM4 

We will do experiments at our end, with / without this optimization to see if we see any performance improvement post initial implementation.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: Inheriting the existing won't work as we want this channel to implement reads and writes through Bidi API not the existing APIs.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
ShreyasSinha: Agreed. Dropping operation prefix.

gcs/dependency-reduced-pom.xml
```diff
@@ -3,12 +3,12 @@
   <parent>
     <artifactId>bigdataoss-parent</artifactId>
     <groupId>com.google.cloud.bigdataoss</groupId>
-    <version>3.1.4-SNAPSHOT</version>
```
singhravidutt: wrongly added?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -179,12 +182,27 @@ public void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuffe
...
         () -> {
-          long startTimeNs = System.nanoTime();
-          vectoredIOSupplier
-              .get()
-              .readVectored(ranges, allocate, gcsFs, fileInfo, gcsPath, streamStatistics);
-          statistics.incrementReadOps(1);
-          vectoredReadStats.updateVectoredReadStreamStats(startTimeNs);
+          if (channel instanceof ReadVectored) {
+            ReadVectored readVectoredChannel = (ReadVectored) channel;
+            readVectoredChannel.readVectored(
```
singhravidutt: But readVectored is only supported via Grpc for non GRPC it would throw UnsupportedOperation exception.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -577,6 +577,25 @@ public class GoogleHadoopFileSystemConfiguration {
           "fs.gs.operation.move.enable",
           GoogleCloudStorageOptions.DEFAULT.isMoveOperationEnabled());
 
+  /** Flag to enable the bidirectional Rapid Storage API. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_OPERATION_BIDI_API_ENABLE =
```
singhravidutt, line 581: would there be multiple bidi APIs? If yes, wouldn't it be controlled via same config?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -0,0 +1,154 @@
...
+public class GoogleCloudStorageBidiReadChannel implements SeekableByteChannel, ReadVectored {
+
+  private final Storage storage;
+  private final StorageResourceId resourceId;
+  private final BlobId blobId;
+  private final GoogleCloudStorageReadOptions readOptions;
+  private final BlobReadSession blobReadSession;
+  private ExecutorService boundedThreadPool;
+
+  public GoogleCloudStorageBidiReadChannel(
```
singhravidutt, line 32: why we are creatign a seprate channel for it. Wouldn't it make sense to just add implementation of `readVectored` API in existing read channels?

Do we expect `read` API flow to differ from GCSClientReadChannel?
This will add too much tech debt.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ReadVectored.java
```diff
@@ -0,0 +1,11 @@
+package com.google.cloud.hadoop.gcsio;
```
singhravidutt: License headers

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ReadVectored.java
```diff
@@ -0,0 +1,11 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.function.IntFunction;
+
+public interface ReadVectored {
```
singhravidutt: This should extend form SeekableByteChannel and add another method `readVectored`. We could have just updated our open() API to return new custom `ByteChannel` instead of ` SeekableByteChannel`.

How would caller of vectoredIO even know about the `readVectored` API?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -179,12 +182,27 @@ public void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuffe
...
+                ranges.stream()
```
singhravidutt, line 189: I would recommend having integration in gcsio as a followup PR.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
ShreyasSinha: This flow will never be triggered for non-GRPC flow.

GoogleCloudStorageBidiReadChannel (instance of readVectored) is only enabled for gRPC not JSON.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
ShreyasSinha: Acknowledged.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
ShreyasSinha: Yes, there would be multiple API. All will be configured via the same flag.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
ShreyasSinha: We are creating a separate channel as it will implement all flows using the Bidi Api (mandatory for zonal buckets). The other (existing) channel implements using the regular api.

Read flow will be totally different between the two.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ReadVectored.java
ShreyasSinha: Keeping readVectored as a separate interface helps us differentiate with the existing Read Channel (which uses regular APIs) GoogleCloudStorageBidiReadChannel implements both ReadVectored and SeekableByteChannel.

Caller of vectoredIO (assuming you are asking from Hadoop layer) should intentionally not be aware of this interface.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/ReadVectored.java
ShreyasSinha: Copied from another file. Let me know if thats not how that works.

PR #1423, opened 2025-07-23
[ON HOLD] feat: Storage client caching

shrutisinghania:
This PR introduces a caching mechanism for the GCS Connector's Storage Client to enhance performance and resource management, particularly for gRPC. This feature is disabled by default and can be enabled with the flag `fs.gs.storage.client.caching.enable`

## Testing
* <b>Correctness:</b> Verified shared VeneerGrpcClient instance across multiple FileSystem objects with identical configurations 
* <b>Performance:</b> With caching enabled, benchmark tests demonstrated that jobs previously failing due to out-of-memory (OOM) errors now complete quickly and efficiently, showing improved performance when reading from GCS buckets.

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageJavaStorageClientCachingIntegrationTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageJavaStorageClientCachingIntegrationTest.java
@@ -1,6 +1,5 @@
 package com.google.cloud.hadoop.gcsio;
 
-import static com.google.cloud.hadoop.gcsio.integration.GoogleCloudStorageTestHelper.getCredentials;
 import static com.google.cloud.hadoop.gcsio.integration.GoogleCloudStorageTestHelper.getStandardOptionBuilder;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
@@ -53,7 +52,6 @@ public class GoogleCloudStorageJavaStorageClientCachingIntegrationTest {
     optionsBuilder.setStorageClientCachingEnabled(enableCaching);
     return GoogleCloudStorageClientImpl.builder()
         .setOptions(optionsBuilder.build())
-        .setCredentials(getCredentials())
         .setCredentials(null)
         .setDownscopedAccessTokenFn(ignore -> "testDownscopedAccessToken")
         .build();


gcsio/pom.xml
```diff
@@ -206,6 +206,12 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-testing</artifactId>
     </dependency>
+    <dependency>
```
ajayky-os, line 213: gcsio does not depend on any hadoop specific classes.. this seems to be going anti pattern.

gcs/CONFIGURATION.md
```diff
@@ -486,6 +486,13 @@ better latency and increased bandwidth. Currently supported only for read/write
    * `CHUNK_UPLOAD` uploads file in chunks, size of chunks are configurable via
      `fs.gs.outputstream.upload.chunk.size`
 
+*  `fs.gs.storage.client.caching.enable` (default: `false`) enables gRPC storage
```
cpriti-os: Just a reword suggestion: *  `fs.gs.storage.client.caching.enable` (default: `false`)
   Enables caching of gRPC storage clients. When set to `true`, FileSystem instances with similar configuration parameters will reuse the same gRPC client, which can improve performance and reduce resource usage.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -508,6 +508,23 @@ public class GoogleHadoopFileSystemConfiguration {
...
+          "fs.gs.storage.client.caching.enable",
+          GoogleCloudStorageOptions.DEFAULT.isStorageClientCachingEnabled());
+
+  /** Configuration key for maximum number of gRPC storage clients to cache. */
+  public static final HadoopConfigurationProperty<Integer> GCS_STORAGE_CLIENT_CACHE_MAX_SIZE =
+      new HadoopConfigurationProperty<>("fs.gs.storage.client.cache.maxSize", 10);
+
+  /** Configuration key for gRPC storage client cache expiry time. */
+  public static final HadoopConfigurationProperty<Long> GCS_STORAGE_CLIENT_CACHE_TIME =
+      new HadoopConfigurationProperty<>("fs.gs.storage.client.cache.time", 6_00_000L); // 10 minutes
```
cpriti-os: Nit: Something like TimeUnit.MINUTES.toMillis(10) makes this more readable.

cpriti-os (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
animesh-g: +1, Use seconds as other configs. We can consider using minutes as well.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -328,6 +332,7 @@ public void grpcConfiguration() {
     config.setTimeDuration("fs.gs.grpc.read.timeout", 7, MILLISECONDS);
     config.setTimeDuration("fs.gs.grpc.write.message.timeout", 25, MILLISECONDS);
     config.setTimeDuration("fs.gs.grpc.write.timeout", 20, MILLISECONDS);
+    config.setTimeDuration("fs.gs.http.connect-timeout", 10, MINUTES);
```
animesh-g: Why are we adding this config here? If it is intentional, please assert this as well?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageProvider.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Provides GCS SDK Storage object which is used to access the GCS. Also Caches the storage objects,
+ * so they can be used re-used.
+ */
+public class StorageProvider {
```
animesh-g: I think we should rename StorageProvider to StorageClientProvider for consistency, as we're adding the Client suffix to all veneer client-related code.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageJavaStorageClientCachingIntegrationTest.java
```diff
@@ -0,0 +1,61 @@
...
+      @Nullable GoogleCloudStorageOptions.Builder optionsBuilder, boolean enableCaching)
+      throws IOException {
+    if (optionsBuilder == null) {
+      optionsBuilder = getStandardOptionBuilder();
+    }
+    optionsBuilder.setStorageClientCachingEnabled(enableCaching);
+    return GoogleCloudStorageClientImpl.builder()
+        .setOptions(optionsBuilder.build())
+        .setCredentials(getCredentials())
+        .setCredentials(null)
```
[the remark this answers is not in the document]
animesh-g: Why is this comment resolved? Isn't it a valid comment?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageProvider.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+    // Decrement the reference count of the object.
+    storageClientToReferenceMap.put(storage, storageClientToReferenceMap.get(storage) - 1);
+    if (storageClientToReferenceMap.get(storage) == 0) {
+      logger.atFinest().log("close() called on storage object inside cache.");
+      StorageProviderCacheKey key = storageToCacheKeyMap.get(storage);
+      cache.invalidate(key);
+      storageToCacheKeyMap.remove(storage);
+      storageClientToReferenceMap.remove(storage);
+      closeStorage(storage.getStorage());
```
animesh-g: Here we are closing the storage objects which are present in cache but it is possible that some storage objects were evicted from cache(in cases where cache is full or expired). These objects are never closed and may cause resource leaks. Consider using cache removal listener to close these objects as well https://guava.dev/releases/21.0/api/docs/com/google/common/cache/RemovalListener.html .

animesh-g (changes requested):

animesh-g:
One more question: Why we didn't cherry pick from staging branch and then build on top of that? It is easier to maintain history and identify the original authors that way.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageProvider.java
shrutisinghania: Done.

shrutisinghania:
> One more question: Why we didn't cherry pick from staging branch and then build on top of that? It is easier to maintain history and identify the original authors that way.

Commits from the staging branch were not cherry picked because of design differences. Specifically, the caching of the delegation token provider, while beneficial for Uber, is undesirable for us due to potential credential sharing and issues with stateful/request-specific tokens.

animesh-g (commented):
I couldn't understand why we moved a new caching library. Can you add an explanation on why Guava cache is not sufficient in this use case?

shrutisinghania:
> I couldn't understand why we moved to a new caching library. Can you add an explanation on why Guava cache is not sufficient in this use case?

As suggested by @ajayky-os  caffeine provided superior performance over guava with its advanced cache eviction policy

animesh-g:
> > I couldn't understand why we moved to a new caching library. Can you add an explanation on why Guava cache is not sufficient in this use case?
> 
> As suggested by @ajayky-os caffeine provided superior performance over guava with its advanced cache eviction policy

Do we have any data to back this claim up for our use case?  Removal notification is available in both so that cannot be deciding factor as per the doc comment.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageClientProvider.java
```diff
@@ -0,0 +1,328 @@
...
+          this);
+    }
+
+    if (cache == null) {
+      synchronized (this) {
+        if (cache == null) {
+          cache =
+              Caffeine.newBuilder()
+                  .maximumSize(storageOptions.getStorageClientCacheMaxSize())
+                  .expireAfterWrite(storageOptions.getStorageClientCacheExpiryTime())
```
ajayky-os, line 89: Will it evict frequently used object as well based on time ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageClientProvider.java
```diff
@@ -0,0 +1,328 @@
...
+    // Only remove the entry from the reverse map.
+    // The storage object itself will be closed via the close() method
+    // when its reference count drops to zero.
+    storageToCacheKeyMap.remove(storageWrapper);
+  }
+
+  @VisibleForTesting
+  StorageClientProviderCacheKey computeCacheKey(
+      Credentials credentials, GoogleCloudStorageOptions storageOptions) {
+    List<Object> credentialsKey = getCredentialsCacheKey(credentials);
```
ajayky-os, line 187: On place of null, return empty list and on place of null check for list is empty.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageClientProvider.java
```diff
@@ -0,0 +1,328 @@
...
+  /**
+   * Creates a stable cache key for a given Credentials object. This is necessary because many
+   * Credentials implementations do not have a stable equals/hashCode implementation, which makes
+   * them unsuitable for use as cache keys. This method extracts stable identifiers (like client
+   * email) where possible.
+   *
+   * @return A list of objects forming the cache key, or {@code null} if a stable key cannot be
+   *     created.
+   */
+  private static List<Object> getCredentialsCacheKey(Credentials credentials) {
```
ajayky-os, line 210: return ImmutableList.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageClientProvider.java
```diff
@@ -0,0 +1,328 @@
...
+      }
+      if (current instanceof com.google.auth.oauth2.UserCredentials) {
+        keyParts.add(((com.google.auth.oauth2.UserCredentials) current).getClientId());
+        break;
+      }
+      if (current instanceof com.google.auth.oauth2.ComputeEngineCredentials) {
+        keyParts.add(com.google.auth.oauth2.ComputeEngineCredentials.class);
+        break;
+      }
+      if (current instanceof com.google.auth.oauth2.ImpersonatedCredentials) {
```
ajayky-os, line 230: We are not setting cache key in this case ?

PR #1429, opened 2025-07-31
Adding support for Bidi read

Dhriti07:
Adding implementation for Bidi Read

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
@@ -29,10 +29,8 @@ public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableBy
   private long objectSize;
   private boolean isOpen = true;
   private boolean gzipEncoded = false;
-  @VisibleForTesting private SeekableByteChannel contentReadChannel;
-  @VisibleForTesting private long currentPosition = 0;
-
-  private boolean isProjectionObtained= false;
+  @VisibleForTesting public SeekableByteChannel contentReadChannel;
+  @VisibleForTesting public long currentPosition = 0;
 
   public GoogleCloudStorageBidiReadChannel(
       Storage storage,
@@ -49,7 +47,8 @@ public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableBy
     this.blobReadSession =
         initializeBlobReadSession(storage, blobId, readOptions.getBidiClientTimeout());
     this.boundedThreadPool = boundedThreadPool;
-    initMetadata(itemInfo.getContentEncoding());
+    initMetadata(itemInfo.getContentEncoding(), itemInfo.getSize());
+    initializeReadSession();
   }
 
   private void initializeReadSession() {
@@ -69,10 +68,6 @@ public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableBy
 
   @Override
   public int read(ByteBuffer dst) throws IOException {
-    if(!isProjectionObtained){
-      isProjectionObtained = true;
-      initializeReadSession();
-    }
     throwIfNotOpen();
     if (!dst.hasRemaining()) {
       return 0;
@@ -87,6 +82,7 @@ public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableBy
       this.currentPosition = objectSize;
       return -1;
     }
+    currentPosition = currentPosition + bytesRead;
 
     return bytesRead;
   }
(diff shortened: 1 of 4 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,51 +1,60 @@
 package com.google.cloud.hadoop.gcsio;
```
ShreyasSinha, line 1: Add finer logging across the implementation for major decisions. This will help in debugging. Refer: GoogleCloudStorageClientReadChannel for major logging spots.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -70,37 +94,61 @@ public int write(ByteBuffer src) throws IOException {
...
+    this.currentPosition = newPosition;
+    return this;
   }
 
   @Override
   public long size() throws IOException {
-    // TODO(dhritichopra) Add read flow
-    return 0;
+    throwIfNotOpen();
+    if (objectSize == -1) {
```
ShreyasSinha, line 138: How is this being set to -1?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -70,37 +94,61 @@ public int write(ByteBuffer src) throws IOException {
...
+      throw new IOException("Size of file is not available");
+    }
+    return objectSize;
   }
 
   @Override
   public SeekableByteChannel truncate(long size) throws IOException {
-    // TODO(dhritichopra) Add read flow
-    return null;
+    throw new UnsupportedOperationException("Cannot truncate a read-only channel");
```
ShreyasSinha, line 148: GoogleCloudStorageEventBus.postOnException() is missing?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: It will not be set to -1 in the BidiReadChannel. Item info (From where we are populating objectSize) itself defaults to -1 if the object does not exist. I had added this for debugging. Although this case is probably unlikely that the object would not exist and size() function would be called on this channel without having failed at the initialization point. So we can consider removing this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done. Added in the following commits. Please ignore the debugging code that got pushed in the commit. I fixed it in a following commit.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
```diff
@@ -72,17 +146,10 @@ public void write_unsupportedOperationException() throws IOException {
         () -> bidiReadChannel.write(ByteBuffer.allocateDirect(0)));
   }
 
-  // TODO(dhritichopra) Dummy test to complete coverage, remove with actual test once we have these
-  // implemented.
   @Test
-  public void readPath_dummyImplementation() throws IOException {
+  public void truncaate_unsupportedOperationException() throws IOException {
```
shrutisinghania, line 150: There is a typo in the test name. It should be truncate_unsupportedOperationException

shrutisinghania (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,106 +1,170 @@
...
-  private ExecutorService boundedThreadPool;
+  private final ExecutorService boundedThreadPool;
+  private static final String GZIP_ENCODING = "gzip";
+  private long objectSize;
+  private boolean isOpen = true;
+  private boolean gzipEncoded = false;
+  private final StorageResourceId resourceId;
+
+  private final Duration readTimeout;
+  @VisibleForTesting public SeekableByteChannel contentReadChannel;
```
cpriti-os, line 39: Can this be package private?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,106 +1,170 @@
...
+    int bytesRead = contentReadChannel.read(dst);
+
+    while (bytesRead == 0) {
+      if (System.nanoTime() - startTime >= timeoutNanos) {
+        GoogleCloudStorageEventBus.postOnException();
+        throw new IOException(
+            String.format(
+                "Read timeout on %s after %s. No data was received.", resourceId, readTimeout));
+      }
+      bytesRead = contentReadChannel.read(dst);
```
cpriti-os, line 103: Nit: This loop is a form of busy-waiting, should we add a sleep (maybe in ms) to yield the CPU?

Dhriti07: Since this is performance sensitive, adding any delay even a small one would lead to increased latency.

PR #1448, opened 2025-08-13
Add documentation for fs.gs.cloud.logging.enable flag

cpriti-os:
Add documentation for the fs.gs.cloud.logging.enable configuration property to the gcs/CONFIGURATION.md file.

The new documentation covers all the necessary details for a user to enable this feature, including:

* The purpose of the flag to enable log exporting.
* The required IAM permission (roles/logging.logWriter) that the service account must have for the feature to work.
* The naming convention for the logs, explaining that they will be named gcs-connector and can be suffixed with the value of fs.gs.application.name.suffix if it is set.

### DIFF
--- a/gcs/CONFIGURATION.md
+++ b/gcs/CONFIGURATION.md
@@ -55,6 +55,14 @@
     return `null` for that file. Supported checksum types are `NONE`, `CRC32C`
     and `MD5`
 
+*   `fs.gs.cloud.logging.enable` (default: `false`)
+
+    If `true`, exports Cloud Storage Connector logs to GCP Logging service.The service account used by the connector
+    must have `roles/logging.logWriter` IAM permission in the GCP project for this feature to work.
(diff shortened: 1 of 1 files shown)


gcs/CONFIGURATION.md
```diff
@@ -55,6 +55,17 @@
...
     and `MD5`
 
+*   `fs.gs.cloud.logging.enable` (default: `false`)
+
+    If `true`, exports Cloud Storage Connector logs to the GCP Logging service. The service account
+    used by the connector must have `roles/logging.logWriter` IAM permission in the GCP project
+    for this feature to work.
+
+    The logs will be labelled with the log name `gcs-connector`, which will be suffixed with the
+    value of `fs.gs.application.name.suffix` if set. For example, if
```
animesh-g, line 65: I think it would be good update description of `fs.gs.application.name.suffix` as well.

cpriti-os: Added

PR #1456, opened 2025-08-19
[HNS Optimization Implementation] Add createFolder and getFolderInfo methods 

dheerajsngh:
* Added configiration flag for HNS optimizations 
* Added createFolder method
* Added getFolderInfo method 
* Added relevant unit and integration tests

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -571,6 +571,13 @@ public class GoogleHadoopFileSystemConfiguration {
           "fs.gs.write.rolling.checksum.enable",
           AsyncWriteChannelOptions.DEFAULT.isRollingChecksumEnabled());
 
+
+  /** Configuration key to enable Hierarchical Namespace (HNS) optimizations. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_HNS_OPTIMIZATION_ENABLE_KEY =
+      new HadoopConfigurationProperty<>(
+          "fs.gs.hierarchical.namespace.folders.optimization.enable",
+          GoogleCloudStorageOptions.DEFAULT.isHnOptimizationEnabled());
+
   /** Configuration key for enabling move operation in gcs instead of copy+delete. */
   public static final HadoopConfigurationProperty<Boolean> GCS_OPERATION_MOVE_ENABLE =
       new HadoopConfigurationProperty<>(
@@ -659,7 +666,8 @@ public class GoogleHadoopFileSystemConfiguration {
         .setOperationTraceLogEnabled(GCS_OPERATION_TRACE_LOG_ENABLE.get(config, config::getBoolean))
         .setTrafficDirectorEnabled(GCS_GRPC_TRAFFICDIRECTOR_ENABLE.get(config, config::getBoolean))
         .setWriteChannelOptions(getWriteChannelOptions(config))
-        .setMoveOperationEnabled(GCS_OPERATION_MOVE_ENABLE.get(config, config::getBoolean));
+        .setMoveOperationEnabled(GCS_OPERATION_MOVE_ENABLE.get(config, config::getBoolean))
+        .setHnOptimizationEnabled(GCS_HNS_OPTIMIZATION_ENABLE_KEY.get(config, config::getBoolean));
   }
 
   @VisibleForTesting
--- a/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
+++ b/gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
@@ -83,6 +83,7 @@ public class GoogleHadoopFileSystemConfigurationTest {
           put("fs.gs.grpc.write.message.timeout", 3_000L);
           put("fs.gs.grpc.write.enable", false);
           put("fs.gs.hierarchical.namespace.folders.enable", false);
+          put("fs.gs.hierarchical.namespace.folders.optimization.enable", false);
           put("fs.gs.grpc.write.timeout", 600_000L);
           put("fs.gs.http.connect-timeout", 5_000L);
           put("fs.gs.http.max.retry", 10);
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageOptions.java
@@ -67,7 +67,8 @@ public abstract class GoogleCloudStorageOptions {
         .setWriteChannelOptions(AsyncWriteChannelOptions.DEFAULT)
         .setHnBucketRenameEnabled(false)
         .setGrpcWriteEnabled(false)
-        .setMoveOperationEnabled(false);
+        .setMoveOperationEnabled(false)
+        .setHnOptimizationEnabled(false);
   }
 
   public abstract Builder toBuilder();
@@ -147,6 +148,8 @@ public abstract class GoogleCloudStorageOptions {
   public abstract boolean isOperationTraceLogEnabled();
 
   public abstract boolean isMoveOperationEnabled();
+
+  public abstract boolean isHnOptimizationEnabled();
 
   public RetryHttpInitializerOptions toRetryHttpInitializerOptions() {
     return RetryHttpInitializerOptions.builder()
@@ -236,6 +239,8 @@ public abstract class GoogleCloudStorageOptions {
     public abstract Builder setGrpcWriteEnabled(boolean grpcWriteEnabled);
 
     public abstract Builder setMoveOperationEnabled(boolean moveOperationEnabled);
+
+    public abstract Builder setHnOptimizationEnabled(boolean hnOptimizationEnabled);
 
     abstract GoogleCloudStorageOptions autoBuild();
 


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -571,6 +571,12 @@ public class GoogleHadoopFileSystemConfiguration {
           "fs.gs.write.rolling.checksum.enable",
           AsyncWriteChannelOptions.DEFAULT.isRollingChecksumEnabled());
 
+  /** Configuration key to enable Hierarchical Namespace (HNS) optimizations. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_HNS_OPTIMIZATION_ENABLE =
+      new HadoopConfigurationProperty<>(
+          "fs.gs.hierarchical.namespace.folders.optimization.enable",
```
animesh-g, line 593: Not necessarily in this PR but please update the documentation as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -234,6 +235,14 @@ public void createBucket(String bucketName, CreateBucketOptions options) throws
       bucketInfoBuilder.setStorageClass(
           StorageClass.valueOfStrict(options.getStorageClass().toUpperCase()));
     }
+    if (options.getHierarchicalNamespaceEnabled()) {
```
animesh-g, line 246: I think this change should be part of separate PR. It is actually a bug fix instead of a new feature.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -546,13 +548,25 @@ private void mkdirsInternal(StorageResourceId resourceId) throws IOException {
 
     // Create only a leaf directory because subdirectories will be inferred
     // if leaf directory exists
+    final boolean isHns =
+        this.options.getCloudStorageOptions().isHnOptimizationEnabled() && gcs.isHnBucket(path);
```
animesh-g: Consider moving these two checks to a simple utility.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -1017,10 +1031,19 @@ public List<FileInfo> listFileInfo(URI path, ListFileOptions listOptions) throws
   @Override
   public FileInfo getFileInfo(URI path) throws IOException {
     checkArgument(path != null, "path must not be null");
+
+    StorageResourceId resourceId = StorageResourceId.fromUriPath(path, true);
+    if (this.options.getCloudStorageOptions().isHnOptimizationEnabled()
+        && this.gcs.isHnBucket(path)) {
```
animesh-g: We are checking only HNS optimization flag but not the current flag here. We should check for both. We can move this into a utility method as suggested above and have a check that optimization flag are only applicable if current flag is set as true.

Otherwise we will go into the situations where mkdir creates HNS folder but 'delete' operation doesn't delete them.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -1017,10 +1031,19 @@ public List<FileInfo> listFileInfo(URI path, ListFileOptions listOptions) throws
...
+      FileInfo fileInfo = FileInfo.fromItemInfo(gcs.getFolderInfo(resourceId.toDirectoryId()));
```
animesh-g, line 1040: Shouldn't we call getFolderInfo for only directories? https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/5dcb0407ab56c7e3833fa527f01bf0aa8a632464/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfo.java#L335

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
```diff
@@ -1178,8 +1201,14 @@ public void mkdir(URI path) throws IOException {
     // Ensure that the path looks like a directory path.
     resourceId = resourceId.toDirectoryId();
 
-    // Not a top-level directory, create 0 sized object.
-    gcs.createEmptyObject(resourceId);
+    // Check if HNS optimization is enabled, it's an HNS bucket
+    if (this.options.getCloudStorageOptions().isHnOptimizationEnabled() && gcs.isHnBucket(path)) {
```
animesh-g: Same as above. Use a util which adds check for current flag. Applies to other places as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
```diff
@@ -2433,6 +2434,74 @@ private StorageControlClient lazyGetStorageControlClient() throws IOException {
...
+   * See {@link GoogleCloudStorage#getFolderInfo(StorageResourceId)} for details about expected
+   * behavior.
+   */
+  @Override
+  public GoogleCloudStorageItemInfo getFolderInfo(StorageResourceId resourceId) throws IOException {
+    logger.atInfo().log("getFolderInfo(%s)", resourceId);
+
+    if (resourceId.isBucket()) {
+      // This is not a folder. Do not proceed.
+      // Throwing a NotFoundException is the appropriate GCS client library pattern.
```
animesh-g: Why is this the appropriate response? Don't we create createItemInfoForBucket() today for getItemInfo on bucket?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfo.java
```diff
@@ -123,6 +123,45 @@ public static GoogleCloudStorageItemInfo createObject(
...
+    long creationTime =
+        folder.hasCreateTime()
+            ? (folder.getCreateTime().getSeconds() * 1000)
+                + (folder.getCreateTime().getNanos() / 1_000_000)
+            : 0;
+    long modificationTime =
+        folder.hasUpdateTime()
+            ? (folder.getUpdateTime().getSeconds() * 1000)
+                + (folder.getUpdateTime().getNanos() / 1_000_000)
+            : 0;
```
animesh-g: I think we can use `toMillis ` from `com.google.protobuf.util.Timestamps` to avoid this custom implementation.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfo.java
```diff
@@ -331,6 +370,16 @@ public boolean isGlobalRoot() {
     return isRoot() && exists();
   }
 
+  /**
+   * Indicates whether {@code itemInfo} is a native HNS folder. This is different from a placeholder
+   * object or an inferred directory.
+   */
+  public boolean isNativeFolder() {
```
animesh-g: nit: can we call it isNativeHNSFolder to make clearer?

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfoTest.java
```diff
@@ -131,4 +133,103 @@ public void testEquals() throws Exception {
         .addEqualityGroup(GoogleCloudStorageItemInfo.createNotFound(testDirectory))
         .testEquals();
   }
+
+  @Test
+  public void testCreateFolderEquals() {
+    StorageResourceId testFolder =
+        StorageResourceId.fromStringPath("gs://test-bucket/test-folder/");
+    long createTimeMillis = 1672531200000L;
+    long updateTimeMillis = 1672617600000L;
```
animesh-g: nit: Although it is only for testing, but please add a comment on what time it actually represents to understand better. e.g.

    long createTimeMillis = 1672531200000L; //Jan 01 2023 00:00:00

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfoTest.java
```diff
@@ -131,4 +133,103 @@ public void testEquals() throws Exception {
...
+    long metaGeneration = 5L;
+
+    com.google.protobuf.Timestamp createTime =
+        com.google.protobuf.Timestamp.newBuilder().setSeconds(createTimeMillis / 1000).build();
```
animesh-g: Why don't we declare time in seconds instead of converting it evreywhere?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
animesh-g, line 1: The changes in this file doesn't seem to be unit-tested. Please add the same.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfo.java
dheerajsngh: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfo.java
dheerajsngh: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
dheerajsngh: Actually we have not implemented any of the HNS optimization for `gRPC` client.  Its only implemented in JSON client. 
The only change we have done in this file, is to update the `createBucket` function to also created an HNS Bucket, Will add UT for it.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfoTest.java
dheerajsngh: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageItemInfoTest.java
dheerajsngh: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
dheerajsngh: During testing we observed that from the input, "/" removed, so doing directory check is not reliable, hence directly calling `getFolderInfo()`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
dheerajsngh: Actually this is needed for the tests where we are testing HNS features using the gRPC client.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
dheerajsngh: Ack.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
dheerajsngh: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
dheerajsngh: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
dheerajsngh: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageImpl.java
dheerajsngh: Makes sense.  
We should return `createItemInfoForBucket()` for bucket similar to `getItemInfo`. 

Updated. Thanks for pointing out.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageFileSystemImpl.java
animesh-g: Can you write a comment to clarify the same?

dheerajsngh: Added a comment. 
Thanks.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
```diff
@@ -326,6 +327,60 @@ public void deleteFolders(List<FolderInfo> folders) throws IOException {
...
+      throw new FileAlreadyExistsException("Folder or object '" + resourceId + "' already exists");
+    }
+
+    // Simulate the Folder object that the real API would return.
+    long now = clock.currentTimeMillis();
+    Timestamp timestamp =
+        Timestamp.newBuilder()
+            .setSeconds(now / 1000)
+            .setNanos((int) ((now % 1000) * 1_000_000))
+            .build();
```
animesh-g: Use `Timestamp timestamp = Timestamps.fromMillis(now);` to simplify this.

ajayky-os (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/testing/InMemoryGoogleCloudStorage.java
dheerajsngh: Done.

PR #1457, opened 2025-08-19
Prevent infinite loop in read channels due to connection loss

cpriti-os:
This PR addresses a bug where a lost connection during a read operation could cause an infinite loop in GoogleCloudStorageReadChannel and GoogleCloudStorageClientReadChannel.

The problem: The underlying contentStream.skip() can return 0 bytes if the connection to the GCS server is lost while reading a response causing the while loop to spin indefinitely, blocking customer processes.
A similar issue could occur in the gRPC read path.

The solution: This change introduces checks to detect when skip() or read() returns 0.

* In GoogleCloudStorageReadChannel.java, if contentStream.skip(bytesToSkip) returns a value <= 0, an IOException is thrown to prevent the infinite loop
* In GoogleCloudStorageClientReadChannel.java, if byteChannel.read(...) returns 0, an IOException is now thrown
Related bug: https://b.corp.google.com/issues/438790622

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -484,6 +484,13 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
         try {
           int bufferSize = toIntExact(min(skipBuffer.length, seekDistance));
           int bytesRead = byteChannel.read(ByteBuffer.wrap(skipBuffer, 0, bufferSize));
+          if (bytesRead == 0) {
+            GoogleCloudStorageEventBus.postOnException();
+            throw new IOException(
+                String.format(
+                    "Read 0 bytes without blocking while skipping %d bytes from object: '%s'",
+                    seekDistance, resourceId));
+          }
           if (bytesRead < 0) {
             logger.atInfo().log(
                 "Somehow read %d bytes trying to skip %d bytes to seek to position %d, size: %d",
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -1056,6 +1056,10 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
             bytesToSkip, contentChannelPosition, currentPosition, resourceId);
         while (bytesToSkip > 0) {
           long skippedBytes = contentStream.skip(bytesToSkip);
+          if (skippedBytes <= 0) {
+            throw new IOException(
+                "contentStream.skip() returned " + skippedBytes + ", potential connection loss");
+          }
           logger.atFiner().log(
               "Skipped %d bytes from %d position for '%s'",
               skippedBytes, contentChannelPosition, resourceId);
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -1060,7 +1060,9 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
             GoogleCloudStorageEventBus.postOnException();
             // If we skipped 0 bytes, it means that the connection was lost.
             throw new IOException(
-                "contentStream.skip() returned " + skippedBytes + ", potential connection loss");
+                String.format(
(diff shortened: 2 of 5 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1056,6 +1056,12 @@ protected InputStream openStream(long bytesToRead) throws IOException {
             bytesToSkip, contentChannelPosition, currentPosition, resourceId);
         while (bytesToSkip > 0) {
           long skippedBytes = contentStream.skip(bytesToSkip);
+          if (skippedBytes <= 0) {
+            GoogleCloudStorageEventBus.postOnException();
+            // If we skipped 0 bytes, it means that the connection was lost.
+            throw new IOException(
+                "contentStream.skip() returned " + skippedBytes + ", potential connection loss");
```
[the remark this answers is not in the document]
cpriti-os: done

ajayky-os (commented):
Write unit tests for the changes.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1056,6 +1056,14 @@ protected InputStream openStream(long bytesToRead) throws IOException {
             bytesToSkip, contentChannelPosition, currentPosition, resourceId);
         while (bytesToSkip > 0) {
           long skippedBytes = contentStream.skip(bytesToSkip);
+          if (skippedBytes <= 0) {
+            GoogleCloudStorageEventBus.postOnException();
+            // If we skipped 0 bytes, it means that the connection was lost.
+            throw new IOException(
+                String.format(
+                    "contentStream.skip() returned %d, potential connection loss for object: '%s'",
+                    skippedBytes, resourceId));
```
animesh-g, line 1065: nit: I think it might be worth adding pending bytesToSkip in this message.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -484,6 +484,13 @@ private void skipInPlace() {
         try {
           int bufferSize = toIntExact(min(skipBuffer.length, seekDistance));
           int bytesRead = byteChannel.read(ByteBuffer.wrap(skipBuffer, 0, bufferSize));
+          if (bytesRead == 0) {
```
animesh-g, line 487: Isn't it a valid case? See https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ReadableByteChannel.html

It says, 
`
A read operation might not fill the buffer, and in fact it might not read any bytes at all. Whether or not it does so depends upon the nature and state of the channel.
`

I think it is valid case and we shouldn't throw exception in this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1056,6 +1056,14 @@ protected InputStream openStream(long bytesToRead) throws IOException {
             bytesToSkip, contentChannelPosition, currentPosition, resourceId);
         while (bytesToSkip > 0) {
           long skippedBytes = contentStream.skip(bytesToSkip);
+          if (skippedBytes <= 0) {
```
animesh-g, line 1059: Similar to above... is it possible that bytes are not arrived yet in this stream to skip, given it is a non-blocking stream?

cpriti-os: Valid point, it's getting tricky. Ajay pointed out that the ideal solution would be if the underlying channel threw a proper error in case of connection loss which would be worth investigating and the best long term fix.
In the meantime I am thinking how we can avoid this infinite loop case: one solution is retry with small delays --
If it returns 0, this indicates a lack of progress. Instead of looping immediately:
Increment a "no-progress" counter.
If the counter is below a defined threshold (e.g., 5 to 10 attempts), pause for a short, fixed duration (e.g., 10ms) before retrying.
If the counter exceeds the threshold, it's reasonable to assume the connection is permanently stalled. At this point, throw an IOException to break the loop. 
At the first instance of positive return, reset the counter. What do you suggest?

animesh-g: I think this solution will still be very hacky. I think it would be better

- Work with user and check if there is any case where `bytesToSkip` is calculated incorrectly? Check how Veneer client handle this scenario while opening the channel as this codeblock doesn't exist in GoogleCloudStorageClientReadChannel.java .
- Since it is in Apiary, we probably cannot fix it in the underlying framework as it is autogenerated. Maybe ask customer to use STORAGE_CLIENT which provides better error handling.

PR #1458, opened 2025-08-19
 Fix: Gracefully close cloud logger in GoogleHadoopFileSystem

cpriti-os:
This PR introduces the following changes:

The LoggingInterceptor instance is now stored in a member variable upon initialization.
The close() method of GoogleHadoopFileSystem has been updated to:
1. Remove the loggingInterceptor from the root logger.
2. Call the close() method on the loggingInterceptor to release its resources.
3. Set the loggingInterceptor reference to null.
4. Error handling has been added to catch and log any exceptions that occur during the logging service shutdown.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -557,30 +557,14 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
   }
 
   private void skipInPlace(long seekDistance) {
-    if (skipBuffer == null) {
-      skipBuffer = new byte[SKIP_BUFFER_SIZE];
-    }
-    while (seekDistance > 0 && contentChannel != null) {
-      try {
-        int bufferSize = toIntExact(min(skipBuffer.length, seekDistance));
-        int bytesRead = contentChannel.read(ByteBuffer.wrap(skipBuffer, 0, bufferSize));
-        if (bytesRead < 0) {
-          // Shouldn't happen since we called validatePosition prior to this loop.
-          logger.atInfo().log(
-              "Somehow read %d bytes trying to skip %d bytes to seek to position %d, size: %d",
-              bytesRead, seekDistance, currentPosition, size);
-          closeContentChannel();
-        } else {
-          seekDistance -= bytesRead;
-          contentChannelPosition += bytesRead;
-        }
-      } catch (IOException e) {
-        GoogleCloudStorageEventBus.postOnException();
-        logger.atInfo().withCause(e).log(
-            "Got an IO exception on contentChannel.read(), a lazy-seek will be pending for '%s'",
-            resourceId);
-        closeContentChannel();
-      }
+    try {
+      contentChannelPosition += skip(contentChannel, seekDistance);
+    } catch (IOException e) {
+      GoogleCloudStorageEventBus.postOnException();
+      logger.atInfo().withCause(e).log(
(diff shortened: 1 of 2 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -1815,6 +1817,17 @@ public void close() throws IOException {
...
 
+    if (loggingInterceptor != null) {
+      try {
+        Logger.getLogger("").removeHandler(loggingInterceptor);
+        loggingInterceptor.close();
+        loggingInterceptor = null;
+      } catch (RuntimeException e) {
+        logger.atSevere().withCause(e).log("Failed to stop cloud logging service");
+      }
+      loggingInterceptor = null;
```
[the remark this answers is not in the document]
cpriti-os: done

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -322,6 +333,114 @@ public void testTotalTimeStatistics() throws IOException {
...
+      assertThat(Arrays.asList(rootLogger.getHandlers())).doesNotContain(mockInterceptor);
+
+      // Verify exception was logged
+      testLogHandler.flush();
+      assertThat(logOutput.toString()).contains("Failed to stop cloud logging service");
+    } finally {
+      // Ensure handler is removed even if test fails, to not affect other tests
+      Logger.getLogger("").removeHandler(mockInterceptor);
+      ghfsLogger.removeHandler(testLogHandler);
+      ghfsLogger.setLevel(Level.OFF);
```
animesh-g: Optional: Isn't it possible that any test fails and misses to remove handlers? Should we move this cleanup in `@After`? It is probably ok for now as this feature is behind a feature flag but maybe worth considering.

animesh-g (approved):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
cpriti-os: Done.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -322,6 +354,107 @@ public void testTotalTimeStatistics() throws IOException {
...
+      verify(mockInterceptor, times(1)).close();
+      assertThat(Arrays.asList(rootLogger.getHandlers())).doesNotContain(mockInterceptor);
+    } finally {
+      // Ensure handler is removed even if test fails, to not affect other tests
+      Logger.getLogger("").removeHandler(mockInterceptor);
+    }
+  }
+
+  @Test
+  public void loggingInterceptorCloseFails() throws Exception {
```
ajayky-os: Consider including the expected outcome of the test in the test name. Same for other specs.

One common pattern of test names is as follows:

```
<subjectUnderTest>_<condition>_<expectedOutcome>
<behaviorAcrossManyMethods>_<condition1>_<condition2>_<expectedOutcome>
```

> A test's name should fully describe what a test does, including not only the state and details of the method being tested but also its expected outcome.

cpriti-os: Done

PR #1467, opened 2025-08-25
Adding Read functionality for bidi Read channel

Dhriti07:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,106 +1,178 @@
...
+      DisposableByteString disposableByteString =
+          futureBytes.get(readTimeout.toNanos(), TimeUnit.NANOSECONDS);
+
+      int bytesRead;
+      try (DisposableByteString dbs = disposableByteString) {
+        ByteString byteString = dbs.byteString();
+        bytesRead = byteString.size();
+
+        if (bytesRead == 0 && position < objectSize) {
+          throw new IOException(
```
ShreyasSinha, line 109: Is there any documentation why this is an error scenario?

Dhriti07: As Discussed offline, this check is added to make sure read does not return 0 bytes when the object still has not reached its size limit

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,106 +1,178 @@
...
+              String.format("Read 0 bytes without blocking from object: '%s'", resourceId));
+        }
+
+        if (bytesRead > 0) {
+          byteString.copyTo(dst);
+          position += bytesRead;
+        }
+      }
+
+      return bytesRead > 0 ? bytesRead : -1;
```
ShreyasSinha: Can you replace -1 with a Integer variable with an appropriate name like "Unread Bytes" so that its more readable. Replace -1 everywhere in the file and test so that its clear you are evaluating this condition.

Dhriti07: How about adding a comment here instead explaining the reason behind the -1 return value instead ?
The test case already explains what path it is testing and the expected value, and this variable/constant is not used at multiple places across the channel.

PR #1475, opened 2025-09-01
Storage Client Caching for downscoped token

shrutisinghania:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -213,7 +216,7 @@ public WritableByteChannel create(StorageResourceId resourceId, CreateObjectOpti
     }
 
     return new GoogleCloudStorageClientWriteChannel(
-        storage, storageOptions, resourceIdWithGeneration, options);
+        storage.getStorage(), storageOptions, resourceIdWithGeneration, options);
```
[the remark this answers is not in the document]
ajayky-os: Please look at this suggestion for implementation of `StorageClientWrapper`

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -121,7 +120,9 @@ public class GoogleCloudStorageClientImpl extends ForwardingGoogleCloudStorage {
   private static final int MAXIMUM_PRECONDITION_FAILURES_IN_DELETE = 4;
 
   private final GoogleCloudStorageOptions storageOptions;
-  private final Storage storage;
+  @VisibleForTesting final StorageClientWrapper storage;
```
animesh-g: nit: Can we call it something like `storageWrapper` instead of `storage` because `storage.getStorage()` seems less intuitive to me?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
```diff
@@ -137,7 +137,7 @@ public class GoogleHadoopFileSystemConfigurationTest {
           put("fs.gs.bidi.thread.count", 16);
           put("fs.gs.bidi.client.timeout", 30);
           put("fs.gs.bidi.finalize.on.close", false);
-          put("fs.gs.client.caching.experiment.enabled", false);
+          put("fs.gs.storage.client.cache.enable", false);
```
animesh-g, line 140: Open question: Should we have a fallback on `fs.gs.client.caching.experiment.enabled` flag as well so that users of staging branch don't have to update it in all their jobs?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shrutisinghania: Done.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfigurationTest.java
shrutisinghania: IMO, instead of having a fallback, the flag in the staging branch should be updated. The word "experiment" in `fs.gs.client.caching.experiment.enabled` implies that the feature is not stable and is subject to change. 
Although, this will require a one-time update for users on the staging branch, it will avoid confusion for new users.

PR #1479, opened 2025-09-01
Fix: Preserve Existing Log Formatting When Adding Invocation ID

cpriti-os:
This PR addresses an issue where the introduction of invocation ID prefixing in logs inadvertently overwrote existing log formats, leading to test failures as reported in [b/442341898](https://b.corp.google.com/issues/442341898). The LoggingFormatter was previously replacing the entire log format.

The key changes in this PR are:

* Decorator Pattern for LoggingFormatter: The LoggingFormatter has been refactored to wrap the existing Formatter instance present on each Handler. This ensures that any user-defined or application-specific log formatting is preserved.
* Improved Tests: Unit tests for LoggingFormatter have been updated and expanded to cover the new wrapping logic.


util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
```diff
@@ -25,24 +36,30 @@ public class LoggingFormatter extends Formatter {
...
             && !invocationId.isEmpty()
-            && !optMessage.map(this::isJson).orElse(false);
+            && !isJson(originalMessage);
 
+    String messageToFormat = originalMessage;
     if (shouldFormat) {
       // Prefix the invocation ID to the log message.
-      return String.format("[%s]: %s%n", invocationId, optMessage.orElse(""));
+      messageToFormat =
+          String.format("[%s]: %s", invocationId, originalMessage == null ? "" : originalMessage);
```
animesh-g: Nit(optional): This ternary can also be written as `Objects.toString(originalMessage, "")` from java.util.Objects;

Applies to the other ternary as well.

animesh-g (approved):

util/src/main/java/com/google/cloud/hadoop/util/LoggingFormatter.java
cpriti-os: Done

util/src/test/java/com/google/cloud/hadoop/util/LoggingFormatterTest.java
```diff
@@ -95,4 +102,41 @@ public void addFormatter_toLogger_addsFormatterToAllHandlers() {
...
+          @Override
+          public String format(LogRecord record) {
+            return record.getMessage();
+          }
+        });
+
+    // First call should wrap the formatter.
+    LoggingFormatter.addFormatterToHandler(handler);
+    Formatter formatter1 = handler.getFormatter();
+    assertTrue(formatter1 instanceof LoggingFormatter);
```
ajayky-os: nit: group it with assert at end of spec.

cpriti-os: Done

ajayky-os (approved):

Issue #1180, opened 2024-06-14
Bug in `GoogleCloudStorageReadChannel` can cause an infinite loop


rathinb-db:
In https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/b0fff1c6e45e6c0bd74fc7ce228a9f86da439ce8/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L1054, `contentStream.skip` can return 0 bytes. 

See the commit message of 
[https://github.com/googleapis/google-http-java-client/commit/8ae9cef3d8ea548ff2c55825dc89e109ec908a47](https://github.com/googleapis/google-http-java-client/commit/8ae9cef3d8ea548ff2c55825dc89e109ec908a47), the JDK's `HttpUrlConnection` fails to notify clients when the connection is lost when reading a response (eg. a server gets restarted).

When this happens during a `skip`, we spin indefinitely. The fix is simple, detect this condition and throw an error.

cpriti-os:
The fix for this issue is now available in versions: 3.1.7, 3.0.11 and 2.2.29

PR #1480, opened 2025-09-01
Fix infinite loop issue when EOF is reached during a skip operation

cpriti-os:
This PR addresses a critical bug in GoogleCloudStorageReadChannel where a skip() operation on the content stream could lead to an infinite loop.

Problem:

As reported in [Issue](https://github.com/GoogleCloudDataproc/hadoop-connectors/issues/1180), the contentStream.skip() method can return 0 bytes. On further investigation it was found that this skip method call returns 0 even when EOF is reached instead of a negative value. This behavior causes an infinite loop as the code repeatedly attempts to skip bytes that lead to EOF, effectively stalling the process.

Solution:

To resolve this, the call to contentStream.skip() has been replaced with contentStream.read() into a skipBuffer. This approach allows for a more reliable detection of the end of the stream. If the stream ends prematurely while attempting to skip bytes (indicated by read() returning a negative value), an EOFException is now thrown. This change ensures that the loop is broken and an error is properly propagated.

Was able to reproduce this manually and verified  proposed changes solves this issue.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -1054,13 +1054,20 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
         logger.atFiner().log(
             "Skipping %d bytes from %d position to %d position for '%s'",
             bytesToSkip, contentChannelPosition, currentPosition, resourceId);
+        if (skipBuffer == null) {
+          skipBuffer = new byte[SKIP_BUFFER_SIZE];
+        }
         while (bytesToSkip > 0) {
-          long skippedBytes = contentStream.skip(bytesToSkip);
-          logger.atFiner().log(
-              "Skipped %d bytes from %d position for '%s'",
-              skippedBytes, contentChannelPosition, resourceId);
-          bytesToSkip -= skippedBytes;
-          contentChannelPosition += skippedBytes;
+          int bytesRead =
+              contentStream.read(skipBuffer, 0, toIntExact(min(bytesToSkip, skipBuffer.length)));
+          if (bytesRead < 0) {
+            throw new EOFException(
+                String.format(
+                    "Unexpected end of stream trying to skip %d bytes to seek to position %d, size: %d for '%s'",
+                    bytesToSkip, currentPosition, size, resourceId));
+          }
+          bytesToSkip -= bytesRead;
+          contentChannelPosition += bytesRead;
         }
       }
 
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -557,30 +557,14 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
   }
 
   private void skipInPlace(long seekDistance) {
-    if (skipBuffer == null) {
-      skipBuffer = new byte[SKIP_BUFFER_SIZE];
-    }
-    while (seekDistance > 0 && contentChannel != null) {
-      try {
-        int bufferSize = toIntExact(min(skipBuffer.length, seekDistance));
-        int bytesRead = contentChannel.read(ByteBuffer.wrap(skipBuffer, 0, bufferSize));
-        if (bytesRead < 0) {
-          // Shouldn't happen since we called validatePosition prior to this loop.
-          logger.atInfo().log(
-              "Somehow read %d bytes trying to skip %d bytes to seek to position %d, size: %d",
-              bytesRead, seekDistance, currentPosition, size);
-          closeContentChannel();
-        } else {
-          seekDistance -= bytesRead;
-          contentChannelPosition += bytesRead;
-        }
-      } catch (IOException e) {
-        GoogleCloudStorageEventBus.postOnException();
-        logger.atInfo().withCause(e).log(
-            "Got an IO exception on contentChannel.read(), a lazy-seek will be pending for '%s'",
-            resourceId);
-        closeContentChannel();
-      }
+    try {
+      contentChannelPosition += skip(contentChannel, seekDistance);
+    } catch (IOException e) {
+      GoogleCloudStorageEventBus.postOnException();
+      logger.atInfo().withCause(e).log(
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -1039,13 +1039,20 @@ protected InputStream openStream(long bytesToRead) throws IOException {
...
+          skipBuffer = new byte[SKIP_BUFFER_SIZE];
+        }
         while (bytesToSkip > 0) {
-          long skippedBytes = contentStream.skip(bytesToSkip);
-          logger.atFiner().log(
-              "Skipped %d bytes from %d position for '%s'",
-              skippedBytes, contentChannelPosition, resourceId);
-          bytesToSkip -= skippedBytes;
-          contentChannelPosition += skippedBytes;
+          int bytesRead =
```
animesh-g: Why cannot we call skipInPlace() method directly instead of repeating the logic here? It also handles stuff like closing the channel which doesn't happen here.

cpriti-os: Tried that, the existing tests failed which indicated something might be amiss

cpriti-os: skipInPlace has a few additional lines, I tried taking out the common logic to another method but that was adding complexity of updating the parameters passed by reference

animesh-g: ok...I don't think there are any logic in the skipInPlace which is not applicable here. 

However If it is overkill, we can also consider just removing the while loop in this block which is unnecessary. Skip() method itself runs over the loop with read and exits if it finds EOF.

cpriti-os: A direct call from openStream looks incorrect because the two methods have different error handling semantics, and I thought of preserving that distinction.•openStream must propagate I/O exceptions because a failure to skip bytes means the entire stream opening operation has failed.•skipInPlace, on the other hand, is designed to handle I/O exceptions as recoverable errors during a read() operation by closing the underlying stream and allowing a retry on the next read.To address this and also remove the duplicated logic, I've refactored the code by extracting the common byte-skipping logic into a new private helper method, skip(). Both openStream and skipInPlace now call this helper, but they handle its potential IOException according to their specific requirements. This keeps the code clean and ensures the error handling remains correct for both scenarios.Regarding your question about channel closing, I, the channels are being closed at the appropriate places IIUC. The closeContentChannel() method is called in the following scenarios, which covers all the necessary cases for proper resource management:1.On normal close: When the main GoogleCloudStorageReadChannel is closed by the user.2.On read/skip error: When an IOException occurs during a read or an in-place seek, the underlying channel is closed to allow for a clean retry on the next operation.3.On out-of-range seek: When a seek is performed that requires a new HTTP range request (e.g., a backward seek or a forward seek beyond the inplaceSeekLimit), the old channel is closed before opening a new one.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -590,6 +574,27 @@ private void skipInPlace(long seekDistance) {
         currentPosition);
   }
 
+  private long skip(ReadableByteChannel channel, long bytesToSkip) throws IOException {
```
ajayky-os: Add a comment explaining the reason why we implemented skip this way for future reference.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannelTest.java
```diff
@@ -897,6 +898,43 @@ public int read() {
         .inOrder();
   }
 
+  @Test
+  public void openStream_gzipped_throwsEofException_onPrematureEofInSkip() throws IOException {
```
ajayky-os: Usually conditons are in the middle `<behaviorAcrossManyMethods>_<condition1>_<condition2>_<expectedOutcome>`

 `openStream_gzipped_onPrematureEofInSkip_throwsEofException`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -557,30 +557,14 @@ public SeekableByteChannel position(long newPosition) throws IOException {
...
-          contentChannelPosition += bytesRead;
-        }
-      } catch (IOException e) {
-        GoogleCloudStorageEventBus.postOnException();
-        logger.atInfo().withCause(e).log(
-            "Got an IO exception on contentChannel.read(), a lazy-seek will be pending for '%s'",
-            resourceId);
-        closeContentChannel();
-      }
+    try {
```
ajayky-os, line 560: Is `contentChannel != null` relevant for existing implementation ? The new implementation skips this check and will result in NPE.

ajayky-os (approved):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannelTest.java
cpriti-os: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
cpriti-os: added

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
cpriti-os: skipInPlace is only called from inside an `if` that checks if contentChannel is not null. The earlier implementation relied on that condition to break out of the loop.
To add, the added function skip() is called from skipInPlace and from openStream that creates a new channel before calling.
I have added a NonNull annotation to the channel parameter in skip() so that no other implementation in the future ends up sending null.

animesh-g (approved):

PR #1487, opened 2025-09-08
Auto-generate TrackedFeatures enum from JSON

cpriti-os:
This PR introduces a build-time mechanism to auto-generate the TrackedFeatures.java enum from a feature-map.json file.

This simplifies adding and managing feature flags by using a single source of truth, reducing manual errors and improving maintainability. Developers can now update features by modifying the JSON file directly. 

The JSON file is required to provide a machine-readable single source of truth for the job tracking feature-bit-version for all clients.


gcsio/pom.xml
```diff
@@ -206,6 +211,11 @@
       <groupId>io.grpc</groupId>
       <artifactId>grpc-testing</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
```
ajayky-os: What is use of this package ? So far gcsio module has boundary of being not dependent on any hadoop specific classes and features, all the hadoop specific business logics are in gcs module. Make sure we are not breaking this boundary.

animesh-g (commented):
Let's discuss this when you get back. I am not sure if we even need this. I think a simple enum in Java combined with unit tests will suffice here.

PR #1488, opened 2025-09-08
Implement feature usage tracking header

cpriti-os:
[Based on PR #1487 ] 
The core of this change is the new FeatureUsageHeader class which is responsible for generating the header value. The value is a Base64 encoded bitmask representing enabled features. 
The bitmask is populated from two sources:
- Configuration-based features: These are features enabled in the Hadoop configuration (e.g., fs.gs.cloud.logging.enable).
- Operation-based features: These are features used during a specific Hadoop operation. These are tracked using an InheritableThreadLocal to ensure they are cleared after the request


cpriti-os:
Created another PR, closing this.

PR #1506, opened 2025-09-11
Read functionality for Bidi Read Channel

Dhriti07:
Adding Read functionality for bidi Read channel

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
@@ -27,7 +27,7 @@ import javax.annotation.Nullable;
 public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableByteChannel {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
 
-  private static final int EOF_RETURN_VALUE = -1;
+  private static final int EOF = -1;
   private final StorageResourceId resourceId;
   private final BlobId blobId;
   private final BlobReadSession blobReadSession;
@@ -84,7 +84,7 @@ public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeek
     }
 
     if (position >= objectSize) {
-      return EOF_RETURN_VALUE;
+      return EOF;
     }
 
     logger.atFinest().log(
@@ -113,7 +113,7 @@ public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeek
         position += bytesRead;
       }
 
-      return bytesRead > 0 ? bytesRead : EOF_RETURN_VALUE;
+      return bytesRead > 0 ? bytesRead : EOF;
 
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
       GoogleCloudStorageEventBus.postOnException();
@@ -124,7 +124,6 @@ public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeek
 
   @Override
   public long position() throws IOException {
-    throwIfNotOpen();
     return position;
   }
 
@@ -150,7 +149,6 @@ public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeek
 
   @Override
   public long size() throws IOException {
-    throwIfNotOpen();
     if (objectSize == -1) {
       GoogleCloudStorageEventBus.postOnException();
       throw new IOException("Size of file is not available");
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
@@ -233,29 +233,10 @@ public class GoogleCloudStorageBidiReadChannelTest {
   }
 
   @Test
-  public void position_onClosedChannel_throwsClosedChannelException() throws IOException {
-    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel(DEFAULT_ITEM_INFO);
-
-    bidiReadChannel.close();
-
-    assertThrows(ClosedChannelException.class, () -> bidiReadChannel.position(1));
-    assertThrows(ClosedChannelException.class, bidiReadChannel::position);
-  }
-
-  @Test
   public void size_successful() throws IOException {
     GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel(DEFAULT_ITEM_INFO);
 
     assertEquals(OBJECT_SIZE, bidiReadChannel.size());
-  }
-
-  @Test
-  public void size_onClosedChannel_throwsClosedChannelException() throws IOException {
-    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel(DEFAULT_ITEM_INFO);
-
-    bidiReadChannel.close();
-
-    assertThrows(ClosedChannelException.class, bidiReadChannel::size);
   }
 
   @Test


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,53 +1,61 @@
...
 
   public GoogleCloudStorageBidiReadChannel(
       Storage storage,
       GoogleCloudStorageItemInfo itemInfo,
       GoogleCloudStorageReadOptions readOptions,
       ExecutorService boundedThreadPool)
       throws IOException {
-    this.storage = storage;
+
+    // TODO(dhritichopra) Remove grpcReadTimeout if redundant and rename to bidiReadTimeout.
```
animesh-g, line 49: Why not now?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -168,13 +241,47 @@ private long processBytesAndCompleteRange(
...
+    if (!isOpen()) {
+      GoogleCloudStorageEventBus.postOnException();
+      throw new ClosedChannelException();
+    }
+  }
+
+  protected void initMetadata(@Nullable String encoding, long sizeFromMetadata)
+      throws UnsupportedOperationException {
+    gzipEncoded = nullToEmpty(encoding).contains(GZIP_ENCODING);
+    if (gzipEncoded) {
```
animesh-g, line 259: Is GZIP support coming in further PRs? If not, I would suggest to update documentation because `fs.gs.inputstream.support.gzip.encoding.enable` sets expectation to user that gzip is supported. If we are planning to add the same, please add a TODO.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +68,120 @@ private static BlobReadSession initializeBlobReadSession(
...
     return readSession;
   }
 
   @Override
   public int read(ByteBuffer dst) throws IOException {
-    // TODO(dhritichopra) Add read flow
-    return 0;
-  }
+    throwIfNotOpen();
+    final int EOFReturnValue = -1;
```
animesh-g: nit: It should be static and can be moved before constructor along with other constants. Also use CAPITAL_CASE for this similar to other constants.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +68,120 @@ private static BlobReadSession initializeBlobReadSession(
...
+      ApiFuture<DisposableByteString> futureBytes =
+          blobReadSession.readAs(
+              ReadProjectionConfigs.asFutureByteString()
+                  .withRangeSpec(RangeSpec.of(position, bytesToRequest)));
+
+      DisposableByteString disposableByteString =
+          futureBytes.get(readTimeout.toNanos(), TimeUnit.NANOSECONDS);
+
+      int bytesRead;
+      try (DisposableByteString dbs = disposableByteString) {
```
animesh-g: Why disposableByteString is reassigned to dbs? Why don't we write this block as following,

```
try (DisposableByteString dbs = futureBytes.get(readTimeout.toNanos(), TimeUnit.NANOSECONDS)) 
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +68,120 @@ private static BlobReadSession initializeBlobReadSession(
...
+      throw new IOException(
+          String.format("Read failed on %s at position %d", resourceId, position), e);
+    }
   }
 
   @Override
   public long position() throws IOException {
-    // TODO(dhritichopra) Add read flow
-    return 0;
+    throwIfNotOpen();
```
animesh-g: nit(optional): GoogleCloudStorageClientReadChannel doesn't throw in this case but JSON channel does. Can we have a consistent behavior across all channels?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +68,120 @@ private static BlobReadSession initializeBlobReadSession(
...
+  public int write(ByteBuffer src) throws IOException {
+    GoogleCloudStorageEventBus.postOnException();
+    throw new UnsupportedOperationException("Cannot mutate read-only channel");
   }
 
   @Override
   public long size() throws IOException {
-    // TODO(dhritichopra) Add read flow
-    return 0;
+    throwIfNotOpen();
```
animesh-g: Same as position(). There should be consistent behavior across all three channels.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,53 +1,61 @@
...
-  private final Storage storage;
   private final StorageResourceId resourceId;
   private final BlobId blobId;
-  private final GoogleCloudStorageReadOptions readOptions;
   private final BlobReadSession blobReadSession;
-  private ExecutorService boundedThreadPool;
+  private final ExecutorService boundedThreadPool;
+  private static final String GZIP_ENCODING = "gzip";
+  private long objectSize;
+  private boolean isOpen = true;
```
animesh-g: nit: Let's call this variable `open` to avoid confusion with method isOpen.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +68,120 @@ private static BlobReadSession initializeBlobReadSession(
...
+    logger.atFinest().log(
+        "Seek from %d to %d position for '%s'", this.position, newPosition, resourceId);
+
+    this.position = newPosition;
+    return this;
+  }
+
+  @Override
+  public int write(ByteBuffer src) throws IOException {
+    GoogleCloudStorageEventBus.postOnException();
```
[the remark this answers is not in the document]
animesh-g: This comment is correct. We should handle the same.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
```diff
@@ -64,40 +66,205 @@ public void readVectored_successfulRead()
...
+
+  @Test
+  public void read_onClosedChannel_throwsClosedChannelException() throws IOException {
+    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel(DEFAULT_ITEM_INFO);
+
+    bidiReadChannel.close();
+
+    assertThrows(ClosedChannelException.class, () -> bidiReadChannel.read(ByteBuffer.allocate(1)));
+  }
+
```
animesh-g, line 142: I think we should add a test where channel is re-read because first call threw IOException.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: I think we can have a separate PR, so we can validate that we are not breaking anything by renaming this flag. Would also be easier to roll back just in case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: I think for now keeping this channel consistent with the behavior of GRPC channel would be ideal. It can be a follow up initiative to make  it consistent across channels including JSON.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Same as above.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
Dhriti07: I did't understand which specific case you are referring to

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: We are planning to add it in the future. Added TODO for the same.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

Dhriti07: Thanks for pointing out. Had added it for debugging purpose, it was redundant.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
animesh-g: I meant for the IOException case. Usually IOException is treated as transient error in Hadoop/Spark and they retry on the same. In the retry, they read again from the same channel. I wanted to be that scenario correctly tested in UTs.

Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,53 +1,63 @@
...
+import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.IntFunction;
+import javax.annotation.Nullable;
 
 public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableByteChannel {
```
ShreyasSinha: Make this class final

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -60,55 +70,116 @@ private static BlobReadSession initializeBlobReadSession(
...
+          futureBytes.get(readTimeout.toNanos(), TimeUnit.NANOSECONDS); ) {
+        ByteString byteString = dbs.byteString();
+        bytesRead = byteString.size();
+
+        if (bytesRead == 0 && position < objectSize) {
+          throw new IOException(
+              String.format("Read 0 bytes without blocking from object: '%s'", resourceId));
+        }
+
+        if (bytesRead > 0) {
```
ShreyasSinha: is this condition required?

byteString.size == 0 is already handled and .size can't be negative?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -168,13 +239,48 @@ private long processBytesAndCompleteRange(
       int size = byteString.size();
       bytesRead += size;
       ByteBuffer buf = allocate.apply(size);
-      // This loop efficiently copies data without creating intermediate byte arrays on the heap
```
ShreyasSinha: any specific reason for removing the zero copy comments?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -168,13 +239,48 @@ private long processBytesAndCompleteRange(
...
+    if (!isOpen()) {
+      GoogleCloudStorageEventBus.postOnException();
+      throw new ClosedChannelException();
+    }
+  }
+
+  protected void initMetadata(@Nullable String encoding, long sizeFromMetadata)
+      throws UnsupportedOperationException {
+    gzipEncoded = nullToEmpty(encoding).contains(GZIP_ENCODING);
+    // TODO(dhritichopra) Add Support for GZIP Encoding
```
ShreyasSinha, line 258: Do create a buganizer bug for this and add it to connector backlog in taskflow.

Dhriti07: Okay sure.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done thanks

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,53 +1,63 @@
...
 import java.util.function.IntFunction;
+import javax.annotation.Nullable;
 
-public class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableByteChannel {
-
+public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableByteChannel {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
-  private final Storage storage;
+
+  private static final int EOF_RETURN_VALUE = -1;
```
animesh-g: nit: can be just called EOF.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
animesh-g: Since we are keeping the behavior same as gRPC, it should be removed. https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/f33f12a0a6c4308e159c854a22c807a8c530d704/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java#L125

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
animesh-g: Same as above. gRPC channel doesn't throw exception in this case. 
https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/f33f12a0a6c4308e159c854a22c807a8c530d704/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java#L157

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

PR #1508, opened 2025-09-11
Footer Caching for Bidi Read Channel

Dhriti07:
Footer Caching for Bidi Read Channel

### DIFF
--- a/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
+++ b/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
@@ -54,6 +54,64 @@ public class GoogleCloudStorageBidiReadChannelTest {
           /* verificationAttributes= */ null);
 
   @Test
+  public void constructor_whenSessionInitializationFails_throwsIOException() {
+    Storage storage = mock(Storage.class);
+    when(storage.blobReadSession(any()))
+        .thenReturn(
+            ApiFutures.immediateFailedFuture(
+                new ExecutionException("Failed to create session", null)));
+
+    IOException e =
+        assertThrows(
+            IOException.class,
+            () ->
+                new GoogleCloudStorageBidiReadChannel(
+                    storage,
+                    DEFAULT_ITEM_INFO,
+                    GoogleCloudStorageReadOptions.builder().build(),
+                    Executors.newSingleThreadExecutor()));
+
+    // The original exception is wrapped in IOException by initializeBlobReadSession
+    assertThat(e).hasCauseThat().isInstanceOf(ExecutionException.class);
+    assertThat(e).hasCauseThat().hasMessageThat().contains("Failed to create session");
+  }
+
+  // TODO(dhritichopra): Remove test after support for Gzip is added
+  @Test
+  public void constructor_whenItemIsGzipEncoded_throwsUnsupportedOperationException()
+      throws IOException {
+    Storage storage = mock(Storage.class);
+    // Mock session creation to succeed (so the constructor gets to the metadata check)
+    when(storage.blobReadSession(any()))
+        .thenReturn(ApiFutures.immediateFuture(new FakeBlobReadSession()));
+
+    GoogleCloudStorageItemInfo gzipItemInfo =
+        GoogleCloudStorageItemInfo.createObject(
+            RESOURCE_ID,
+            /* creationTime= */ 10L,
+            /* modificationTime= */ 15L,
+            /* size= */ OBJECT_SIZE,
+            /* contentType= */ "text/plain",
+            /* contentEncoding= */ "gzip", // <-- Set gzip encoding
+            /* metadata= */ null,
+            /* contentGeneration= */ 1,
+            /* metaGeneration= */ 2L,
+            /* verificationAttributes= */ null);
+
+    UnsupportedOperationException e =
+        assertThrows(
+            UnsupportedOperationException.class,
+            () ->
+                new GoogleCloudStorageBidiReadChannel(
+                    storage,
+                    gzipItemInfo, // Use the gzip item info
+                    GoogleCloudStorageReadOptions.builder().build(),
+                    Executors.newSingleThreadExecutor()));
+
+    assertThat(e).hasMessageThat().isEqualTo("Gzip Encoded Files are not supported");
+  }
+
+  @Test
   public void readVectored_successfulRead()
       throws IOException, ExecutionException, InterruptedException, TimeoutException,
           URISyntaxException {
@@ -239,6 +297,14 @@ public class GoogleCloudStorageBidiReadChannelTest {
   }
 
   @Test
+  public void position_onClosedChannel_throwsClosedChannelException() throws IOException {
+    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel();
+    bidiReadChannel.close();
+
+    assertThrows(ClosedChannelException.class, () -> bidiReadChannel.position(10));
+  }
+
+  @Test
   public void size_successful() throws IOException {
     GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel();
 
@@ -325,6 +391,39 @@ public class GoogleCloudStorageBidiReadChannelTest {
 
     verify(fakeSession, times(2)).readAs(any());
     verifyNoMoreInteractions(fakeSession);
+  }
+
+  @Test
+  public void read_whenReadExactRequestedBytesEnabled_skipsFooterCache() throws Exception {
+    int footerSize = 20;
+    GoogleCloudStorageReadOptions readOptions =
+        GoogleCloudStorageReadOptions.builder()
+            .setMinRangeRequestSize(footerSize)
+            .setReadExactRequestedBytesEnabled(true)
+            .build();
+
+    FakeBlobReadSession fakeSession = spy(new FakeBlobReadSession());
+    Storage storage = mock(Storage.class);
+    when(storage.blobReadSession(any())).thenReturn(ApiFutures.immediateFuture(fakeSession));
+
+    GoogleCloudStorageBidiReadChannel channel =
+        new GoogleCloudStorageBidiReadChannel(
+            storage, DEFAULT_ITEM_INFO, readOptions, Executors.newSingleThreadExecutor());
+
+    // Verify footer is not initially cached
+    assertNull("Footer should be null initially", getPrivateField(channel, "footerContent"));
+
+    // Position the read within the footer range
+    long readPosition = OBJECT_SIZE - 10; // Inside 20-byte footer range
+    channel.position(readPosition);
+
+    ByteBuffer buffer = ByteBuffer.allocate(5);
+    int bytesRead = channel.read(buffer);
+
+    assertEquals(5, bytesRead);
+
+    // Verify footer is STILL null, because ReadExact disabled the optimization
+    assertNull("Footer should remain null", getPrivateField(channel, "footerContent"));
   }
 
   @Test


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -283,4 +263,164 @@ private void validatePosition(long position) throws IOException {
...
+      int bytesRead = readFromCache(dst);
+      return Optional.of(bytesRead);
+    }
+
+    // The footer is not cached, but we should fetch it as position is in range.
+    boolean shouldPrefetchFooter =
+        !gzipEncoded
+            && footerContent == null
+            && isFooterRead()
+            && !readOptions.isReadExactRequestedBytesEnabled();
```
ShreyasSinha, line 353: Is this option disabled by default?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -283,4 +263,164 @@ private void validatePosition(long position) throws IOException {
...
+    logger.atFiner().log(
+        "Prefetching footer for '%s'. Position: %d, Size: %d",
+        resourceId, footerStartPosition, footerSize);
+
+    ApiFuture<DisposableByteString> futureBytes =
+        blobReadSession.readAs(
+            ReadProjectionConfigs.asFutureByteString()
+                .withRangeSpec(RangeSpec.of(footerStartPosition, footerSize)));
+
+    try (DisposableByteString dbs = futureBytes.get(readTimeout.toNanos(), TimeUnit.NANOSECONDS)) {
```
ShreyasSinha: Refactor this common read implementation here and in performStandardRead to a new method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Yes, the default is false

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
...
     }
   }
+
+  // Helper functions related to Footer Caching.
+
+  /**
+   * Check if the read is current position of the channel belongs to the last Range Request. If yes
+   * that will be cached as the footer and future requests in the footer region will be served from
+   * the cache.
+   */
```
animesh-g, line 273: As mentioned above, this should be private so it doesn't need java doc as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
...
+  @VisibleForTesting
+  public void cacheFooter()
+      throws InterruptedException, ExecutionException, TimeoutException, IOException {
+    long footerStartPosition = max(0, objectSize - readOptions.getMinRangeRequestSize());
+    long footerSize = objectSize - footerStartPosition;
+
+    if (footerSize <= 0) {
+      return;
+    }
```
animesh-g, line 291: Shouldn't this be IllegalState? Use checkState() to validate this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
...
+    logger.atFiner().log(
+        "Prefetching footer for '%s'. Position: %d, Size: %d",
+        resourceId, footerStartPosition, footerSize);
+    ByteString byteString = readBytes(footerStartPosition, footerSize);
+    if (byteString.size() != footerSize) {
+      throw new IOException(
+          String.format(
+              "Failed to read complete footer for '%s'. Expected %d, got %d",
+              resourceId, footerSize, byteString.size()));
+    }
```
animesh-g, line 302: Use checkState() for these kind of checks.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -168,13 +217,206 @@ private long processBytesAndCompleteRange(
...
+  /** Serve the data from the cached footer if the footer Cache is already populated */
+  @VisibleForTesting
+  public int readFromCache(ByteBuffer dst) {
+    // Calculate offset inside the footerContent byte array
+
+    long footerStartPosition = objectSize - footerContent.length;
+
+    // The Offset in the cache array will be the number bytes to start reading from the beginning of
+    // the footer.
+    int offsetInFooterCache = (int) (position - footerStartPosition);
```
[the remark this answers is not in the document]
animesh-g: Please handle this. Use `toIntExact` for the same.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
...
+   * Attempts to fulfill the read request from the footer cache.
+   *
+   * <p>If the footer is already cached and the read is within its range, it reads from the cache.
+   * If the footer is not cached but the read is in the footer region, it fetches the footer, caches
+   * it, and then reads from the new cache.
+   *
+   * @return An {@link Optional} containing the number of bytes read if the request was handled by
+   *     the cache, or {@link Optional#empty()} if the request is not a footer request.
+   * @throws IOException if fetching the footer fails.
+   */
```
animesh-g, line 340: No need of javadocs in private methods.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
...
+  private Optional<Integer> tryReadFromFooter(ByteBuffer dst) throws IOException {
+    // The footer is already cached and our read position is within it.
+    if (footerContent != null && position >= objectSize - footerContent.length) {
+      int bytesRead = readFromCache(dst);
+      return Optional.of(bytesRead);
+    }
```
animesh-g, line 346: This call can be moved to try to simplify the method and duplication of logic.. It will look like

```

  boolean isCachedFooterRead = footerContent != null && position >= objectSize - footerContent.length;

  boolean shouldPrefetchFooter =
      !gzipEncoded
          && footerContent == null
          && isFooterRead()
          && !readOptions.isReadExactRequestedBytesEnabled();

  if (isCachedFooterRead || shouldPrefetchFooter) {
    try {
      if (shouldPrefetchFooter) {
        cacheFooter();
      }
      return Optional.of(readFromCache(dst));
    } catch (InterruptedException | ExecutionException | TimeoutException e) {
      throw new IOException(String.format("Footer prefetch failed on %s", resourceId), e);
    }
  }

  return Optional.empty();
}
```

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -280,4 +263,155 @@ private void validatePosition(long position) throws IOException {
               position, objectSize, resourceId));
     }
   }
+
+  // Helper functions related to Footer Caching.
```
animesh-g, line 267: Most of these newly added helper methods in this file are public but they should be private. I don't see any reason of them being used outside this file.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -87,39 +92,18 @@ public int read(ByteBuffer dst) throws IOException {
...
-      throw new IOException(
-          String.format("Read failed on %s at position %d", resourceId, position), e);
+    // Attempt to service the read from the footer cache logic.
+    // This method will handle both reading from an existing cache and populating it if necessary.
+    // It returns an Optional<Integer>:
+    //  - If present, the read was handled by the cache, and the value is the bytes read.
+    //  - If empty, the read is not in the footer range, and a standard read should be performed.
+    Optional<Integer> bytesReadFromCache = tryReadFromFooter(dst);
+    if (bytesReadFromCache.isPresent()) {
+      return bytesReadFromCache.get();
```
animesh-g, line 102: As per this logic, bytes are read either from cache or server.
Isn't it possible that the requested range has overlapping byte with the cached footer? Why haven't we handled this case where some bytes are read from server and some from cache.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
```diff
@@ -71,12 +128,12 @@ public void readVectored_successfulRead()
   }
 
   @Test
-  public void read_successful() throws IOException {
-    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel(DEFAULT_ITEM_INFO);
+  public void performStandardRead_successful() throws IOException {
```
animesh-g, line 131: As mentioned in the above file, new helper methods needs to be private thus don't need explicit tests for the same. It should just just test the regular reads.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: I have marked them as Visible for Testing. It is was done to test the functionality for good coverage and testing.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
Dhriti07: The regular reads are already tested. These functions are visibleForTesting. It would be good to have testing for the functionality of the helper functions as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
animesh-g: Methods like isFooterRead, cacheFooter, readFromCache, performStandardRead are `public`. Methods shouldn't be made public to have better testing coverage.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Added rationale for public VisisbleForTesting functions in another comment thread. The java doc would also be nice for future readers to understand what this function does.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: Same as above

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: The primary source of latency for any read operation is the network call. Even if we served a few partial bytes from the cache, we would still need to make that expensive network call to get the remaining bytes from the server. Since we can't avoid the network call, adding the complexity to handle this partial read doesn't save us any real time. A "look-ahead buffer" feature (in an upcoming PR) will properly address this by caching data outside of the footer area. This will effectively serve sequential small reads from the cache and truly help reduce server calls.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
animesh-g: When we test the public read() method, isn't performStandardRead already tested as part of that. Why should we test the same thing again? Even read from readFromCache(), doesn't need to called from test and still can be tested. 

Calling private method is required very rarely for unit testing. See tests written in GoogleCloudStorageClientReadChannelTest. They don't need to call private method to test the caching behavior.

PR #1510, opened 2025-09-11
Look ahead buffer and Lazy client Init 

Dhriti07:


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -376,6 +397,64 @@ && isFooterRead()
...
+  public void refillInternalBuffer() throws IOException {
+    invalidateBuffer();
+    if (position >= objectSize) return;
+
+    long bytesToRequest = Math.min(readOptions.getMinRangeRequestSize(), objectSize - position);
+
+    if (bytesToRequest <= 0) return;
+
+    try {
+      ApiFuture<DisposableByteString> futureBytes =
```
ShreyasSinha: Use the method refactored out in the previous PR for this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,125 +1,196 @@
...
-          System.currentTimeMillis() - clientInitializationDurationStartTime);
-    } catch (InterruptedException | ExecutionException | TimeoutException e) {
-      throw new IOException(e);
+  private synchronized BlobReadSession getBlobReadSession() {
+    if (blobReadSession == null) {
+      BlobReadSession readSession = null;
+      try {
+        readSession = sessionFuture.get(readOptions.getBidiClientTimeout(), TimeUnit.SECONDS);
+      } catch (InterruptedException | ExecutionException | TimeoutException e) {
+        throw new RuntimeException("Failed to get BlobReadSession", e);
```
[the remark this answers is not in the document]
ShreyasSinha: Do take care of this as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -1,125 +1,196 @@
...
   public void close() throws IOException {
-    blobReadSession.close();
+    if (isOpen) {
+      isOpen = false;
+      logger.atFinest().log("Closing channel for '%s'", resourceId);
+      if (blobReadSession != null) {
+        blobReadSession.close();
+      }
+    }
   }
```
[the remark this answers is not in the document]
ShreyasSinha: This is important as well.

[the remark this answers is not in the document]
Dhriti07: My understanding is that GoogleHadoopFSInputStream acts as a thread-safe wrapper. Since its close() method is synchronized, it correctly serializes access and prevents any race conditions when calling the underlying BidiReadChannel.close()

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -368,6 +387,54 @@ && isFooterRead()
...
 
+  // Helper functions for the Look ahead read buffer.
+  public boolean isBufferValid() {
+    return internalBuffer != null
+        && position >= bufferStartPosition
+        && position < bufferStartPosition + internalBuffer.limit();
+  }
+
+  @VisibleForTesting
+  public void invalidateBuffer() {
```
ShreyasSinha: This looks very simple and can probably be tested via the public methods.

Any specific reason its made public for testing?

Dhriti07: We can test this function via the public method (position in this case). Even though possible in my opinion it would result in the tests being subject to not being extendible and dependent on other functions. 

An indirect test would be tightly coupled to the implementation details of both position() and isBufferValid(). This would make the test brittle, as any future refactoring of the conditions within isBufferValid() could cause the invalidateBuffer() test to fail, even if the method itself remains correct.

The test setup would require a setup so that isBufferValid to be false resulting in the function call of invalidateBuffer. Any changes in the position function deciding to trigger this function in a different condition etc would break the test of this function without it being the root cause. Hence my preference is to test the functions via VisibleForTesting.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -384,6 +405,53 @@ && isFooterRead()
     return Optional.empty();
   }
 
+  // Helper functions for the Look ahead read buffer.
+  public boolean isBufferValid() {
```
animesh-g, line 409: It should be private.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -398,36 +466,42 @@ public int performStandardRead(ByteBuffer dst) throws IOException {
...
 
-    ByteString byteString;
-    try {
-      byteString = readBytes(position, bytesToRequest);
+    int bytesRead;
+
+    try (DisposableByteString dbs = readBytes(position, bytesToRequest)) {
+      ByteString byteString = dbs.byteString();
+      bytesRead = byteString.size();
+      bytesRead = byteString.size();
```
animesh-g: `bytesRead` is populated twice unnecessarily. Please fix.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
Dhriti07: As discussed offline, I will take up an initiative to refactor this and other public methods for tests to private. Will raise a followup PR for the same.

PR #1625, opened 2026-01-30
Adding a local version of google.p12 file

suryasoma:
Adding a local version of google.p12 keystore file, which is extracted from google-java-api-client and added Certigna CA certificate to it.

The default `GoogleUtils.getCertificateTrustStore()`, does not have Certigna ca cert which is needed for TPC in it.

`google-java-api-client` team has said, they will no longer support this keystore, rather mentioned for us to have this local copy in the interim and then move to read from the Java Keystore.


util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -166,8 +170,18 @@ && getRequestingPort() == proxyUri.getPort()) {
   static NetHttpTransport.Builder createNetHttpTransportBuilder(
       @Nullable URI proxyUri, @Nullable Duration readTimeout)
       throws IOException, GeneralSecurityException {
-    NetHttpTransport.Builder builder =
-        new NetHttpTransport.Builder().trustCertificates(GoogleUtils.getCertificateTrustStore());
+
+    KeyStore keyStore = SecurityUtils.getPkcs12KeyStore();
```
singhravidutt, line 174: Blast radius of it could be huge, does it make sense to put this change behind a flag and slowly deprecate it?

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
```diff
@@ -45,6 +47,8 @@
 /** Factory for creating HttpTransport types. */
 public class HttpTransportFactory {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+  private static final String KEYSTORE_FILE_NAME = "google.p12";
```
singhravidutt, line 50: Given it is going to be used just one it's fine to jsut hardcode it. 

Also, we need to package the `google.p12` relative to `HttpTransportFactory`. is it verified?

singhravidutt (commented):
Mind adding a test case similar to [GoogleUtilsTest].(https://github.com/googleapis/google-api-java-client/blob/main/google-api-client/src/test/java/com/google/api/client/googleapis/GoogleUtilsTest.java#L29)

animesh-g:
@suryasoma Can this PR be raised against the master branch? We use master for development and port commits to release branches. Also, the latest release branch is 4.0.x, not 3.1.x.

singhravidutt:
> @suryasoma Can this PR be raised against the master branch? We use master for development and port commits to release branches. Also, the latest release branch is 4.0.x, not 3.1.x.

Not sure if we want to keep it on long term. Idea is to unlock customer as of now. Maintaining the certificate file will be a huge tech debt.

util/src/main/java/com/google/cloud/hadoop/util/HttpTransportFactory.java
animesh-g: +1

PR #1659, opened 2026-03-17
Add Additional Metrics to track thread_wait, TTFB and data transfer

shameersss1:
###   Summary of Changes

This PR introduces new telemetry and performance metrics to track latency across Google Cloud Storage (GCS) reads, specifically focusing on thread queue times, Time To First Byte (TTFB), and data transfer durations.

  High-Level Summary of Changes:

   1. Introduced Event-Driven Latency Tracking: 
     We created a new, reusable stream decorator (GcsReadDurationTracker). Instead of hardcoding timing logic inside the GCS read channels, this decorator cleanly wraps the network stream. It measures exactly
  how long it takes to transfer data and automatically publishes those metrics (like data_transfer_duration and latency_breached_count) to the global Hadoop Event Bus.

   2. Added Vectored IO Wait Time Metrics:
     We updated the asynchronous VectoredIOImpl engine. When Hadoop requests multiple file ranges in parallel, we now explicitly track how long those background tasks sit waiting in the thread pool queue
  (STREAM_READ_VECTORED_THREAD_WAIT_DURATION) before execution begins.

   3. Improved High-Latency Logging:
     When a read operation takes longer than the expected threshold (1000ms), the system now logs a warning that surgically includes the specific GCS x-guploader-uploadid. This makes it significantly easier for user to correlate slow Hadoop reads with backend Google Cloud server logs.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
@@ -25,6 +25,7 @@ import com.google.cloud.hadoop.util.GCSChecksumFailureEvent;
 import com.google.cloud.hadoop.util.GcsJsonApiEvent;
 import com.google.cloud.hadoop.util.GcsJsonApiEvent.EventType;
 import com.google.cloud.hadoop.util.GcsJsonApiEvent.RequestType;
+import com.google.cloud.hadoop.util.GcsReadMetricEvent;
 import com.google.cloud.hadoop.util.GcsRequestExecutionEvent;
 import com.google.cloud.hadoop.util.IGcsJsonApiEvent;
 import com.google.common.annotations.VisibleForTesting;
@@ -62,6 +63,7 @@ public class GoogleCloudStorageEventSubscriber {
       logger.atFiner().log("Subscriber class invoked for first time");
       INSTANCE = new GoogleCloudStorageEventSubscriber(storageStatistics);
     }
+    GoogleCloudStorageEventSubscriber.storageStatistics = storageStatistics;
     return INSTANCE;
   }
 
@@ -96,6 +98,10 @@ public class GoogleCloudStorageEventSubscriber {
       storageStatistics.incrementCounter(GoogleCloudStorageStatistics.GCS_API_TIME, duration);
 
       RequestType requestType = (RequestType) event.getProperty(GcsJsonApiEvent.REQUEST_TYPE);
+      if (requestType == RequestType.GET_MEDIA) {
+        storageStatistics.updateStats(
+            GhfsStatistic.STREAM_READ_CONNECTION_DURATION, duration, eventContext);
+      }
       if (requestToGcsStatMap.containsKey(requestType)) {
         updateMetric(requestToGcsStatMap.get(requestType), duration, eventContext);
       } else if (requestToGhfsStatMap.containsKey(requestType)) {
@@ -112,6 +118,26 @@ public class GoogleCloudStorageEventSubscriber {
           GoogleCloudStorageStatistics.GCS_BACKOFF_TIME, backOffTime);
     } else if (eventType == EventType.EXCEPTION) {
       storageStatistics.incrementGcsExceptionCount();
+    }
+  }
+
+  @Subscribe
+  private void subscriberOnGcsReadMetricEvent(@Nonnull GcsReadMetricEvent event) {
+    if (event.getConnectionDurationMs() > 0) {
+      storageStatistics.updateStats(
+          GhfsStatistic.STREAM_READ_CONNECTION_DURATION,
+          event.getConnectionDurationMs(),
+          event.getStreamPath());
+    }
+    if (event.getDataTransferDurationMs() > 0) {
+      storageStatistics.updateStats(
+          GhfsStatistic.STREAM_READ_DATA_TRANSFER_DURATION,
+          event.getDataTransferDurationMs(),
+          event.getStreamPath());
+      if (event.isLatencyThresholdBreached()) {
+        storageStatistics.incrementCounter(
+            GoogleCloudStorageStatistics.GCS_READ_DATA_TRANSFER_LATENCY_BREACHED_COUNT, 1);
+      }
     }
   }
 
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
@@ -123,21 +123,23 @@ public class GoogleCloudStorageEventSubscriber {
 
   @Subscribe
   private void subscriberOnGcsReadMetricEvent(@Nonnull GcsReadMetricEvent event) {
-    if (event.getConnectionDurationMs() > 0) {
-      storageStatistics.updateStats(
-          GhfsStatistic.STREAM_READ_CONNECTION_DURATION,
-          event.getConnectionDurationMs(),
-          event.getStreamPath());
-    }
-    if (event.getDataTransferDurationMs() > 0) {
-      storageStatistics.updateStats(
-          GhfsStatistic.STREAM_READ_DATA_TRANSFER_DURATION,
-          event.getDataTransferDurationMs(),
-          event.getStreamPath());
-      if (event.isLatencyThresholdBreached()) {
-        storageStatistics.incrementCounter(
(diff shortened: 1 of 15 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -471,6 +479,22 @@ public boolean isOpen() {
     return channelIsOpen;
   }
 
+  /**
+   * Returns the duration of the GCS media request in milliseconds. Returns 0 if no request has been
+   * made yet.
+   */
+  public long getMediaRequestDurationMs() {
```
singhravidutt: We do have another ReadChannel GCSClientReadChannel which will be used with gRPC, does it make sense to expand this to that channel as well?

shameersss1: Ack. I will raise a new revision to fix it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannelMetrics.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.gcsio;
+
+/** Interface for providing GCS read channel metrics. */
+public interface GoogleCloudStorageReadChannelMetrics {
```
singhravidutt: can it be private?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -943,7 +968,10 @@ protected InputStream openStream(long bytesToRead) throws IOException {
     Storage.Objects.Get getObject = createDataRequest(rangeHeader);
     HttpResponse response;
     try {
+      long startTime = System.currentTimeMillis();
       response = getObject.executeMedia();
+      long endTime = System.currentTimeMillis();
+      mediaRequestDurationMs = endTime - startTime;
```
singhravidutt: So, if multiple channels are created per GCSReadChannel, `mediaRequestDurationMs` will capture the time of last channel, same for `dataTransferDurationMs`?
Is it a correct expectation?

In cases of FSInputStream.read() we might endup creating multiple channels with gcs backend inside a single stream.

shameersss1: Yes, for non-vectored reads, channels will be re-used and the metrics will be only for the final reads. 
Since in this PR we are specifically targeting VectorIO, we should be good and i will add a todo:// here and fix this in a separate PR.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
```diff
@@ -115,6 +121,26 @@ private void subscriberOnGcsRequestExecutionEvent(IGcsJsonApiEvent event) {
     }
   }
 
+  @Subscribe
+  private void subscriberOnGcsReadMetricEvent(@Nonnull GcsReadMetricEvent event) {
+    if (event.getConnectionDurationMs() > 0) {
```
singhravidutt: are these if checks required, these values will always be >0

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2024 Google LLC
```
singhravidutt: nit: 2026?

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  @Override
+  public void close() throws IOException {
+    try {
+      super.close();
+    } finally {
+      boolean latencyThresholdBreached = cumulativeTransferDurationMs > latencyLoggingThresholdMs;
+      if (latencyThresholdBreached) {
```
singhravidutt, line 75: better to extract idempotencyToken too.

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.api.client.http.HttpHeaders;
+import com.google.common.flogger.GoogleLogger;
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.concurrent.TimeUnit;
+
+/** A decorator for {@link InputStream} that tracks the data transfer duration. */
+public class GcsReadDurationTrackerStream extends FilterInputStream {
```
singhravidutt, line 28: Why FilterInputStream we don't filter anything, just a pass though with tracing, correct?

singhravidutt:
Some minor comments, LGTM.

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: Yes. While we aren't performing any data filtering, FilterInputStream is used here as a standard implementation of the Decorator Pattern. By extending it instead of InputStream, we leverage its built-in
  delegation logic. This allows us to instrument only the specific methods we need (read and close) for metrics collection while avoiding the boilerplate of manually delegating every other method to the
  underlying stream

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: Last time i interacted with GCS team , uploadId  was the once which they were looking for to track the request

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: ack

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleCloudStorageEventSubscriber.java
shameersss1: These checks are required because. The subscriber method is called at two distinct moments during a GCS read operation

```
 The Timeline Example:
   1. 0ms: Request starts.
   2. 200ms: GCS responds. -> Event 1 is fired (connection=200, transfer=0).
   3. 200ms - 1200ms: User reads the file.
   4. 1200ms: User calls stream.close(). -> Event 2 is fired (connection=0, transfer=1000).

  If you didn't have the if checks:
   * At 200ms, you would record a fake 0ms transfer.
   * At 1200ms, you would record a fake 0ms connection.
   ```

  1. The Connection Event
  In GoogleCloudStorageReadChannel.java (inside the openStream logic):
  As soon as the HTTP request completes (executeMedia()), it calculates how long that took and sends the event.

   1 // Line 974 in GoogleCloudStorageReadChannel.java
   2 GoogleCloudStorageEventBus.postReadMetricEvent(
   3     GcsReadMetricEvent.ofConnection(
   4         System.currentTimeMillis() - startTime, URI.create(resourceId.toString())));

  2. The Data Transfer Event
  In GcsReadDurationTrackerStream.java (inside the close() method):
  After the user is done reading the file and closes the stream, it sends the total time spent downloading data.

   1 // Line 84 in GcsReadDurationTrackerStream.java
   2 GoogleCloudStorageEventBus.postReadMetricEvent(
   3     GcsReadMetricEvent.ofDataTransfer(
   4         cumulativeTransferDurationMs, streamPath, latencyThresholdBreached));

singhravidutt: Wouldn't it be better to have two separate events for them or have enum type to segregate them. Separating them based it's value being 0 could lead to issues.

shameersss1: Yes, as discussed we can have a Event Type and capture tehse

singhravidutt (approved):

PR #1670, opened 2026-03-23
Fix zombie threads and background resource leaks in vectored reads

shameersss1:
This change ensures background worker threads terminate immediately when a vectored read is cancelled by the caller (e.g., Spark/Parquet). It implements cancellation propagation in VectoredIOImpl by interrupting worker futures upon range data cancellation. GoogleCloudStorageReadChannel is updated to catch ClosedByInterruptException and InterruptedIOException, triggering immediate cleanup and preventing redundant retries. These improvements prevent background resource saturation and ensure prompt release of network and thread pool resources. Wildcard imports in VectoredIOImpl were also replaced with explicit
ones.

The following is the follow after this change :

   1. The Trigger: Parquet Timeout
       * Spark/Parquet waits for data using a timed wait (e.g., 30 seconds). If the GCS network call is slow, Parquet's CompletableFuture.get(30, SECONDS) expires and throws a TimeoutException.
       * Immediately after the timeout, Parquet calls range.getData().cancel(true).
       * This signals the system: "I no longer want this data; stop all associated work."

  ---

   2. The Link: VectoredIOImpl Cancellation Propagation
           1. Trigger: The whenComplete block on the range's data future is triggered by the cancellation.
           2. Detection: It detects that the exception is a CancellationException.
           3. Targeting: It identifies the specific executionFuture (the background worker thread) associated with that range.
           4. Action: It calls executionFuture.cancel(true), sending the INTERRUPT signal to the worker thread.

  ---

   3. The Execution: Worker Thread Interruption
       * The background worker thread is typically stuck inside the do...while loop of GoogleCloudStorageReadChannel.read(), waiting for data from the GCS network socket.
           1. Signal: Because executionFuture.cancel(true) was called, the JVM sets the interrupt flag on that worker thread.
           2. Reaction: The low-level Java NIO channel (contentChannel.read()) is interruptible. As soon as it detects the interrupt flag, it immediately stops the network transfer and throws a
              java.nio.channels.ClosedByInterruptException.

  ---

   4. The Recovery: Catching the Interrupt
       * The thread jumps out of the read() call and into the new consolidated catch block in GoogleCloudStorageReadChannel.java:
           1. Detection: It catches the IOException and identifies it as an interruption (while carefully ignoring standard SocketTimeoutException).
           2. Cleanup: It calls closeContentChannel() to release the GCS network connection immediately.
           3. State Preservation: It calls Thread.currentThread().interrupt() to ensure the thread's interrupt status remains set for the parent pool.
           4. Stop Retries: Instead of entering the standard IOException retry logic (which would wait and try again up to 10 times), it throws a new IOException wrapping the interrupt, exiting the loop
              immediately.

  ---

   5. Final Result: Efficient Resource Release
       * The exception bubbles up to the worker thread's top level in VectoredIOImpl, where it is logged.
           * Thread Recovery: The thread is returned to the boundedThreadPool immediately, making it available for new tasks.
           * Bandwidth Savings: The GCS connection is closed instantly, stopping unnecessary data downloads and saving costs.
           * System Stability: The user sees the timeout error as expected, but background "ghost" tasks are eliminated, preventing thread pool saturation and system-wide hangs.

Error StackTrace

Screenshot 2026-04-01 at 4 47 52 PM

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
@@ -37,9 +37,11 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -143,19 +145,31 @@ public class VectoredIOImpl implements Closeable {
         // case when ranges are not merged
         for (FileRange sortedRange : sortedRanges) {
           long startTimer = System.currentTimeMillis();
-          boundedThreadPool.submit(
-              () -> {
-                logger.atFiner().log("Submitting range %s for execution.", sortedRange);
-                // Reset thread stats as a new task is being submitted to this thread.
-                // all required stats must be collected before task is finished.
-                resetThreadLocalStats();
-                readSingleRange(sortedRange, allocate, channelProvider);
-                long endTimer = System.currentTimeMillis();
-                storageStatistics.updateStats(
-                    GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
-                    endTimer - startTimer,
-                    channelProvider.gcsPath);
-              });
+          Future<?> executionFuture =
+              boundedThreadPool.submit(
+                  () -> {
+                    logger.atFiner().log("Submitting range %s for execution.", sortedRange);
+                    // Reset thread stats as a new task is being submitted to this thread.
+                    // all required stats must be collected before task is finished.
+                    resetThreadLocalStats();
+                    readSingleRange(sortedRange, allocate, channelProvider);
+                    long endTimer = System.currentTimeMillis();
+                    storageStatistics.updateStats(
+                        GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
+                        endTimer - startTimer,
+                        channelProvider.gcsPath);
+                  });
+          sortedRange
+              .getData()
+              .whenComplete(
+                  (res, ex) -> {
+                    if (ex instanceof CancellationException) {
+                      logger.atFiner().log(
+                          "Cancelling execution future for range %s due to CancellationException",
+                          sortedRange);
+                      executionFuture.cancel(true);
+                    }
+                  });
         }
       } else {
         List<CombinedFileRange> combinedFileRanges = getCombinedFileRange(sortedRanges);
@@ -165,21 +179,33 @@ public class VectoredIOImpl implements Closeable {
           CompletableFuture<ByteBuffer> result = new CompletableFuture<>();
           combinedFileRange.setData(result);
           long startTimer = System.currentTimeMillis();
-          boundedThreadPool.submit(
-              () -> {
-                logger.atFiner().log(
-                    "Submitting combinedRange %s for execution.", combinedFileRange);
+          Future<?> executionFuture =
+              boundedThreadPool.submit(
+                  () -> {
+                    logger.atFiner().log(
+                        "Submitting combinedRange %s for execution.", combinedFileRange);
 
-                // Reset thread stats as a new task is being submitted to this thread.
-                // all required stats must be collected before task is finished.
-                resetThreadLocalStats();
-                readCombinedRange(combinedFileRange, allocate, channelProvider);
-                long endTimer = System.currentTimeMillis();
-                storageStatistics.updateStats(
-                    GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
-                    endTimer - startTimer,
-                    channelProvider.gcsPath);
-              });
+                    // Reset thread stats as a new task is being submitted to this thread.
+                    // all required stats must be collected before task is finished.
+                    resetThreadLocalStats();
+                    readCombinedRange(combinedFileRange, allocate, channelProvider);
+                    long endTimer = System.currentTimeMillis();
+                    storageStatistics.updateStats(
+                        GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
+                        endTimer - startTimer,
+                        channelProvider.gcsPath);
+                  });
+          combinedFileRange
+              .getData()
+              .whenComplete(
+                  (res, ex) -> {
+                    if (ex instanceof CancellationException) {
+                      logger.atFiner().log(
+                          "Cancelling execution future for combined range %s due to CancellationException",
+                          combinedFileRange);
+                      executionFuture.cancel(true);
+                    }
+                  });
         }
       }
     }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
@@ -24,6 +24,7 @@ import static org.apache.hadoop.fs.VectoredReadUtils.validateRangeRequest;
 import com.google.cloud.hadoop.gcsio.FileInfo;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystem;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions;
+import com.google.cloud.hadoop.util.IoExceptionHelper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.flogger.GoogleLogger;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -199,6 +200,11 @@ public class VectoredIOImpl implements Closeable {
      * @param executionFuture the future representing the background thread execution.
      */
     private void addCancellationListener(FileRange range, Future<?> executionFuture) {
+      if (range instanceof CombinedFileRange) {
+        for (FileRange child : ((CombinedFileRange) range).getUnderlying()) {
+          addCancellationListener(child, executionFuture);
+        }
+      }
       range
           .getData()
           .whenComplete(
@@ -321,6 +327,9 @@ public class VectoredIOImpl implements Closeable {
         }
         combinedFileRange.getData().complete(readContent);
       } catch (Exception e) {
+        if (IoExceptionHelper.isInterrupted(e)) {
+          Thread.currentThread().interrupt();
+        }
         logger.atWarning().withCause(e).log(
             "Exception while reading combinedFileRange:%s for path: %s",
             combinedFileRange, channelProvider.gcsPath);
@@ -374,6 +383,9 @@ public class VectoredIOImpl implements Closeable {
             "Read single range completed from range: %s, path: %s", range, channelProvider.gcsPath);
         range.getData().complete(dst);
       } catch (Exception e) {
+        if (IoExceptionHelper.isInterrupted(e)) {
+          Thread.currentThread().interrupt();
+        }
         logger.atWarning().withCause(e).log(
             "Exception while reading range:%s for path: %s", range, channelProvider.gcsPath);
         captureAndResetThreadLocalStats();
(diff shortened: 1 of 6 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -143,19 +145,31 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+                        GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
+                        endTimer - startTimer,
+                        channelProvider.gcsPath);
+                  });
+          sortedRange
+              .getData()
+              .whenComplete(
+                  (res, ex) -> {
+                    if (ex instanceof CancellationException) {
+                      logger.atFiner().log(
```
mudit-97: this should be an ERROR or WARN log right?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -143,19 +145,31 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+                          "Cancelling execution future for range %s due to CancellationException",
```
mudit-97: Can we print something more here like task/containerId or appId?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -165,21 +179,33 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+                    readCombinedRange(combinedFileRange, allocate, channelProvider);
+                    long endTimer = System.currentTimeMillis();
+                    storageStatistics.updateStats(
+                        GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
+                        endTimer - startTimer,
+                        channelProvider.gcsPath);
+                  });
+          combinedFileRange
+              .getData()
+              .whenComplete(
```
mudit-97: can we initialize this complete as a common lambda and pass same at both places

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -165,21 +179,33 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+                  (res, ex) -> {
+                    if (ex instanceof CancellationException) {
+                      logger.atFiner().log(
+                          "Cancelling execution future for combined range %s due to CancellationException",
+                          combinedFileRange);
+                      executionFuture.cancel(true);
```
mudit-97: can executionFuture.cancel fail in some cases?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -363,6 +366,25 @@ public int read(ByteBuffer buffer) throws IOException {
         // each time we make progress we reset the retry counter.
         retriesAttempted = 0;
       } catch (IOException ioe) {
+        // Handle thread interruptions (e.g. from cancellation) separately from general
+        // IOExceptions.
+        // We catch InterruptedIOException but exclude its subclass SocketTimeoutException,
+        // which represents a network-level timeout that should still be retried.
+        if (ioe instanceof ClosedByInterruptException
```
mudit-97: can we keep this block post  closeContentChannel(); below and remove closeContentChannel(); from this>

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
mudit-97: All these things, should still happen even if we see interrupted exceptions

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -363,6 +366,25 @@ public int read(ByteBuffer buffer) throws IOException {
...
+            || (ioe instanceof InterruptedIOException
+                && !(ioe instanceof SocketTimeoutException))) {
+          logger.atInfo().withCause(ioe).log(
+              "Thread interrupted while reading '%s' at position %d. Stopping further processing.",
+              resourceId, currentPosition);
+          closeContentChannel();
+          // Re-assert interrupt status and percolate as a non-retryable exception.
+          Thread.currentThread().interrupt();
```
mudit-97, line 392: Does this ensure that GCS FrontEnd Request is cancelled?

mudit-97 (changes requested):
Minor comments

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
shameersss1: This log line gets printed as part of executor, so we can infer the container/executionId from there

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
shameersss1: WARN makes more sense here since this does not cause any fatal issue.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
shameersss1: Yes, it makes more sense!

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
shameersss1: The code which streams data from GCS `contentChannel.read()`  uses Java NIO channels, which are explicitly built to respond to exactly this signal and hence it is safe to use it for this use case

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
shameersss1: When thread is interrupted or ClosedByInterruptException which essentially mean that the task had failed and new task attempt is already is progress so we shouldn't retry the request in those case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
shameersss1: Yes

When you call closeContentChannel() inside that catch block (and also in the finally/auto-close blocks), the following sequence ensures the GCS FrontEnd request is stopped:

   1. Closing the Stream: closeContentChannel() closes the underlying InputStream (the objectContentStream returned by the GCS client).
   2. TCP Connection Termination: By closing the input stream, you are telling the Java HTTP client to stop reading from the socket.
   3. Socket Reset (RST/FIN): The HTTP client will either gracefully finish the current packet and close the connection or send a TCP reset to the GCS Frontend.
   4. Frontend Notification: When the GCS Frontend (the Google server side) detects that the client has closed the socket while it was still sending data, it immediately terminates its own internal processes for
      that request. It stops billing you for the egress of the remaining data and frees up its own resources.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
shameersss1: Yes, it will me more cleaner

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
mudit-97: But there are bunch of other things like bytesRead etc, will we accumulate those metrics properly even if the task got interrupted?

mudit-97 (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
shameersss1: As per the current design, only on successful read, the bytesRead is propagated back and accounted for. This is the same behavior when we max out the retry attempts as well

updateBytesRead is only called after the read() or readCombinedRange() loop finishes successfully. If the thread is interrupted midway, the bytes that were physically transferred from GCS into                
the buffer are not added to the STREAM_READ_BYTES counter

shameersss1: As discussed we will fix this seperately!

mudit-97 (approved):
LGTM

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -165,25 +169,49 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+    private void addCancellationListener(FileRange range, Future<?> executionFuture) {
+      range
+          .getData()
+          .whenComplete(
+              (res, ex) -> {
+                if (ex instanceof CancellationException) {
+                  logger.atWarning().log(
+                      "Cancelling execution future for range %s due to CancellationException",
+                      range);
+                  executionFuture.cancel(true);
```
singhravidutt, line 216: Why just cancel this specific range's execution and not all ranges which were submitted via single vectoredRead? Are we handling it other way round, as in caller triggers the cancellation?

shameersss1: Yes, The expectation here is that the caller needs to handle it and call the cancel

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -165,25 +169,49 @@ private void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuff
...
+                    // all required stats must be collected before task is finished.
+                    resetThreadLocalStats();
+                    readCombinedRange(combinedFileRange, allocate, channelProvider);
+                    long endTimer = System.currentTimeMillis();
+                    storageStatistics.updateStats(
+                        GhfsStatistic.STREAM_READ_VECTORED_READ_RANGE_DURATION,
+                        endTimer - startTimer,
+                        channelProvider.gcsPath);
+                  });
+          addCancellationListener(combinedFileRange, executionFuture);
```
singhravidutt, line 189: Combined File range is created in VectoredIo my combining the ranges, if caller initiate  cancellation on base range it will not get propagated to combined range future, will it?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -367,6 +370,27 @@ public int read(ByteBuffer buffer) throws IOException {
             "Closing contentChannel after %s exception for '%s'.", ioe.getMessage(), resourceId);
         closeContentChannel();
 
+        // Handle thread interruptions (e.g. from cancellation) separately from general
+        // IOExceptions.
+        // Interruptions are treated as non-retryable events because they indicate the caller has
+        // abandoned the current request (e.g., due to a timeout) and may already be initiating
+        // a new attempt. We catch InterruptedIOException but exclude SocketTimeoutException,
+        // which is a network-level timeout that should still be retried.
+        if (ioe instanceof ClosedByInterruptException
```
singhravidutt: Are we sure that the cancelling the future via `addCancellationListener` will actually trigger an interrupt here?

Mocking reader thread sleep, will it be a real wold scenario?Will thread waiting for network packets actually honor the interrupt flag?

shameersss1: Yes, It does. 

Image

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
shameersss1: Yes, it won't - We need to handle it. I will fix this in the next revision

shameersss1: Added in the addCancellationListener

singhravidutt (approved):

PR #1682, opened 2026-03-27
Implement fast fail for Bidi Read

dheerajsngh:
Fail-Fast implementation for Bidi Reads

More Details : go/bidi-fast-fail-connector


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -532,6 +568,64 @@ private DisposableByteString readBytes(long offset, long length)
...
+    if (metadataInitialized) {
+      return;
+    }
+
+    try {
+      if (this.sessionFuture != null) {
+        // Resolve the ApiFuture to get the actual BlobReadSession object.
+        // This will block until the session is successfully established,
+        // effectively deferring the connection latency to the first read().
+        BlobReadSession session = this.sessionFuture.get();
```
shrutisinghania: We should avoid using an unbounded `get()` here to prevent the reading thread from blocking indefinitely if the future hangs. 
Since we already have a configured timeout in `getBlobReadSession()`, let's apply the same timeout here.

shrutisinghania (changes requested):
Could we also update the documentation in the configuration file for this change.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
```diff
@@ -357,11 +583,21 @@ public void truncate_unsupportedOperationException() throws IOException {
   }
 
   @Test
-  public void initMetadata_unsupportedOperationException() throws IOException {
-    GoogleCloudStorageBidiReadChannel bidiReadChannel = getMockedBidiReadChannel();
+  public void initMetadata_whenGzipNotSupported_throwsIOException() throws Exception {
```
shrutisinghania: nit: The test method is named `initMetadata_whenGzipNotSupported_throwsIOException`, but we are actually asserting an `UnsupportedOperationException`. We should update the test name to match the assertion.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -532,6 +568,64 @@ private DisposableByteString readBytes(long offset, long length)
...
+          "Interrupted while waiting for BlobReadSession initialization for '%s'", resourceId);
+    } catch (Exception e) {
+      logger.atWarning().withCause(e).log(
+          "Failed to get metadata from BlobReadSession future for '%s'", resourceId);
+    }
+
+    // Fallback: If the session didn't have it, or the future failed to resolve, do the explicit
+    // network call
+    logger.atInfo().log("Fallback to explicit metadata fetch for '%s'", resourceId);
+    fetchMetadata();
```
shrutisinghania, line 605: Currently, if `sessionFuture.get()` throws an `InterruptedException`, we restore the thread's interrupt flag and log a warning, but then execution falls through to `fetchMetadata()` anyway.
Since `fetchMetadata()` initiates another blocking network call (`storage.get(...)`), this defeats the purpose of the interrupt (which usually indicates the thread or task is being cancelled and should stop immediately).

IMO, we should only use the fallback when the task itself fails (e.g., `ExecutionException` or `TimeoutException`), but we should abort immediately if the thread is interrupted.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -294,15 +313,32 @@ private void throwIfNotOpen() throws IOException {
...
+
     gzipEncoded = nullToEmpty(encoding).contains(GZIP_ENCODING);
     // TODO(dhritichopra) Add Support for GZIP Encoding
     if (gzipEncoded) {
       GoogleCloudStorageEventBus.postOnException();
       throw new UnsupportedOperationException("Gzip Encoded Files are not supported");
     }
-    objectSize = sizeFromMetadata;
+
+    objectSize = gzipEncoded ? -1 : sizeFromMetadata;
```
shrutisinghania: We already throw an `UnsupportedOperationException`  if `gzipEncoded` is `true`.
Execution will never reach this line if it is true, so we can safely revert it back to `objectSize = sizeFromMetadata`

shrutisinghania (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Updated. 
Thanks for pointing it out. Had refered to the gRPC and JSON implementation where reading gzip encoded file is supported while its not supported currently for Bidi.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Thanks....Agreed ...

Updated the code to throw an error on `InterruptedException`, fallback in case of other exception. This also aligns the implementation to JSON([code/](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/c8099157afe71e4c4d4f5e90cfb8824cc63cf185/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L238))

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannelTest.java
dheerajsngh: Updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Makes sense to make add a timeout. 
Updated. Thanks.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -49,42 +50,56 @@
 public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeekableByteChannel {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
   private static final int EOF = -1;
-  private final StorageResourceId resourceId;
+  private StorageResourceId resourceId;
   private final BlobId blobId;
```
shrutisinghania: The `blobId` field is final and is initialized in the constructor. If the channel is initialized without a generation ID and then learns one via `initMetadata()`, the `resourceId` field is updated, but the `blobId` is not. While the `sessionFuture` already uses the original `blobId`, any logic that might recreate a session later (e.g., in a retry scenario) would use the version without a generation ID, potentially losing read consistency.

shrutisinghania (changes requested):
Except for the `blobId` concern, the changes LGTM.

shrutisinghania:
nit: Currently, all unit tests for lazy metadata initialization trigger the initialization logic specifically by calling `channel.size()`.  Add at least one unit test where the first operation on the channel is a `read()` or `readVectored()` call, to verify that initialization works correctly when triggered by a data request

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: It was initially final. Had not evaluated this retry scenario, where we might be using stale blobInfo. Updated. 
Thanks for pointed it out.

shrutisinghania (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -49,42 +50,56 @@
...
     this.readOptions = readOptions;
+    this.boundedThreadPool = boundedThreadPool;
+    this.readTimeout = readOptions.getGrpcReadTimeout();
+
+    if (itemInfo != null) {
+      validate(itemInfo);
+      initMetadata(
+          itemInfo.getContentEncoding(), itemInfo.getSize(), itemInfo.getContentGeneration());
+    } else {
+      if (readOptions.isFastFailOnNotFoundEnabled()) {
```
animesh-g: nit: Use `else if` maybe?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -532,6 +569,67 @@ private DisposableByteString readBytes(long offset, long length)
...
+          return;
+        }
+      }
+    } catch (InterruptedException e) {
+      GoogleCloudStorageEventBus.postOnException();
+      Thread.currentThread().interrupt();
+      logger.atWarning().withCause(e).log(
+          "Interrupted while waiting for BlobReadSession initialization for '%s'", resourceId);
+      throw new IOException("Thread interrupt received.", e);
+    } catch (Exception e) {
```
animesh-g: Catching broad exception can cause issues. If it fails with `NullPointerException` or something similar, it can mask some legitimate bugs.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageClientInterceptorIntegrationTest.java
```diff
@@ -149,6 +149,45 @@ public void testWriteLogs() throws IOException {
...
+        GCS_TRACE_OPTIONS.toBuilder().setBidiEnabled(true).build();
+    GoogleCloudStorage gcsImpl = getGCSClientImpl(storageOption);
+    GoogleCloudStorageReadOptions readOptions =
+        GoogleCloudStorageReadOptions.builder().setFastFailOnNotFoundEnabled(false).build();
+    assertObjectContent(gcsImpl, resourceId, readOptions, partition, partitionsCount);
+    boolean hasGetObject = false;
+    boolean hasBidiRead = false;
+
+    // Use a try-catch loop to read all logs since getLogCount() is not available
+    int i = 0;
```
animesh-g: `getLogCount` isn't available but [assertLogCount](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/649dc1c4c8a73c1891dca2119733aa2547f0b8ba/gcsio/src/test/java/com/google/cloud/hadoop/gcsio/AssertingLogHandler.java#L63) seems to be available. Why cannot we use the same? Relying on exceptions to control the flow(here exiting the loop) is an antipattern thus should be avoided.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
animesh-g: Wouldn't it apply to 'ExecutionException' as well? `sessionFuture.get()` can throw errors if it receives 404 or other errors. We will mask error in catch and then fail again fetchMetadata().

animesh-g (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageClientInterceptorIntegrationTest.java
dheerajsngh: Updated the test.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Updated the code to catch underneath exceptions and handled them seperately. Added unit tests for each case.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: > Wouldn't it apply to 'ExecutionException' as well? `sessionFuture.get()` can throw errors if it receives 404 or other errors. We will mask error in catch and then fail again fetchMetadata().

Updated the code to check for the 404 errors during ExecutionException and throw the error, instead of falling back to the fetchMetadata.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -532,6 +568,82 @@ private DisposableByteString readBytes(long offset, long length)
...
+      if (this.sessionFuture != null) {
+        // Resolve the ApiFuture to get the actual BlobReadSession object.
+        // We use the configured BidiClientTimeout to prevent the reading thread
+        // from blocking indefinitely if the future hangs.
+        BlobReadSession session =
+            this.sessionFuture.get(readOptions.getBidiClientTimeout(), TimeUnit.SECONDS);
+
+        BlobInfo blobInfo = session.getBlobInfo();
+        if (blobInfo != null) {
+          initMetadata(blobInfo.getContentEncoding(), blobInfo.getSize(), blobInfo.getGeneration());
```
animesh-g: Should we synchronize ensureMetadataInitialized or at least this block to avoid double initialization? Otherwise we may end in exception due to `checkState` in `initMetadata` method.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -49,42 +51,54 @@
...
+  private StorageResourceId resourceId;
+  private BlobId blobId;
   private final ApiFuture<BlobReadSession> sessionFuture;
   private volatile BlobReadSession blobReadSession;
   private final ExecutorService boundedThreadPool;
   private static final String GZIP_ENCODING = "gzip";
-  private long objectSize;
+  private long objectSize = -1;
   private boolean open = true;
+  private boolean metadataInitialized = false;
```
animesh-g: Should we keep this variable `volatile` similar to itemInfo?

animesh-g (approved):
Overall looks good. Added a couple of suggestions related to thread safety.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Updated

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
dheerajsngh: Updated the code to wrap it with `synchronize` to ensure thread safety.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -569,54 +569,64 @@ private DisposableByteString readBytes(long offset, long length)
...
+              resourceId.getBucketName(), resourceId.getObjectName(), new IOException(e));
+        }
+        logger.atWarning().withCause(e).log(
+            "Failed to get metadata from BlobReadSession future for '%s'", resourceId);
+      } catch (TimeoutException e) {
+        GoogleCloudStorageEventBus.postOnException();
+        logger.atWarning().withCause(e).log(
+            "Timeout waiting for BlobReadSession initialization for '%s'", resourceId);
+      } catch (Exception e) {
+        logger.atWarning().withCause(e).log(
```
animesh-g: I think we can remove this as we have pretty good coverage of exceptions above. This will unnecessary  mask NPE or some similar fatal errors.

dheerajsngh: Updated.

PR #1685, opened 2026-04-08
Fast Fail Bug Fix during gRPC Reads

dheerajsngh:
Fast Fail Bug Fix during gRPC Reads

More Details : go/grpc-fast-fail-fix

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -404,8 +404,51 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
 
       setChannelBoundaries(bytesToRead);
 
+      logger.atFiner().log(
+          "Opening lazy gRPC channel for '%s' from %d to %d",
+          resourceId, contentChannelCurrentPosition, contentChannelEnd);
+
       ReadableByteChannel readableByteChannel =
           getStorageReadChannel(contentChannelCurrentPosition, contentChannelEnd);
+
+      if (!metadataInitialized) {
+        logger.atInfo().log(
+            "Metadata not initialized. Forcing lazy channel connection with 0-byte read for '%s'",
+            resourceId);
+
+        // The 0-byte read triggers the actual ReadObjectRequest over the network
+        readableByteChannel.read(ByteBuffer.allocate(0));
+        // Extract the headers to populate objectSize
+        ensureMetadataInitialized(readableByteChannel);
+
+        logger.atInfo().log(
+            "Metadata extracted! True objectSize is now %d for '%s'", objectSize, resourceId);
+
+        if (objectSize == 0) {
+          readableByteChannel.close();
+          return Channels.newChannel(new ByteArrayInputStream(new byte[0]));
+        }
+
+        // Check if file is GZIP Encoded
+        if (gzipEncoded) {
+          if (currentPosition == 0) {
+            contentChannelEnd = objectSize;
+          } else {
+            // We guessed the boundary wrong for a GZIP file. Reopen it properly.
+            readableByteChannel.close();
+            // Reset channel boundaries
+            contentChannelCurrentPosition = -1;
+            contentChannelEnd = -1;
+            return openByteChannel(bytesToRead);
+          }
+        }
+      }
+
+      if (contentChannelEnd == Long.MAX_VALUE && objectSize != -1) {
+        contentChannelEnd = getRangeRequestEnd(contentChannelCurrentPosition, bytesToRead);
+        logger.atInfo().log(
+            "Recalculated contentChannelEnd to %d for '%s'", contentChannelEnd, resourceId);
+      }
 
       if (contentChannelEnd == objectSize
           && (contentChannelEnd - contentChannelCurrentPosition)
@@ -461,7 +504,7 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
     }
 
     private void setChannelBoundaries(long bytesToRead) {
-      contentChannelCurrentPosition = getRangeRequestStart();
+      contentChannelCurrentPosition = getRangeRequestStart(bytesToRead);
       contentChannelEnd = getRangeRequestEnd(contentChannelCurrentPosition, bytesToRead);
       checkState(
           contentChannelEnd >= contentChannelCurrentPosition,
@@ -513,10 +556,25 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
       return Channels.newChannel(new ByteArrayInputStream(footerContent, offset, length));
     }
 
-    private long getRangeRequestStart() {
+    private long getRangeRequestStart(long bytesToRead) {
       if (gzipEncoded) {
         return 0;
       }
+
+      // Guess the start boundary if the size is unknown
+      if (objectSize == -1) {
+        if (readOptions.getFadvise() == Fadvise.SEQUENTIAL
+            || bytesToRead >= readOptions.getMinRangeRequestSize()) {
+          return currentPosition;
+        }
+        // Prefetch footer (bytes before 'currentPosition') lazily.
+        // Max prefetch size is (minRangeRequestSize / 2) bytes.
+        if (bytesToRead <= readOptions.getMinRangeRequestSize() / 2) {
+          return Math.max(0, currentPosition - readOptions.getMinRangeRequestSize() / 2);
+        }
+        return Math.max(0, currentPosition - (readOptions.getMinRangeRequestSize() - bytesToRead));
+      }
+
       if (readOptions.getFadvise() != Fadvise.SEQUENTIAL
           && isFooterRead()
           && !readOptions.isReadExactRequestedBytesEnabled()) {
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -477,18 +477,23 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
         return readableByteChannel;
 
       } catch (Exception e) {
-        // Safely close the orphaned channel BEFORE re-throwing the error
+        logger.atWarning().withCause(e).log(
+            "Exception occurred during metadata initialization for '%s'. Attempting to safely close orphaned channel.",
+            resourceId);
         try {
           if (readableByteChannel != null && readableByteChannel.isOpen()) {
             readableByteChannel.close();
+            logger.atInfo().log("Successfully closed orphaned channel for '%s'", resourceId);
           }
         } catch (IOException closeException) {
+          logger.atWarning().withCause(closeException).log(
+              "Failed to close orphaned channel for '%s' during cleanup", resourceId);
           e.addSuppressed(closeException);
         }
         if (e instanceof IOException) {
           throw (IOException) e;
         }
-        throw new RuntimeException(e);
+        throw new IOException("Unexpected error during channel initialization", e);
       }
     }
 
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -445,25 +445,14 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
           logger.atFiner().log(
               "Metadata extracted! True objectSize is now %d for '%s'", objectSize, resourceId);
 
-          if (objectSize != -1 && currentPosition >= objectSize) {
-            readableByteChannel.close();
-            contentChannelCurrentPosition = currentPosition;
-            contentChannelEnd = currentPosition;
-            return Channels.newChannel(new ByteArrayInputStream(new byte[0]));
+          ReadableByteChannel eofChannel = validateEndOfFile(readableByteChannel);
+          if (eofChannel != null) {
+            return eofChannel;
           }
 
-          // Check if file is GZIP Encoded
-          if (gzipEncoded) {
-            if (currentPosition == 0) {
-              contentChannelEnd = objectSize;
-            } else {
-              // We guessed the boundary wrong for a GZIP file. Reopen it properly.
-              readableByteChannel.close();
-              // Reset channel boundaries
-              contentChannelCurrentPosition = -1;
-              contentChannelEnd = -1;
-              return openByteChannel(bytesToRead);
-            }
+          ReadableByteChannel gzipChannel = handleGzipDiscovery(readableByteChannel, bytesToRead);
+          if (gzipChannel != null) {
+            return gzipChannel;
           }
         }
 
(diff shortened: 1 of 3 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -404,9 +404,52 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
+        // Extract the headers to populate objectSize
+        ensureMetadataInitialized(readableByteChannel);
+
+        logger.atInfo().log(
+            "Metadata extracted! True objectSize is now %d for '%s'", objectSize, resourceId);
+
+        if (objectSize == 0) {
+          readableByteChannel.close();
+          return Channels.newChannel(new ByteArrayInputStream(new byte[0]));
+        }
```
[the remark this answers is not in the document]
dheerajsngh: Resolved in later commit.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -404,9 +409,54 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
+      if (!metadataInitialized) {
+        logger.atInfo().log(
+            "Metadata not initialized. Forcing lazy channel connection with 0-byte read for '%s'",
+            resourceId);
+
+        // The 0-byte read triggers the actual ReadObjectRequest over the network
+        readableByteChannel.read(ByteBuffer.allocate(0));
+        // Extract the headers to populate objectSize
+        ensureMetadataInitialized(readableByteChannel);
+
```
shrutisinghania, line 428: If `readableByteChannel.read()` or `ensureMetadataInitialized()` throws an IOException (e.g., a connection error or NOT_FOUND), the `readableByteChannel` remains open and is leaked because it hasn't been assigned to the instance's `byteChannel` field yet. This initialization logic should be wrapped in a try-catch block to ensure the channel is closed on failure.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
shrutisinghania, line 395: As noted by the automated checks, `openByteChannel` exceeds the allowed cognitive complexity. The metadata initialization and validation logic can be extracted into a helper method.

shrutisinghania (changes requested):
AI suggested(needs to be validated): 

"The EOF check `currentPosition >= objectSize` added on line 301 and line 432 may cause premature termination when reading GZIP-encoded files. For compressed objects, GCS metadata provides the compressed size, but the client reads uncompressed data. Once the number of uncompressed bytes read reaches the compressed size, this check will trigger an EOF, even though more data remains in the stream.

In existing implementations within this codebase, the object size is typically set to `Long.MAX_VALUE` for GZIP files to avoid this issue:
 `size = gzipEncoded ? Long.MAX_VALUE : sizeFromMetadata;`

Please ensure that `ensureMetadataInitialized` follows this pattern or that the EOF check is skipped when `gzipEncoded` is true."

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Updated the code to wrap it in a try and catch block.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Moved the logic into a seperate function.

dheerajsngh:
> AI suggested(needs to be validated):
> 
> "The EOF check `currentPosition >= objectSize` added on line 301 and line 432 may cause premature termination when reading GZIP-encoded files. For compressed objects, GCS metadata provides the compressed size, but the client reads uncompressed data. Once the number of uncompressed bytes read reaches the compressed size, this check will trigger an EOF, even though more data remains in the stream.
> 
> In existing implementations within this codebase, the object size is typically set to `Long.MAX_VALUE` for GZIP files to avoid this issue: `size = gzipEncoded ? Long.MAX_VALUE : sizeFromMetadata;`
> 
> Please ensure that `ensureMetadataInitialized` follows this pattern or that the EOF check is skipped when `gzipEncoded` is true."

The above suggestion is not valid here. For `gzip`, we are initializing the `objectSize` to be -1. 
For all the `currentPosition >= objectSize` checks, it is preceded by `objectSize!=-1`, hence it will not cause premature termination for GZIP-encoded files.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageClientInterceptorIntegrationTest.java
```diff
@@ -209,6 +211,63 @@ public void testReadLogs() throws IOException {
...
+          getObjectCount++;
+        }
+
+        // ONLY count when the stream is initially opened ("request"),
+        // and ignore "response" and "onClose".
+        if (method != null && method.contains("ReadObject") && "request".equals(streamOp)) {
+          readObjectCount++;
+        }
+        i++;
+      } catch (AssertionError | IndexOutOfBoundsException e) {
```
animesh-g: As mentioned in the other PR, using exception to control the flow is antipattern. Try to avoid the same using getLogCount or assertLogCount.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -419,6 +430,73 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
+          if (gzipEncoded) {
+            if (currentPosition == 0) {
+              contentChannelEnd = objectSize;
+            } else {
+              // We guessed the boundary wrong for a GZIP file. Reopen it properly.
+              readableByteChannel.close();
+              // Reset channel boundaries
+              contentChannelCurrentPosition = -1;
+              contentChannelEnd = -1;
+              return openByteChannel(bytesToRead);
```
animesh-g: Is this recursive call intended i.e. openByteChannel -> initializeMetadataAndValidateChannel -> openByteChannel ? AFAIK gzip don't need fast fail handling so this logic can be moved to  openByteChannel itself.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -419,6 +430,73 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
+
+        if (contentChannelEnd == Long.MAX_VALUE && objectSize != -1) {
+          contentChannelEnd = getRangeRequestEnd(contentChannelCurrentPosition, bytesToRead);
+          logger.atInfo().log(
+              "Recalculated contentChannelEnd to %d for '%s'", contentChannelEnd, resourceId);
+        }
+
+        return readableByteChannel;
+
+      } catch (Exception e) {
```
animesh-g: Mentioned in the other PR as well. Broad Exception catching can lead to multiple issues.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -404,9 +409,15 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
       setChannelBoundaries(bytesToRead);
 
+      logger.atFiner().log(
+          "Opening lazy gRPC channel for '%s' from %d to %d",
+          resourceId, contentChannelCurrentPosition, contentChannelEnd);
+
       ReadableByteChannel readableByteChannel =
           getStorageReadChannel(contentChannelCurrentPosition, contentChannelEnd);
 
+      readableByteChannel = initializeMetadataAndValidateChannel(readableByteChannel, bytesToRead);
```
animesh-g: readableByteChannel was initialized above and got reassigned again making the above line a dead code. Is this intended?
If yes, please initialize it in the above line itself and consider marking it final as well.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannelTest.java
```diff
@@ -772,6 +774,161 @@ public void testChunkedRead_whenObjectSizeUnknown_doesNotTruncateSize() throws I
...
+    ByteBuffer buffer = ByteBuffer.allocate(10);
+
+    int bytesRead = readChannel.read(buffer);
+
+    assertThat(bytesRead).isEqualTo(0);
+    assertThat(readChannel.size()).isEqualTo(0L);
+  }
+
+  @Test
+  public void read_whenLazyInitFails_closesOrphanedChannel() throws Exception {
```
animesh-g: nit: Please use consistent naming for tests. Other added tests have 'test' prefix while this one doesn't.

animesh-g (changes requested):

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannelTest.java
dheerajsngh: Done.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Actually in the first assignment, it is just the lazy initialization and then in the later assignment, we are doing the actual initialization. 

Updated the code to make, so that it doesn't look like a dead code.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Updated the catch block to explicitly handle `IOException` and `RuntimeException` instead of a broad Exception. The caught exception is now re-thrown directly, as the parent method handles it.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Yes, this recursive call is intentional. 

This  is for handling the case, when fast fail is disabled, and we discover later that its gzip file, hence we need to restart the channel as we had earlier started from a non-zero start index and gzip does not support non-zero start index. 

This is similar pattern which is done in JSON as well([link/](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/649dc1c4c8a73c1891dca2119733aa2547f0b8ba/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L998))

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/integration/GoogleCloudStorageClientInterceptorIntegrationTest.java
dheerajsngh: Updated.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
animesh-g, line 1: Chnages are fine however logger.atInfo seems high for some of the logs. Consider Fine/Finer for some of them.

animesh-g (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: Updated

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -419,6 +429,72 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
       return readableByteChannel;
     }
 
+    private ReadableByteChannel initializeMetadataAndValidateChannel(
```
shrutisinghania, line 455: Consider refactoring this method by extracting the GZIP discovery logic and the EOF validation into smaller, private helper methods (e.g., `handleGzipDiscovery` and `validateEndOfFile`). This would align with the principle of keeping functions small for readability and directly address the SonarCloud check failure regarding Cognitive Complexity.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -419,6 +429,72 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
...
+          if (gzipEncoded) {
+            if (currentPosition == 0) {
+              contentChannelEnd = objectSize;
+            } else {
+              // We guessed the boundary wrong for a GZIP file. Reopen it properly.
+              readableByteChannel.close();
+              // Reset channel boundaries
+              contentChannelCurrentPosition = -1;
+              contentChannelEnd = -1;
+              return openByteChannel(bytesToRead);
```
shrutisinghania: Since the else block ends with a return statement we can simplify the control flow by using an early return pattern. This makes the code "flatter" and easier to read. For example:
```
// Check if file is GZIP Encoded
if (gzipEncoded) {
  if (currentPosition != 0) {
    // We guessed the boundary wrong for a GZIP file. Reopen it properly.
    readableByteChannel.close();
    // Reset channel boundaries
    contentChannelCurrentPosition = -1;
    contentChannelEnd = -1;
    return openByteChannel(bytesToRead);
  }
  // At currentPosition == 0, we can just set the end boundary
  contentChannelEnd = objectSize;
}

**Review by shrutisinghania — APPROVED** [review]

File: gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java (hunk 8 above)
Comment (dheerajsngh) [reply gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java:455]: Done

File: gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java (hunk 9 above)
Comment (dheerajsngh) [reply gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java]: Done

File: gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java [hunk 10]
```diff
@@ -348,12 +359,19 @@ public int readContent(ByteBuffer dst) throws IOException {
...
+              // Only throw the overshoot exception if we actually read extra bytes.
+              if (totalBytesRead > 0) {
+                GoogleCloudStorageEventBus.postOnException();
+                throw new IOException(
+                    String.format(
+                        "Received end of stream result beyond the object size; at offset: %d "
+                            + "whereas stream was supposed to end at: %d for resource: %s of size: %d",
+                        currentPosition, contentChannelEnd, resourceId, objectSize));
+              } else {
+                // Cleanly exit for seek past EOF
```
animesh-g: As discussed, there should be an exception here as well.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -297,6 +297,17 @@ public int readContent(ByteBuffer dst) throws IOException {
...
+        if (objectSize != -1 && currentPosition >= objectSize) {
+          // Only break if we haven't read any bytes yet (a clean seek past EOF)
+          // OR if we are exactly at the object boundary.
+          if (totalBytesRead == 0 || currentPosition == objectSize) {
+            break;
+          }
+          // If totalBytesRead > 0 AND currentPosition > objectSize, do NOT break.
+          // We overshot mid-read. Let the channel read '-1' so the overshoot handler can run.
+        }
+
```
animesh-g: We may not need this block.

dheerajsngh: Removed the code. We don´t need this. 
The code was for cases, where we user seeked >= objectSize. 

It is now handled in `validateEndOfFile` function.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
dheerajsngh: The above logic was for handling the case where the user seeked beyond the object size. 
Moved that check into `validateEndOfFile` and throwing EOF.

PR #1687, opened 2026-04-09
Add GCS Analytics Core dependency and configuration support

shrutisinghania:


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -625,6 +625,10 @@ public class GoogleHadoopFileSystemConfiguration {
               "fs.gs.bidi.finalize.on.close",
               GoogleCloudStorageOptions.DEFAULT.isFinalizeBeforeClose());
 
+  /** Configuration key for enabling GCS Analytics Core. */
+  public static final HadoopConfigurationProperty<Boolean> GCS_ANALYTICS_CORE_ENABLE =
```
dheerajsngh, line 629: We can discuss, if we just want a single flag or different flags for read and write. 
For now, we can submit the change, since its on a feature branch.

dheerajsngh (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
animesh-g: Let's have a single flag to avoid confusion. We can add more if needed.

gcs/pom.xml
```diff
@@ -84,6 +84,10 @@
       <artifactId>gcsio</artifactId>
       <version>${bigdataoss.version}</version>
     </dependency>
+    <dependency>
```
animesh-g, line 87: Can you add a task in backlog to consider using BOM for analytics-core as well? It will be easier to manage SDK and other dependencies

PR #1693, opened 2026-04-20
Add Integration Tests for analytics core read path migration

shrutisinghania:
This PR adds integration tests for the Analytics Core read path migration and parameterizes existing tests for both enabled and disabled configurations. It also refines seek and range validation logic to handle edge cases like unknown file sizes and improves telemetry by tracking async read durations and GCS request statistics.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -36,12 +36,12 @@ class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
 
   private final long size;
   private final GoogleCloudStorageInputStream inputStream;
-  private boolean open;
+  private boolean channelIsOpen;
 
   GcsAnalyticsCoreInputStreamWrapper(GoogleCloudStorageInputStream inputStream, long size) {
     this.inputStream = inputStream;
     this.size = size;
-    this.open = true;
+    this.channelIsOpen = true;
   }
 
   @Override
@@ -89,14 +89,14 @@ class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
 
   @Override
   public boolean isOpen() {
-    return open;
+    return channelIsOpen;
   }
 
   @Override
   public void close() throws IOException {
-    if (open) {
+    if (channelIsOpen) {
       inputStream.close();
-      open = false;
+      channelIsOpen = false;
     }
   }
 
@@ -112,7 +112,7 @@ class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
   }
 
   private void checkOpen() throws ClosedChannelException {
-    if (!open) {
+    if (!channelIsOpen) {
       throw new ClosedChannelException();
     }
   }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -30,6 +30,9 @@ import java.util.function.IntFunction;
 /**
  * Adapter to expose {@link GoogleCloudStorageInputStream} from Analytics Core library as {@link
  * SeekableByteChannel} for use in {@link GoogleHadoopFSInputStream}.
+ *
+ * <p>TODO(user): Wrap exception-throwing calls to post exceptions to GoogleCloudStorageEventBus on
+ * failure when adding comprehensive logging.
  */
 class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
 
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -79,6 +79,9 @@ class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
       throw new EOFException(
           "Invalid seek offset: position value (" + newPosition + ") must be >= 0");
     }
+    // Throwing EOFException aligns with the Hadoop spec for seek >= len(data).
+    // See:
+    // https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/filesystem/fsdatainputstream.html#Seekable.seek.28s.29
     if (size >= 0 && newPosition >= size) {
       GoogleCloudStorageEventBus.postOnException();
       throw new EOFException(
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -32,13 +32,13 @@ import java.util.function.IntFunction;
  * Adapter to expose {@link GoogleCloudStorageInputStream} from Analytics Core library as {@link
  * SeekableByteChannel} for use in {@link GoogleHadoopFSInputStream}.
  */
-class AnalyticsCoreChannelAdapter implements SeekableByteChannel {
+class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
 
   private final long size;
   private final GoogleCloudStorageInputStream inputStream;
   private boolean open;
 
-  AnalyticsCoreChannelAdapter(GoogleCloudStorageInputStream inputStream, long size) {
+  GcsAnalyticsCoreInputStreamWrapper(GoogleCloudStorageInputStream inputStream, long size) {
     this.inputStream = inputStream;
     this.size = size;
     this.open = true;
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -19,6 +19,7 @@ package com.google.cloud.hadoop.fs.gcs;
 import com.google.cloud.gcs.analyticscore.client.GcsObjectRange;
 import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageInputStream;
 import com.google.cloud.hadoop.util.GoogleCloudStorageEventBus;
+import java.io.EOFException;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
@@ -74,12 +75,12 @@ class GcsAnalyticsCoreInputStreamWrapper implements SeekableByteChannel {
     checkOpen();
     if (newPosition < 0) {
       GoogleCloudStorageEventBus.postOnException();
-      throw new java.io.EOFException(
+      throw new EOFException(
           "Invalid seek offset: position value (" + newPosition + ") must be >= 0");
     }
     if (size >= 0 && newPosition >= size) {
       GoogleCloudStorageEventBus.postOnException();
-      throw new java.io.EOFException(
+      throw new EOFException(
           "Invalid seek offset: position value ("
               + newPosition
               + ") must be between 0 and "
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
@@ -25,6 +25,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.NonWritableChannelException;
 import java.nio.channels.SeekableByteChannel;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.function.IntFunction;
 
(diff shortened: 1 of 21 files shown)


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  @Test
+  public void seek_beyondFileSize_throwsEOFException() throws Exception {
+    String testContent = "test content";
+    TestFixture fixture = setupFixture("seek_beyond_file", testContent);
+
+    FileSystem.Statistics statistics = new FileSystem.Statistics(fixture.ghfs.getScheme());
+    try (GoogleHadoopFSInputStream in =
```
dheerajsngh: Can we use the helper method `createGhfsInputStream` at all the places?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+    Configuration config = GoogleHadoopFileSystemIntegrationHelper.getTestConfig();
+    config.setBoolean("fs.gs.analytics.core.enable", true);
+
+    GoogleHadoopFileSystem ghfs = GoogleHadoopFileSystemIntegrationHelper.createGhfs(path, config);
+    gcsFsIHelper.writeTextFile(path, content);
+    return new TestFixture(ghfs, path, content);
+  }
+
+  @Test
+  public void read_singleBytes_returnsData() throws Exception {
```
dheerajsngh, line 71: Currently we have test for adding single byte.
* Can we also add test to reading larger chunk `read_buffer_returnsData`
* Also a test for the case where the input buffer is already full.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+    gcsFsIHelper = GoogleCloudStorageFileSystemIntegrationHelper.create();
+    gcsFsIHelper.beforeAllTests();
+  }
+
+  @AfterClass
+  public static void afterClass() {
+    gcsFsIHelper.afterAllTests();
+  }
+
+  private static class TestFixture {
```
dheerajsngh: We can move this into `GoogleHadoopFileSystemIntegrationHelper` class.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
```diff
@@ -70,7 +69,19 @@ public long position() throws IOException {
   @Override
   public SeekableByteChannel position(long newPosition) throws IOException {
     checkOpen();
-    checkArgument(newPosition >= 0 && newPosition <= size, "Invalid position: %s", newPosition);
+    if (newPosition < 0) {
```
dheerajsngh, line 77: This is an additional change apart from the integration tests, lets update the PR description for it.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
```diff
@@ -70,7 +69,19 @@ public long position() throws IOException {
...
+      GoogleCloudStorageEventBus.postOnException();
```
dheerajsngh, line 78: Similar to this,we should also post the exception in the GCSEventBus, at other places like `checkOpen`, all the reads, `close`. 

I guess those will be added when we will add all the logging.In that case we should have TO DO added for them.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
dheerajsngh, line 1: Simiar comment for the test structure as previous PR. 
Follow Arrange, Act and Assert and seperate them with blank lines.

dheerajsngh (changes requested):
Overall Looks Good.
Added comments for adding some more tests. 

Also can we add additional test in `GoogleHadoopFSInputStreamIntegrationTest` to tests the behavior with and without the analytics-core flag.

dheerajsngh:
> Overall Looks Good. Added comments for adding some more tests.
> 
> Also can we add additional test in `GoogleHadoopFSInputStreamIntegrationTest` to tests the behavior with and without the analytics-core flag.

Looks like we are updating GoogleHadoopFSInputStreamIntegrationTest in the follow up PR : https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1701.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
shrutisinghania: Used `createGhfsInputStream` at all the applicable places

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
shrutisinghania: Added cases for these scenarios as well

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
shrutisinghania: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
shrutisinghania: Updated the PR description

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
shrutisinghania: Added a TODO

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStreamAnalyticsIntegrationTest.java
shrutisinghania: Done

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
```diff
@@ -31,6 +31,9 @@
 /**
  * Adapter to expose {@link GoogleCloudStorageInputStream} from Analytics Core library as {@link
  * SeekableByteChannel} for use in {@link GoogleHadoopFSInputStream}.
+ *
+ * <p>TODO(user): Wrap exception-throwing calls to post exceptions to GoogleCloudStorageEventBus on
```
dheerajsngh, line 36: nit: add username

dheerajsngh (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
```diff
@@ -70,7 +73,19 @@ public long position() throws IOException {
   @Override
   public SeekableByteChannel position(long newPosition) throws IOException {
     checkOpen();
-    checkArgument(newPosition >= 0 && newPosition <= size, "Invalid position: %s", newPosition);
+    if (newPosition < 0) {
+      GoogleCloudStorageEventBus.postOnException();
+      throw new EOFException(
+          "Invalid seek offset: position value (" + newPosition + ") must be >= 0");
+    }
+    if (size >= 0 && newPosition >= size) {
```
animesh-g, line 85: In previous state, newPosition == size was a valid case but here we are throwing exception. Is it intentional?

animesh-g (changes requested):
I was reviewing this and realized there are redundant tests in GoogleHadoopFSInputStreamAnalyticsIntegrationTest . I think https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1701 is the right way to handle this. Let's cancel one of PR and merge changes into one for easier review and maintenance.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
shrutisinghania: Yes, this change is intentional. It was modified to align with the behavior of existing implementations like [GoogleCloudStorageReadChannel](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L695) and [GoogleCloudStorageClientReadChannel](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java#L755).

shrutisinghania:
> I was reviewing this and realized there are redundant tests in GoogleHadoopFSInputStreamAnalyticsIntegrationTest . I think #1701 is the right way to handle this. Let's cancel one of PR and merge changes into one for easier review and maintenance.

Integrated changes from PR #1701 into this PR

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -221,17 +220,120 @@ private GoogleHadoopFSInputStream(
...
+                    GcsObjectRange.builder()
+                        .setLength(range.getLength())
+                        .setOffset(range.getOffset())
+                        .setByteBufferFuture(range.getData())
+                        .build())
+            .collect(Collectors.toList());
+    analyticsChannel.readVectored(gcsObjectRanges, allocate);
+    // Update metrics manually as Analytics Core bypasses the standard GCS event
+    // bus.
+    storageStatistics.incrementGcsTotalRequestCount();
```
animesh-g: The Analytics Core library uses intelligent buffering, prefetching and range merging etc. Incrementing the request count on every read() call will lead to inaccurate metrics, reporting multiple GCS requests for a single actual request. I think it should be reported by analytics-core itself.
Applies to the other places in the file as well.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
animesh-g: I am fine with the current implementation but [HDFS implementation](https://github.com/apache/hadoop/blob/da7a26b39de4791d0d73324d8ee1ff6034df9cdd/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSInputStream.java#L1651) doesn't throw EOF exception on targetPos==blockEnd. However the above implementation is aligned with [Hadoop spec](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/filesystem/fsdatainputstream.html#Seekable.seek.28s.29) itself. Can you add a comment with the rationale here to avoid confusion in future?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -221,17 +220,120 @@ private GoogleHadoopFSInputStream(
...
+    storageStatistics.incrementCounter(GoogleCloudStorageStatistics.GCS_GET_MEDIA_REQUEST, 1);
```
animesh-g: nit: I think it is recording metric regardless of success or failure. Please confirm if metrics `GCS_GET_MEDIA_REQUEST` expects the same behavior.

animesh-g (commented):
Integration suite looks good. I had a couple of concerns with telemetry.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
shrutisinghania: As discussed offline, removed the logic to manually update metrics and ignored the tests for metric check

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
shrutisinghania: Removed this metric

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
shrutisinghania: Added a comment

animesh-g (approved):

animesh-g (approved):

PR #1694, opened 2026-04-20
Log and drop while receiving the additional bytes from server

animesh-g:
As discussed in the [internal doc](http://shortn/_mfCylknbNH), in a very rare cases GCS can send more bytes than requested. Even though it should not be a fatal error from the client perspective, connector still throws error. This PR fixes the same scenario by adding warning log instead of throwing an exception as well as dropping the additional bytes. It also corrects the current log messages as they were slightly misleading. The fix is applied to both JSON and gRPC path. BiDi readchannel is not applicable here as it can move in both directions and we haven't observed this issue for BiDi.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -359,6 +359,11 @@ class GoogleCloudStorageClientReadChannel implements SeekableByteChannel {
                   "Received end of stream result after the channel end; at offset: %d "
                       + "whereas stream was suppose to end at: %d for resource: %s of size: %d",
                   currentPosition, contentChannelEnd, resourceId, objectSize);
+              // Dropping additional bytes.
+              int overshoot = (int) (currentPosition - contentChannelEnd);
+              dst.position(dst.position() - overshoot);
+              currentPosition = contentChannelEnd;
+
               closeContentChannel();
               continue;
             }
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
@@ -25,13 +25,18 @@ import static java.lang.Math.max;
 import static java.lang.Math.min;
 import static java.lang.Math.toIntExact;
 
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageReadOptions.Fadvise;
 import com.google.cloud.hadoop.util.ErrorTypeExtractor;
 import com.google.cloud.hadoop.util.GoogleCloudStorageEventBus;
+import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
+import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.Storage.BlobSourceOption;
+import com.google.cloud.storage.StorageException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.flogger.GoogleLogger;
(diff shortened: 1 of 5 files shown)


animesh-g:
@ajayky-os reminder to look into this.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
```diff
@@ -407,6 +430,10 @@ private ReadableByteChannel openByteChannel(long bytesToRead) throws IOException
       ReadableByteChannel readableByteChannel =
           getStorageReadChannel(contentChannelCurrentPosition, contentChannelEnd);
 
+      logger.atFiner().log(
```
ajayky-os, line 433: do we need this information ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -325,14 +325,37 @@ public int read(ByteBuffer buffer) throws IOException {
           // actual size of the data stream when stream compression is used, so we can
           // only ignore
           // this case here.
-          if (currentPosition != contentChannelEnd && currentPosition != size) {
+          if (currentPosition < contentChannelEnd && currentPosition < size) {
```
ajayky-os: both the GRPC & JSON code are identical, consider moving it to common place ?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
```diff
@@ -325,14 +325,37 @@ public int read(ByteBuffer buffer) throws IOException {
...
+                currentPosition, contentChannelEnd, resourceId, size);
+
+            // Dropping additional bytes.
+            int overshoot = (int) (currentPosition - contentChannelEnd);
+            buffer.position(buffer.position() - overshoot);
+            currentPosition = contentChannelEnd;
+            totalBytesRead -= overshoot;
+            contentChannelPosition -= overshoot;
+
+            closeContentChannel();
```
ajayky-os: This `continue` and `closeContentChannel` can be simplified. 

Line 364 always executes if we don't do continue in this block, we can remove line 357 & 358 and still the behaviour will be same.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
ajayky-os: Optional:

Conditions can be organised as :
```
if (currentPosition > contentChannelEnd && currentPosition < size) {
...
} else {
...
}
```
IMO this simplifies effort to understand the condition.

ajayky-os (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
animesh-g: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
animesh-g: Done

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
animesh-g: Overall both channels shared a lot of common code and should be refactored. Analytics-core migration is already consolidating these two so we don't need to handle right away.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientReadChannel.java
animesh-g: It is intentional. We need logs to identify and debug this issue if it happens again in the future.

PR #1697, opened 2026-04-24
chore: update base image to trusted openjdk 11 in Dockerfile

suni72:


cloudbuild/Dockerfile
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 # This Dockerfile creates an image for running presubmit tests.
-FROM eclipse-temurin:11-jdk
+# The image below is for OpenJDK 11.
+FROM us-docker.pkg.dev/artifact-foundry-prod/docker-3p-trusted/openjdk@sha256:e81b7f317654b0f26d3993e014b04bcb29250339b11b9de41e130feecd4cd43c
```
animesh-g, line 17: Why we don't need curl and wget here?

suni72: for jdk 17, only the slim version was available. So few things like `org.apache.maven.wrapper.MavenWrapperMain` were not present. hence had to download them separately

animesh-g (approved):

PR #1701, opened 2026-04-27
Parameterize the existing Integration Tests for read path

shrutisinghania:
This PR parameterizes `GoogleHadoopFSInputStreamIntegrationTest` to verify the read path with Analytics Core both enabled and disabled. It also introduces manual metric updates and range validation to ensure consistency.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
@@ -32,6 +32,7 @@ import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageInputStream;
 import com.google.cloud.hadoop.gcsio.FileInfo;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystem;
 import com.google.cloud.hadoop.gcsio.GoogleCloudStorageFileSystemOptions;
+import com.google.cloud.hadoop.gcsio.GoogleCloudStorageStatistics;
 import com.google.cloud.hadoop.gcsio.ReadVectoredSeekableByteChannel;
 import com.google.cloud.hadoop.gcsio.StorageResourceId;
 import com.google.cloud.hadoop.gcsio.VectoredIORange;
@@ -223,6 +224,7 @@ class GoogleHadoopFSInputStream extends FSInputStream implements IOStatisticsSou
           streamStatistics,
           STREAM_READ_VECTORED_OPERATIONS.getSymbol(),
           () -> {
+            long startTimeNs = System.nanoTime();
             if (channel instanceof ReadVectoredSeekableByteChannel) {
               ReadVectoredSeekableByteChannel readVectoredSeekableByteChannelChannel =
                   (ReadVectoredSeekableByteChannel) channel;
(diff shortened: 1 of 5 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -319,6 +335,11 @@ public synchronized int read(@Nonnull byte[] buf, int offset, int length) throws
             // the underlying channel.
             int numRead = channel.read(ByteBuffer.wrap(buf, offset, length));
             if (numRead > 0) {
+              if (channel instanceof AnalyticsCoreChannelAdapter) {
```
dheerajsngh: Why this check for `AnalyticsCoreChannelAdapter`? Shouldn't this be updated regardless if it is the Adpater class? 
Or for the case when it is not the adapter class, the metrics are already updated? In that case we can add a comment stating the same.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
```diff
@@ -114,7 +116,19 @@ public void close() throws IOException {
   public void readVectored(List<GcsObjectRange> ranges, IntFunction<ByteBuffer> allocate)
       throws IOException {
     checkOpen();
-    inputStream.readVectored(ranges, allocate);
+    List<GcsObjectRange> validRanges = new ArrayList<>();
```
dheerajsngh, line 122: This seems to be additional change in the logic. 
Lets add this in the PR description.

Can we also add a test for this case?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
```diff
@@ -242,15 +244,24 @@ public void readVectored(List<? extends FileRange> ranges, IntFunction<ByteBuffe
                                   .build())
                       .collect(Collectors.toList()),
                   allocate);
+              vectoredReadStats.updateVectoredReadStreamStats(startTimeNs);
```
dheerajsngh: This will measure the `synchronous submission time` not the actual vectored read time. 
Since `readVectored` is an async call. we should wait for the future to complete, to calculate the actual time.

We can do something like this:

```
// Gather all the futures from the ranges into an array
CompletableFuture<?>[] allFutures = ranges.stream()
    .map(FileRange::getData)
    .toArray(CompletableFuture[]::new);

// Wait for all of them to complete, then update the stats
CompletableFuture.allOf(allFutures).whenComplete((ignoredResult, ex) -> {
    vectoredReadStats.updateVectoredReadStreamStats(startTimeNs);
});
```

I am okay with this being done a follow up as well, since this the current state for the other cases as well.

dheerajsngh (commented):
Overall Look good. 
Added comment for updating how we measure vectored read time. 
We can do it in a follow up.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
shrutisinghania: Standard channels automatically update metrics via client event-bus listeners on standard HTTP/gRPC requests, whereas Analytics Core bypasses this transport layer, requiring manual updates to ensure they are tracked. Added a comment for the same.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFSInputStream.java
shrutisinghania: Thanks for pointing it out. `vectoredReadStats` is not thread-safe, and updating it from the async callback running on a pool thread will cause data races. Also, if the stream is closed before the reads finish, `vectoredReadStats.close()` will reset the state, leading to lost metrics. 

To fix this safely, I bypassed `vectoredReadStats` and used the thread-safe `DurationTracker` and direct global `storageStatistics` updates to record the async duration without race conditions.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreInputStreamWrapper.java
shrutisinghania: Updated the PR description and added tests for the same

shrutisinghania:
Consolidating these changes into PR #1693 to streamline the review process as suggested.

PR #1705, opened 2026-04-30
Map Connector configurations to Analytics Core

shrutisinghania:
This PR introduces the `AnalyticsCoreConfigMapper` class to handle the mapping of Hadoop Connector configurations to Analytics Core configurations. Applicable configs with direct 1:1 mapping, as documented in the [internal doc](https://docs.google.com/document/d/1yRc5Y7NDeXx4aQJYy7s_I4KDpqCdGzAFDkvAGCriksA/edit?tab=t.dp6btvyilmq4#bookmark=id.o300iiy0ehsr), have been implemented.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.hadoop.fs.gcs;
 
+import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import org.apache.hadoop.conf.Configuration;
 
@@ -27,6 +28,23 @@ final class AnalyticsCoreConfigMapper {
   static final String MAX_MERGE_GAP_KEY = "analytics-core.read.vectored.range.merge-gap.max-bytes";
   static final String MAX_MERGE_SIZE_KEY =
       "analytics-core.read.vectored.range.merged-size.max-bytes";
+
+  private static final ImmutableMap<String, String> HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS =
+      ImmutableMap.<String, String>builder()
+          .put(GoogleHadoopFileSystemConfiguration.GCS_PROJECT_ID.getKey(), PROJECT_ID_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_REQUESTER_PAYS_PROJECT_ID.getKey(),
+              USER_PROJECT_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_THREADS.getKey(),
+              READ_THREAD_COUNT_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_RANGE_MIN_SEEK.getKey(),
+              MAX_MERGE_GAP_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE.getKey(),
+              MAX_MERGE_SIZE_KEY)
+          .build();
 
   private AnalyticsCoreConfigMapper() {
     // Utility class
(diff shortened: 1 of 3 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Maps configurations from Hadoop Configuration to a map suitable for Analytics Core.
+   *
+   * @param config The Hadoop configuration.
+   * @param prefix The prefix used for Analytics Core properties (e.g., "fs.gs.").
+   * @return A map containing the mapped properties.
+   */
+  static Map<String, String> mapConfigs(Configuration config, String prefix) {
```
animesh-g: We can simplify this method. We can declare a map in following way,
```
  private static final ImmutableMap<String, String> HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS =
      ImmutableMap.of(
          GoogleHadoopFileSystemConfiguration.GCS_PROJECT_ID.getKey(),
          PROJECT_ID_KEY,
          GoogleHadoopFileSystemConfiguration.GCS_REQUESTER_PAYS_PROJECT_ID.getKey(),
          USER_PROJECT_KEY,
          GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_THREADS.getKey(),
          READ_THREAD_COUNT_KEY,
          GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_RANGE_MIN_SEEK.getKey(),
          MAX_MERGE_GAP_KEY,
          GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE.getKey(),
          MAX_MERGE_SIZE_KEY);

```

Then use a forEach loop to call mapAndRemoveSource ,

```
  static Map<String, String> mapConfigs(Configuration config, String prefix) {
    Map<String, String> mappedProperties = config.getValByRegex("^" + prefix.replace(".", "\\."));

    // Direct 1:1 mappings from Connector to Analytics Core
    HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS.forEach(
        (hadoopKey, analyticsKey) ->
            mapAndRemoveSource(hadoopKey, mappedProperties, prefix + analyticsKey));

    return mappedProperties;
  }
```

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapperTest.java
```diff
@@ -0,0 +1,99 @@
...
+public class AnalyticsCoreConfigMapperTest {
+
+  @Test
+  public void mapConfigs_mapsConnectorPropertiesToAnalyticsCore() {
+    Configuration config = new Configuration();
+    config.set("fs.gs.project.id", "my-project");
+    config.set("fs.gs.requester.pays.project.id", "user-project");
+    config.set("fs.gs.vectored.read.threads", "10");
+    config.set("fs.gs.vectored.read.min.range.seek.size", "1024");
+    config.set("fs.gs.vectored.read.merged.range.max.size", "2048");
```
animesh-g: Avoid using harcoded strings here. Try using keys directly from GoogleHadoopFileSystemConfiguration .

animesh-g (approved):
Overall looks good. Added a couple of comment for restructuring the code.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
shrutisinghania: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -0,0 +1,78 @@
...
+    Map<String, String> mappedProperties = config.getValByRegex("^" + prefix.replace(".", "\\."));
```
dheerajsngh: `config.getValByRegex(String regex)` scans all the currently set properties in the Configuration object and returns a Map<String, String> containing only the key-value pairs whose keys match the given regular expression ([source/](https://github.com/apache/hadoop/blob/4b224cdedcf585f95305dc04379ab9a69c2b951b/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L3969)).

The issue with this approach is that the connector has custom logic for calculating the values of certain properties—like [vectored read threads](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/f834c65be621051597794e70befd3dc1c5166e75/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L652) and [Requester Pays Project ID](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/f834c65be621051597794e70befd3dc1c5166e75/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L802)—in case the flags are not explicitly set by the user.

Currently, we are only scanning and mapping flags that have been explicitly set. This creates discrepancies between the connector's intended default values and the Analytics Core values. We should explicitly resolve these specific dynamic properties using the connector's existing logic before executing the standard 1:1 mapping.

Something like this : 

```
  static Map<String, String> mapConfigs(Configuration config, String prefix) {
    Map<String, String> mappedProperties = config.getValByRegex("^" + prefix.replace(".", "\\."));

    // 1. Leverage existing builders to resolve dynamic values and fallbacks
    VectoredReadOptions vrOptions = GoogleHadoopFileSystemConfiguration.getVectoredReadOptionBuilder(config).build();
    GoogleCloudStorageOptions gcsOptions = GoogleHadoopFileSystemConfiguration.getGcsOptionsBuilder(config).build();

    // 2. Map the fully resolved values
    mappedProperties.put(prefix + READ_THREAD_COUNT_KEY, String.valueOf(vrOptions.getReadThreads()));
    mappedProperties.put(prefix + "user-agent", gcsOptions.getAppName());
    if (gcsOptions.getRequesterPaysOptions().getProjectId() != null) {
      mappedProperties.put(prefix + USER_PROJECT_KEY, gcsOptions.getRequesterPaysOptions().getProjectId());
    }

    // 3. Clean up the original keys so they aren't accidentally passed
    mappedProperties.remove(GoogleHadoopFileSystemConfiguration.GCS_VECTORED_READ_THREADS.getKey());
    mappedProperties.remove(GoogleHadoopFileSystemConfiguration.GCS_REQUESTER_PAYS_PROJECT_ID.getKey());
    mappedProperties.remove(GoogleHadoopFileSystemConfiguration.GCS_APPLICATION_NAME_SUFFIX.getKey());

    // Proceed with the rest of the 1:1 HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS...
    // ...
```

dheerajsngh (changes requested):
Added a comment to update the implementation for the default values.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
shrutisinghania: As discussed offline, we will use the default value of analytics core.

dheerajsngh (approved):
Can we add a comment in the Mapper stating that we will use Analytics-core defaults when user doe not set the flag values.

PR #1708, opened 2026-05-06
[Backport to 3.1.x]: Feature tracking using header

shrutisinghania:
Backport PR: https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1619


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TrackedFeatures.java
```diff
@@ -0,0 +1,68 @@
...
+  PERFORMANCE_CACHE_ENABLED(6),
+
+  CLOUD_LOGGING_ENABLED(7),
+
+  TRACE_LOG_ENABLED(8),
+
+  OPERATION_TRACE_LOG_ENABLED(9),
+
+  DIRECT_UPLOAD_ENABLED(10),
+
```
animesh-g, line 56: Please add a comment here that 11 is reserved to avoid misuse in future. Also we need to make sure 11th bit is alway set as false.

animesh-g (approved):

PR #1710, opened 2026-05-12
Log requested byte range in high latency logs

shameersss1:
- Increase the logging threshold to 10 seconds
- Log bytesRanges

### DIFF
--- a/util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
+++ b/util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
@@ -30,14 +30,14 @@ public class GcsReadDurationTrackerStream extends FilterInputStream {
   private static final String UPLOAD_ID_HEADER = "x-guploader-uploadid";
 
   private final URI streamPath;
-  private final Object responseHeaders;
+  private final HttpHeaders responseHeaders;
   private final long latencyLoggingThresholdMs;
   private long cumulativeTransferDurationMs;
 
   public GcsReadDurationTrackerStream(
       InputStream delegate,
       URI streamPath,
-      Object responseHeaders,
+      HttpHeaders responseHeaders,
       long latencyLoggingThresholdMs) {
     super(delegate);
     this.streamPath = streamPath;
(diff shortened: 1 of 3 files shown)


util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -73,13 +72,15 @@ public void close() throws IOException {
     } finally {
       boolean latencyThresholdBreached = cumulativeTransferDurationMs > latencyLoggingThresholdMs;
       if (latencyThresholdBreached) {
-        String uploadId =
```
singhravidutt, line 76: don't we have tests for this?

Can you verify the log entry from integration test

singhravidutt (commented):
Approved with comments.

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: I did a manual testing and saw the following

```
26/05/13 08:34:43 INFO GcsReadDurationTrackerStream: Detected high latency for <file> durationMs=547; upload_id=<uploadId>; bytes 4-466861986/466861987
```

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -73,13 +72,15 @@ public void close() throws IOException {
...
-            "Detected high latency for %s. durationMs=%d; upload_id=%s",
-            streamPath, cumulativeTransferDurationMs, uploadId);
+        String uploadId = null;
+        String range = null;
+        if (responseHeaders != null) {
+          uploadId = responseHeaders.getFirstHeaderStringValue(UPLOAD_ID_HEADER);
+          range = responseHeaders.getContentRange();
+        }
+        logger.atInfo().log(
+            "Detected high latency for %s. durationMs=%d; upload_id=%s; %s",
```
animesh-g: nit: use `range=%s` similar to other parameters.

singhravidutt (approved):

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -73,13 +72,15 @@ public void close() throws IOException {
...
-        logger.atInfo().atMostEvery(10, TimeUnit.SECONDS).log(
-            "Detected high latency for %s. durationMs=%d; upload_id=%s",
-            streamPath, cumulativeTransferDurationMs, uploadId);
+        String uploadId = null;
+        String range = null;
+        if (responseHeaders != null) {
+          uploadId = responseHeaders.getFirstHeaderStringValue(UPLOAD_ID_HEADER);
+          range = responseHeaders.getContentRange();
+        }
+        logger.atInfo().log(
```
animesh-g: Q: Did we evaluate log flooding after removing 10 second rate limiter? I understand that default threshold is now 10 second but still it can be flooded by all open streams in Node and could affect disk I/O performance significantly.

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: ack

util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
shameersss1: Yes, the probability of GCS breaching 10seconds is very less, but for a bad day/region it could flood the log files as well and dump a lot of logs. I think bringing back rate limiting similar to RequestTracker class will be the best option to avoid unwanted logging.

PR #1720, opened 2026-05-18
Add FAST_FAIL_ON_NOT_FOUND  flag for tracking

shrutisinghania:
Add FAST_FAIL_ON_NOT_FOUND to connector feature tracking header under bit position 14.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TrackedFeatures.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TrackedFeatures.java
@@ -58,7 +58,9 @@ public enum TrackedFeatures {
 
   RENAME_API(12),
 
-  ANALYTICS_CORE_ENABLED(13);
+  ANALYTICS_CORE_ENABLED(13),
+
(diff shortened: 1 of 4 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TrackedFeatures.java
```diff
@@ -58,7 +58,9 @@ public enum TrackedFeatures {
 
   RENAME_API(12),
 
-  ANALYTICS_CORE_ENABLED(13);
+  ANALYTICS_CORE_ENABLED(13),
+
+  FAST_FAIL_ON_NOT_FOUND(14);
```
animesh-g: nit: can we call it FAST_FAIL_ON_NOT_FOUND_ENABLED similar to other metrics

animesh-g (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/TrackedFeatures.java
shrutisinghania: Done

PR #938, opened 2022-12-30
Clean up, improve and unify configuration and option classes handling of time and size properties

medb:
Changes in this PR:
- support size suffixes in values of size properties using [`Configuration#getLongBytes`](https://hadoop.apache.org/docs/r3.3.4/api/org/apache/hadoop/conf/Configuration.html#getLongBytes-java.lang.String-long-) method
- support time suffixes in values of time properties using [`Configuration#getTimeDuration`](https://hadoop.apache.org/docs/r3.3.4/api/org/apache/hadoop/conf/Configuration.html#getTimeDuration-java.lang.String-long-java.util.concurrent.TimeUnit-) method
- migrate time properties in option classes to `Duration` type
- adjust default values of the properties:
  ```
  fs.gs.outputstream.upload.chunk.size (64m -> 24m)
  fs.gs.http.connect-timeout (20s -> 5s)
  fs.gs.http.read-timeout (20s -> 5s)
  ```
- remove inline default value constants in option classes
- misc configuration/option-related code cleanup

PR #1731, opened 2026-06-02
Fix byte-size config properties to use getLongBytes for Hadoop 3.5.0 compatibility

### LINKED ISSUE #1730: fs.gs.block.size and vectored read size properties not using getLongBytes, causing NumberFormatException with Hadoop 3.5.0
Hadoop 3.5.0 added GCS connector defaults to `core-default.xml` using human-readable byte formats, e.g.:
```xml
<property><name>fs.gs.block.size</name><value>64m</value></property>
```
Calling `Configuration.getLong()` on `"64m"` throws `NumberFormatException` at startup. The fix is `getLongBytes()`, which handles these suffixes.

PR #938 already converted most byte-size properties to `getLongBytes` (`inplace.seek.limit`, `min.range.request.size`, `outputstream.buffer.size`, `rewrite.max.chunk.size`, etc.) but missed `fs.gs.block.size`. `fs.gs.vectored.read.min.range.seek.size` and `fs.gs.vectored.read.merged.range.max.size` were added after #938 and were not converted either.

Notably, `CONFIGURATION.md` already documents the defaults for these properties using human-readable suffixes (`64m`, `4k`, `8m`), meaning users following the documentation and setting values like `fs.gs.block.size=128m` would have hit a `NumberFormatException`. The fix aligns the implementation with what the documentation already implies.

Note: `fs.gs.inputstream.latency.logging.threshold.ms` also uses `getLong` but its key name carries an explicit `.ms` suffix — fixing it cleanly would require a key rename and is left as a follow-up.

laurentgo:
Fixes #1730

Hadoop 3.5.0 defines `fs.gs.block.size = "64m"` in `core-default.xml`. Calling `getLong()` on that value throws `NumberFormatException` at startup. Switch `BLOCK_SIZE`, `GCS_VECTORED_READ_RANGE_MIN_SEEK`, and `GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE` to `getLongBytes()`, which handles human-readable byte suffixes (`k`, `m`, `g`).

The two vectored-read constants are widened from `Integer` to `Long`; `toIntExact()` is applied at the setter call site since `VectoredReadOptions` still accepts `int`.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
@@ -391,16 +391,16 @@ public class GoogleHadoopFileSystemConfiguration {
               GoogleCloudStorageReadOptions.DEFAULT.getLatencyLoggingThreshold());
 
   /** Minimum distance that will be seeked without merging the ranges together. */
-  public static final HadoopConfigurationProperty<Integer> GCS_VECTORED_READ_RANGE_MIN_SEEK =
+  public static final HadoopConfigurationProperty<Long> GCS_VECTORED_READ_RANGE_MIN_SEEK =
       new HadoopConfigurationProperty<>(
           "fs.gs.vectored.read.min.range.seek.size",
-          VectoredReadOptions.DEFAULT.getMinSeekVectoredReadSize());
+          (long) VectoredReadOptions.DEFAULT.getMinSeekVectoredReadSize());
 
   /** Maximum size allowed for a merged range request. */
-  public static final HadoopConfigurationProperty<Integer> GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE =
+  public static final HadoopConfigurationProperty<Long> GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE =
       new HadoopConfigurationProperty<>(
           "fs.gs.vectored.read.merged.range.max.size",
-          VectoredReadOptions.DEFAULT.getMergeRangeMaxSize());
+          (long) VectoredReadOptions.DEFAULT.getMergeRangeMaxSize());
 
   /** Maximum threads to process individual FileRange requests */
   public static final HadoopConfigurationProperty<Integer> GCS_VECTORED_READ_THREADS =
@@ -668,8 +668,10 @@ public class GoogleHadoopFileSystemConfiguration {
     }
     logger.atInfo().log("Using %d threads for vectored reads", readThreads);
     return VectoredReadOptions.builder()
-        .setMinSeekVectoredReadSize(GCS_VECTORED_READ_RANGE_MIN_SEEK.get(config, config::getInt))
-        .setMergeRangeMaxSize(GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE.get(config, config::getInt))
+        .setMinSeekVectoredReadSize(
+            toIntExact(GCS_VECTORED_READ_RANGE_MIN_SEEK.get(config, config::getLongBytes)))
+        .setMergeRangeMaxSize(
+            toIntExact(GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE.get(config, config::getLongBytes)))
         .setReadThreads(readThreads);
   }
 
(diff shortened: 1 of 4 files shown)


laurentgo:
hey @shameersss1, since this issue is causing a bit of friction with latest hadoop versions, I was wondering if you had any feedback on the proposed fix.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -391,16 +391,16 @@ public class GoogleHadoopFileSystemConfiguration {
               GoogleCloudStorageReadOptions.DEFAULT.getLatencyLoggingThreshold());
 
   /** Minimum distance that will be seeked without merging the ranges together. */
-  public static final HadoopConfigurationProperty<Integer> GCS_VECTORED_READ_RANGE_MIN_SEEK =
+  public static final HadoopConfigurationProperty<Long> GCS_VECTORED_READ_RANGE_MIN_SEEK =
       new HadoopConfigurationProperty<>(
           "fs.gs.vectored.read.min.range.seek.size",
-          VectoredReadOptions.DEFAULT.getMinSeekVectoredReadSize());
+          (long) VectoredReadOptions.DEFAULT.getMinSeekVectoredReadSize());
```
shameersss1, line 397: Why are we casting to long here ? Can we make `VectoredReadOptions.DEFAULT.getMinSeekVectoredReadSize()` return long similar to 

```
          "fs.gs.inputstream.min.range.request.size",
          GoogleCloudStorageReadOptions.DEFAULT.getMinRangeRequestSize());
```
for consistency ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -668,8 +668,10 @@ static VectoredReadOptions.Builder getVectoredReadOptionBuilder(Configuration co
     }
     logger.atInfo().log("Using %d threads for vectored reads", readThreads);
     return VectoredReadOptions.builder()
-        .setMinSeekVectoredReadSize(GCS_VECTORED_READ_RANGE_MIN_SEEK.get(config, config::getInt))
-        .setMergeRangeMaxSize(GCS_VECTORED_READ_MERGED_RANGE_MAX_SIZE.get(config, config::getInt))
+        .setMinSeekVectoredReadSize(
+            toIntExact(GCS_VECTORED_READ_RANGE_MIN_SEEK.get(config, config::getLongBytes)))
+        .setMergeRangeMaxSize(
```
shameersss1, line 673: Can we avoid toInt here and use long through out ?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
laurentgo: I used the same pattern as #938 (see [declaration](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/03ae089c991a24776f15767ab387c0a448224ac9/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L287) and [usage](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/03ae089c991a24776f15767ab387c0a448224ac9/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L781)). Unfortunately Hadoop Vectored APIs in [`VectoredReadUtils`](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/VectoredReadUtils.java) take an `int` and not a `long`, so we would not be able to do it end to end

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
laurentgo: Same answer as the previous comment, the value is used with Hadoop Vectored API which only takes an `int`, not a `long`

shameersss1 (approved):
LGTM +1

shameersss1:
@laurentgo There are test failures in VectoredIOImplTest - Make sure you run all the UTs

laurentgo:
> @laurentgo There are test failures in VectoredIOImplTest - Make sure you run all the UTs

Tests pass locally for me, and unfortunately I cannot see the cloud build report myself. Could you share details here?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -475,6 +475,18 @@ public void close_closesAndNullifiesInstrumentation() throws Exception {
     assertThat(instrumentationField.get(ghfs)).isNull();
   }
 
+  @Test
+  public void blockSize_sizeSuffix() throws Exception {
+    Configuration config = new Configuration();
+    config.set("fs.gs.block.size", "128m");
```
shameersss1: intead of hardcode `fs.gs.block.size` can you use the variable name ?

laurentgo: It seems this file uses a mix of string literals and constants, so was not sure what style to adopt but changed per your request.

laurentgo:
@shameersss1 is there anything left you would like me to address?

shameersss1:
Triggered a test run, post the successful run, i will merge the changes

shameersss1:
@laurentgo Thanks for your contribution

laurentgo:
@shameersss1 would you be able to cherrypicks for 3.1 and 4.0? considering this is a friction point with hadoop 3.5.0 today, it would be nice if there would be a release soon which would alleviate the issue.

shameersss1:
@laurentgo - Please raise a PR against branch-3.1.x and branch-4.0.x as well

PR #1741, opened 2026-06-09
Cleanly close the read channel in the GoogleHadoopFSInputStream so that read to EOF does not happen on calling close() unnecessarily

### LINKED ISSUE #1729: GoogleHadoopFSInputStream does not close when calling close(), it reads the file fully instead.
# Summary

An opened `GoogleHadoopFSInputStream` when read() for the first time (after seek at some position, or from 0 position without), [opens a stream from the seeked position towards the end of file](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/03ae089c991a24776f15767ab387c0a448224ac9/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L912).

This is not an issue for the most part, because you may want to read sequentially for some time. But later when you try to close() the input stream, it gets stuck, we saw the following log message in the stderr, close() operation being stuck for more than 8minutes, and sometimes even more (it was a 64gb file with seek somewhere in the middle):

```
May 25, 2026 1:34:13 PM com.google.cloud.hadoop.fs.gcs.GhfsStorageStatistics updateMinMaxStats
INFO: Detected potential high latency for operation stream_read_close_operations. latencyMs=514390; previousMaxLatencyMs=0; operationCount=1; context=gs://path/to/0003.csv 
```

This led to taking a thread dump and analyze what's going on, what I found out was pretty astonishing, it was stuck at Apache HTTP Client's [ContentLengthInputStream.close()](https://github.com/apache/httpcomponents-core/blob/386189e645b1e40e3b8a9a2c668acc7852006e1e/httpcore5/src/main/java/org/apache/hc/core5/http/impl/io/ContentLengthInputStream.java#L96), where it drains the whole stream if it has not:

```
    /**
     * <p>Reads until the end of the known length of content.</p>
     *
     * <p>Does NOT close the underlying stream, but instead leaves it
     * primed to parse the next response.</p>
     * @throws IOException If an IO problem occurs.
     */
    @Override
    public void close() throws IOException {
        if (!closed) {
            try {
                if (pos < contentLength) {
                    final byte[] buffer = new byte[BUFFER_SIZE];
                    while (read(buffer) >= 0) {
                        // keep reading
                    }
                }
            } finally {
                // close after above so that we don't throw an exception trying
                // to read after closed!
                closed = true;
            }
        }
    }
```

This is a pretty weird behavior, because any other language's http library they drop the connection if called close,

- Apache HTTP: Drains everything, no limit
- Python urllib3: Drains up to a threshold, then drops connection
- Rust hyper/reqwest: Drops connection immediately, no drain
- Ruby Net::HTTP: Drops connection, no drain

## Possible workaround we found so far

One workaround for bypassing this weird behavior of Apache's HTTP Client was to open the stream only for what's required using the following options:

```
fs.gs.inputstream.fadvise (String): RANDOM
fs.gs.inputstream.min.range.request.size (int): <stream-size>
```

But since in the connector 2.x `fs.gs.inputstream.min.range.request.size` expected an int, we can't open the stream for more than 2GB-1bytes, since its the max integer value permits.

https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/03ae089c991a24776f15767ab387c0a448224ac9/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java#L953

bytesToRead is `long` while it generally is value something like 8192 bytes, that is loaded into bytebuffer while calling read() multiple times, we want to set `readOptions.getMinRangeRequestSize()` which is `int` to 2GB (2147483648 bytes). This is really weird data type limitation introduced over here.

This is [currently fixed in connector 3.x by changing `fs.gs.inputstream.min.range.request.size` datatype from int to long](https://github.com/GoogleCloudDataproc/hadoop-connectors/commit/35b76e8bdc753802e4cc4727763ef088dcd9c476). We'll update to connector 3.x in future to take advantage of this.

## What could be better?

Though the close() behavior in the inputstream should close the stream, not read the whole file instead.

Animeshz:
Fixes #1729.

Calling close() on `GoogleHadoopFSInputStream` caused the whole stream to get drained, with AUTO/SEQUENTIAL fadvise (default) the stream is opened to the EOF, and hence whole file has to be read once opened even when calling close().

This PR cleanly closes the connection before Apache HTTP library tries to drain the bytes, so that only necessary bytes are read rest is dropped by closure of underlying HTTP response.

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
@@ -556,6 +556,10 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
    * Aborts the underlying HTTP request for the current media body, if any. Must run before closing
    * the {@link #contentChannel} so that closing the Apache {@code ContentLengthInputStream} does
    * not synchronously read the rest of the response body.
+   *
+   * <p>Failures from {@link HttpResponse#disconnect()} are expected during normal teardown (for
+   * example when the socket is already closed) and are only logged here; they are not reported via
+   * {@link GoogleCloudStorageEventBus#postOnException()}.
    */
   void disconnectHttpResponse(@Nullable HttpResponse response) {
     if (response == null) {
@@ -564,7 +568,6 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
     try {
       response.disconnect();
     } catch (Exception e) {
-      GoogleCloudStorageEventBus.postOnException();
       logger.atFine().withCause(e).log(
           "Got an exception on HttpResponse.disconnect() for '%s'; ignoring it.", resourceId);
     }
@@ -1168,7 +1171,11 @@ public class GoogleCloudStorageReadChannel implements SeekableByteChannel {
       return tracked;
     } catch (IOException e) {
       GoogleCloudStorageEventBus.postOnException();
-      disconnectHttpResponse(response);
+      try {
+        response.disconnect();
+      } catch (IOException closeException) {
+        e.addSuppressed(closeException);
+      }
       throw e;
     }
   }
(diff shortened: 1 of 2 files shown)


Animeshz:
Also need https://github.com/googleapis/google-http-java-client/issues/1303 fixed before this is merged. The disconnect is broken at google http library level, it forces drain before disconnect.

dheerajsngh:
Refer to [comment](https://github.com/GoogleCloudDataproc/hadoop-connectors/issues/1729#issuecomment-4748601097). 
The above fix is for master but the issue is only in branch-2.2.x.

PR #1735, opened 2026-06-04
add capability check for lexicographical listing

singhravidutt:

PR #1755, opened 2026-07-07
add capability check for lexicographical listing (#1735)

singhravidutt:


shameersss1 (approved):

animesh-g:
@singhravidutt PR looks good! Just a quick heads-up: it's better to target the main branch first and then backport it to the release branches. We've had issues in the past where things didn't get synced back to main, so keeping that as the source of truth really helps avoid any inconsistencies.

singhravidutt:
> @singhravidutt PR looks good! Just a quick heads-up: it's better to target the main branch first and then backport it to the release branches. We've had issues in the past where things didn't get synced back to main, so keeping that as the source of truth really helps avoid any inconsistencies.

Thanks Animesh, it is indeed a backport of [PR-1735](https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1735)

PR #1758, opened 2026-07-13
Integrate and map Analytics Core fadvise

shrutisinghania:
This PR integrates and maps the Analytics Core `fadvise` configurations while utilizing its default values as mentioned [here](https://github.com/GoogleCloudPlatform/gcs-analytics-core/blob/main/CONFIGURATION.md).

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
@@ -79,16 +79,7 @@ final class AnalyticsCoreConfigMapper {
   static Map<String, String> mapConfigs(Configuration config, String prefix) {
     Map<String, String> mappedProperties = config.getValByRegex("^" + prefix.replace(".", "\\."));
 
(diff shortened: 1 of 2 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -81,4 +107,19 @@ private static void mapAndRemoveSource(
       map.put(analyticsCoreKey, value);
     }
   }
+
+  private static String toFileAccessPattern(String fadvise) {
```
animesh-g, line 105: nit: Why didn't we use HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS here for consistency?

animesh-g (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
shrutisinghania: Done.

PR #1762, opened 2026-07-21
Fix AUTO fadvise one-way latch causing excessive HTTP requests for Parquet reads

RRap0so:
AUTO fadvise mode permanently switched to random-access (bounded 2MB range
requests) after detecting a backward seek, and never recovered. Parquet reads
always trigger this because the format requires reading the footer at the end
of the file before reading row groups sequentially from the beginning.

For a 1GB file this meant ~500 HTTP requests instead of ~2, causing Dataflow
worker timeouts on batch Parquet read+join stages after Beam 2.74 bumped
gcs-connector from v2.2.26 (default SEQUENTIAL) to v3.1.16 (default AUTO).

Changes:
- Let AUTO mode recover to sequential after fadviseRequestTrackCount (default 3)
  consecutive sequential reads
- Reset consecutiveSequentialCount and seek flags on mode transitions to prevent
  oscillation
- Validate fadviseRequestTrackCount > 0 at build time

AUTO_RANDOM behavior is unchanged — it still uses sticky seek flags as a
permanent latch, which is preserved because unsetRandomAccess() is only
reachable from AUTO_RANDOM when both flags are already false.

Tested against a real 36MB Parquet file on GCS:
- Unpatched AUTO: 11 data requests (all bounded), random=true
- Patched AUTO: 5 data requests (3 bounded, then streaming), random=false
- SEQUENTIAL: 2 data requests (unchanged)

### DIFF
--- a/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/FileAccessPatternManager.java
+++ b/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/FileAccessPatternManager.java
@@ -65,7 +65,8 @@ class FileAccessPatternManager {
       return;
     }
     updateSeekFlags(currentPosition);
-    if (readOptions.getFadvise() == Fadvise.AUTO_RANDOM) {
+    if (readOptions.getFadvise() == Fadvise.AUTO_RANDOM
+        || readOptions.getFadvise() == Fadvise.AUTO) {
       if (randomAccess) {
         if (shouldAdaptToSequential(currentPosition)) {
           unsetRandomAccess();
@@ -74,10 +75,6 @@ class FileAccessPatternManager {
         if (shouldAdaptToRandomAccess(currentPosition)) {
           setRandomAccess();
         }
-      }
-    } else if (readOptions.getFadvise() == Fadvise.AUTO) {
-      if (shouldAdaptToRandomAccess(currentPosition)) {
-        setRandomAccess();
       }
     }
   }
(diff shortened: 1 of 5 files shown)


shunping:
@singhravidutt: could you help review this PR?

singhravidutt (commented):
This could be risky for other customers. Reading a parquet file in sequential mode is too risky as it will open channel till EOF. This will eat up on you egress quota even when the data transferred is just for a row group.

I understand default got updated for gcsio but for ghfs it's was always the AUTO mode being default mode.

If suggest creating a new flavour or overriding the default  to use SEQUENTIAL.

singhravidutt:
@animesh-g Please take a look.

RRap0so:
> This could be risky for other customers. Reading a parquet file in sequential mode is too risky as it will open channel till EOF. This will eat up on you egress quota even when the data transferred is just for a row group.
>
> I understand default got updated for gcsio but for ghfs it's was always the AUTO mode being default mode.
>
> If suggest creating a new flavour or overriding the default to use SEQUENTIAL.

Opening till EOF doesn't mean egress till EOF. The client won't let the server run that far ahead.

java-storage 2.61.0, `GapicUnbufferedReadableByteChannel`:

- `:328` `disableAutoInboundFlowControl()`
- `:329` `request(1)` on start, `:334` one `request(1)` per response
- `:121` 2-slot queue
- `:201` `close()` → `:322` `controller.cancel()`

The client asks for one message at a time, so the server can't get ahead of it. Close cancels the stream. Walking away from a range costs one message, about 2 MiB, no matter how big the range was.

Seek-heavy access is where it does cost you. Parquet-shaped pattern over `GoogleCloudStorageClientReadChannel`: 12 MiB of useful reads put 22-26 MiB on the wire under SEQUENTIAL, against 14 MiB under RANDOM. Roughly 2x, from per-abort overshoot and in-place skip reads. That's why the flip requires `fadviseRequestTrackCount` consecutive sequential reads.

One thing I'm unsure about: this leans on java-storage internals. Do you know whether it's safe to count on? If not I'd sooner not build on it, and capping the range size would be the safer shape. Either way, happy to put the recovery behind a feature flag if you'd rather the ghfs default didn't move.

animesh-g:
Hi @RRap0so, thank you for your contribution and apologies for the delay in getting back to you.

We previously evaluated a similar approach but unfortunately observed mild regressions without significant overall improvement. Additionally, both our team and Apache Beam are moving away from the `gcsio` layer. [Apache Beam is transitioning](https://s.apache.org/beam-gcsutil-modernization) to the Java SDK, and we are planning to replace gcsio with the optimized [gcs-analytics-core](https://github.com/GoogleCloudPlatform/gcs-analytics-core) layer.

Because of these upcoming architectural changes and the potential for regressions, we are leaning towards holding off on this PR for now. Let us know if you have a different perspective on this. Thank you again for the effort you invested here.

PR #1763, opened 2026-07-22
Add caffeine dependency to shaded artifact set in pom.xml

shrutisinghania:
Add caffeine dependency to shaded artifact set in pom.xml. This is needed for the analytics-core version upgrade to `1.4.0`

### DIFF
--- a/gcs/pom.xml
+++ b/gcs/pom.xml
@@ -238,6 +238,7 @@
                   <include>io.opentelemetry.semconv</include>
                   <include>io.perfmark</include>
                   <include>org.apache.httpcomponents</include>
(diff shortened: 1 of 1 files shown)


animesh-g (commented):
ques: Could you please clarify why the analytics-core version isn't being updated in this PR? Including that update would help clarify that the new caffeine dependency is required for that specific version.

shrutisinghania:
> ques: Could you please clarify why the analytics-core version isn't being updated in this PR? Including that update would help clarify that the new caffeine dependency is required for that specific version.

The analytics-core version was already upgraded in PR https://github.com/GoogleCloudDataproc/hadoop-connectors/pull/1759

animesh-g (approved):

PR #1764, opened 2026-07-27
Implement BiDi bulk open

shrutisinghania:
DD: [go/connector-bidi-bulk-open](http://goto.google.com/bidi-bulk-open)

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -95,6 +95,7 @@ import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -671,6 +672,31 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
           return new FSDataInputStream(
               GoogleHadoopFSInputStream.create(this, fileStatus.getFileInfo(), statistics));
         });
+  }
+
+  /**
+   * Prewarms channels for reading multiple Hadoop paths. This is GCS-specific extension API.
+   *
+   * <p>This is an experimental API and can change without notice.
+   *
+   * @param pathSizeMap Map of Hadoop paths to their expected sizes.
+   * @throws IOException on IO error
+   */
+  public void multiOpen(Map<Path, Long> pathSizeMap) throws IOException {
+    checkOpen();
+    checkArgument(pathSizeMap != null, "pathSizeMap must not be null");
+
+    logger.atFine().log("multiOpen(pathSizeMap=%s)", pathSizeMap);
+
+    Map<StorageResourceId, Long> resourcesAndSizes = new LinkedHashMap<>();
+    for (Map.Entry<Path, Long> entry : pathSizeMap.entrySet()) {
+      long size = (entry.getValue() == null || entry.getValue() < 0) ? -1L : entry.getValue();
+      URI gcsPath = getGcsPath(entry.getKey());
+      resourcesAndSizes.put(
+          StorageResourceId.fromUriPath(gcsPath, /* allowEmptyObjectName= */ false), size);
+    }
+
+    getGcsFs().getGcs().multiOpen(resourcesAndSizes);
   }
 
   @Override
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -690,6 +690,7 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
 
     Map<StorageResourceId, Long> resourcesAndSizes = new LinkedHashMap<>();
     for (Map.Entry<Path, Long> entry : pathSizeMap.entrySet()) {
+      checkArgument(entry.getKey() != null, "pathSizeMap keys must not be null");
       long size = (entry.getValue() == null || entry.getValue() < 0) ? -1L : entry.getValue();
       URI gcsPath = getGcsPath(entry.getKey());
       resourcesAndSizes.put(
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -77,6 +77,7 @@ import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.flogger.GoogleLogger;
 import com.google.common.io.BaseEncoding;
@@ -95,7 +96,6 @@ import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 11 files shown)


gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -1334,7 +1369,131 @@ private SeekableByteChannel open(
...
+      GoogleCloudStorageBidiReadChannel channel, GoogleCloudStorageReadOptions readOptions) {
+    long idleTimeNs = System.nanoTime() - channel.getLastAccessTimeNs();
+    if (idleTimeNs > TimeUnit.SECONDS.toNanos(readOptions.getBidiCacheExpireSec())) {
+      channel.actualClose();
+      return true;
+    }
+    return false;
+  }
+
+  private synchronized ExecutorService getBoundedThreadPool(int bidiThreadCount) {
```
animesh-g: Not sure why we have synchronize only this method. Other methods are also susceptible  to concurrency issues e.g. closed and returnChannel. In my opinion, either we should make this class thread safe or let the upper layer responsible for synchronization. WDYT?

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
```diff
@@ -1705,6 +1868,34 @@ private static GoogleCloudStorageItemInfo createItemInfoForBlob(
         new VerificationAttributes(md5Hash, crc32c));
   }
 
+  @Override
+  public void returnChannel(
+      StorageResourceId resourceId, GoogleCloudStorageBidiReadChannel channel) {
+    StorageResourceId normalizedKey = normalizeKey(resourceId);
+    if (!closed && channel.isOpen()) {
+      ConcurrentLinkedQueue<GoogleCloudStorageBidiReadChannel> queue =
+          channelPool.asMap().computeIfAbsent(normalizedKey, k -> new ConcurrentLinkedQueue<>());
```
animesh-g: This can be simplified as following which also thread-safe,

channelPool.get(normalizedKey, () -> new ConcurrentLinkedQueue<>())

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -207,81 +262,144 @@ public boolean isOpen() {
   }
 
   @Override
-  public void close() throws IOException {
+  public synchronized void close() throws IOException {
```
animesh-g: Not sure if we really need synchronized here when closeTriggered, pendingReads and returnTriggered are already atomic.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java
```diff
@@ -412,6 +412,14 @@ public class GoogleHadoopFileSystemConfiguration {
       new HadoopConfigurationProperty<>(
           "fs.gs.grpc.enable", GoogleCloudStorageOptions.DEFAULT.isGrpcEnabled());
 
+  /** Configuration key for idle timeout (TTL) for cached connections. */
+  public static final HadoopConfigurationProperty<Integer> GCS_BIDI_CACHE_EXPIRE_SEC =
+      new HadoopConfigurationProperty<>("fs.gs.bidi.cache.expire.sec", 600);
+
+  /** Configuration key for maximum unique files cached in the pool. */
```
animesh-g: nit: let's call it object.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -673,6 +674,32 @@ public FSDataInputStream open(FileStatus status) throws IOException {
...
+   * @param pathSizeMap Map of Hadoop paths to their expected sizes.
+   * @throws IOException on IO error
+   */
+  public void multiOpen(Map<Path, Long> pathSizeMap) throws IOException {
+    checkOpen();
+    checkArgument(pathSizeMap != null, "pathSizeMap must not be null");
+
+    logger.atFine().log("multiOpen(pathSizeMap=%s)", pathSizeMap);
+
+    Map<StorageResourceId, Long> resourcesAndSizes = new LinkedHashMap<>();
```
animesh-g: Why don't we intilizae with size here? i.e. `new LinkedHashMap<>(pathSizeMap.size())` to avoid rehashing and resizing operations at runtime.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -673,6 +674,32 @@ public FSDataInputStream open(FileStatus status) throws IOException {
         });
   }
 
+  /**
+   * Prewarms channels for reading multiple Hadoop paths. This is GCS-specific extension API.
```
animesh-g: I think we should change the language here. The complete list isn't guaranteed to be prewarmed. It depends on cache size as well as individual `prewarmChannel` failures(e.g. IOException) do not result in API failures.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -70,23 +72,51 @@ public final class GoogleCloudStorageBidiReadChannel implements ReadVectoredSeek
...
+    this(storage, resourceId, itemInfo, readOptions, boundedThreadPool, false, null);
+  }
+
+  public GoogleCloudStorageBidiReadChannel(
+      Storage storage,
+      StorageResourceId resourceId,
+      @Nullable GoogleCloudStorageItemInfo itemInfo,
+      GoogleCloudStorageReadOptions readOptions,
+      ExecutorService boundedThreadPool,
+      boolean isMockMetadata,
```
animesh-g: nit: `isMockMetadata` implies like test-only code but it isn't. You can use better name like `isSpeculativeMetadata`, `isEstimatedMetadata` or something similar.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageBidiReadChannel.java
```diff
@@ -207,81 +262,144 @@ public boolean isOpen() {
...
+      if (readSession != null) {
+        readSession.close();
+      }
+    } catch (Exception e) {
+      logger.atFine().withCause(e).log(
+          "Failed to get/close completed BlobReadSession during close() for '%s'", resourceId);
+    }
+  }
+
+  private synchronized void decrementPendingReads(int count) {
```
animesh-g: Similar to other places using synchronized and Atomic both is redundant and will cause unnecessary contention.

animesh-g (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shrutisinghania: Guava's `Cache.get(K, Callable)` throws a checked ExecutionException which would require a try-catch block here (since the lambda can't throw it). IMO, using `channelPool.asMap().computeIfAbsent` is thread-safe and avoids checked exceptions, keeping the implementation cleaner.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shrutisinghania: Only `getBoundedThreadPool()` is synchronized because it is called when a new channel is created, keeping the overhead negligible unlike `returnChannel()` which is called every time a read channel is closed.

This class is thread safe, all variables are either volatile, thread-safe, or initialized inside synchronized blocks.

animesh-g: `volatile` doesn't guarantee atomicity and implementation is still susceptible to concurrency issues.. Look into https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

 In returnChannel, the code evaluates if (!closed && channel.isOpen()). If another thread calls close() immediately after this check, this.closed = true; is executed and the pool is invalidated. However, the first thread will proceed to execute queue.offer(channel);, effectively leaking an open channel into a "closed" client.

Also most of the gcsio implementation is not thread safe and is expected. We have to make sure things are thread-safe at hadoop(gcs) layer and we don't do unnecessary synchronizations at gcsio.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
animesh-g: This was a styling related comment is optional but I feel `get` is more idiomatic and a 3 line try-catch is good trade off here. I will leave it up to you.

animesh-g (approved):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java
shrutisinghania: Done.

PR #1767, opened 2026-08-05
feat: Integrate Analytics Core write path into GoogleHadoopOutputStream

dheerajsngh:
## Description
This PR integrates the GCS Analytics Core write path into the connector when `fs.gs.analytics.core.write.enable` is active (part 2 of the Analytics Core write migration stack).

### Key Changes
* **Stream Routing:** Delegates write stream creation to Analytics Core with legacy fallback.
* **Stream Wrapper:** Added `GcsAnalyticsCoreOutputStreamWrapper` for synchronized delegation around `GoogleCloudStorageOutputStream`.
* **Overwrite & Generation:** Enforces `FileAlreadyExistsException` on `overwrite=false` and resolves generation IDs for overwrites.
* **Buffering:** Wraps streams in `BufferedOutputStream` matching configured buffer size.

### Testing
* Added `GcsAnalyticsCoreOutputStreamWrapperTest` covering stream delegation, close, and synchronization.
* Added unit tests in `GoogleHadoopOutputStreamTest` covering write delegation, overwrite checks, and generation resolution.
* Verified all unit tests and spotless checks pass cleanly.

### Style Guide Compliance Checklist
- [x] Conventional Commit title (`feat: ...`)
- [x] Apache 2.0 license headers verified on all new files

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.hadoop.fs.gcs;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageOutputStream;
 import com.google.common.flogger.GoogleLogger;
 import java.io.IOException;
@@ -33,8 +35,7 @@ class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
   private final GoogleCloudStorageOutputStream delegate;
 
   public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
-    this.delegate = delegate;
-    logger.atInfo().log("GcsAnalyticsCoreOutputStreamWrapper created");
+    this.delegate = checkNotNull(delegate, "delegate cannot be null");
   }
 
   @Override
@@ -51,10 +52,8 @@ class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
 
   @Override
   public synchronized void close() throws IOException {
-    logger.atInfo().log("Closing GcsAnalyticsCoreOutputStreamWrapper");
     try {
       delegate.close();
-      logger.atInfo().log("Closed GcsAnalyticsCoreOutputStreamWrapper successfully");
     } catch (IOException e) {
       logger.atWarning().withCause(e).log("Failed to close GcsAnalyticsCoreOutputStreamWrapper");
       throw e;
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.hadoop.fs.gcs;
+
+import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageOutputStream;
+import com.google.common.flogger.GoogleLogger;
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * A synchronized wrapper around the non-thread-safe GoogleCloudStorageOutputStream.
+ *
+ * <p>TODO(user): Implement GoogleCloudStorageItemInfo.Provider once gcs-analytics-core exposes the
+ * finalized generation ID upon close.
+ */
+class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+
+  private final GoogleCloudStorageOutputStream delegate;
+
+  public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
+    this.delegate = delegate;
+    logger.atInfo().log("GcsAnalyticsCoreOutputStreamWrapper created");
+  }
+
+  @Override
+  public synchronized void write(int b) throws IOException {
+    logger.atFiner().log("write(int)");
+    delegate.write(b);
+  }
+
+  @Override
+  public synchronized void write(byte[] b, int off, int len) throws IOException {
+    logger.atFine().log("write(byte[], off=%d, len=%d)", off, len);
+    delegate.write(b, off, len);
+  }
+
+  @Override
+  public synchronized void close() throws IOException {
+    logger.atInfo().log("Closing GcsAnalyticsCoreOutputStreamWrapper");
+    try {
+      delegate.close();
+      logger.atInfo().log("Closed GcsAnalyticsCoreOutputStreamWrapper successfully");
+    } catch (IOException e) {
+      logger.atWarning().withCause(e).log("Failed to close GcsAnalyticsCoreOutputStreamWrapper");
+      throw e;
+    }
+  }
+}
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
@@ -33,6 +33,7 @@ class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
   private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
 
   private final GoogleCloudStorageOutputStream delegate;
+  private boolean closed = false;
 
   public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
     this.delegate = checkNotNull(delegate, "delegate cannot be null");
@@ -52,11 +53,14 @@ class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
 
   @Override
   public synchronized void close() throws IOException {
+    if (closed) {
+      logger.atFiner().log("close(): Stream already closed, ignoring.");
+      return;
+    }
     try {
       delegate.close();
-    } catch (IOException e) {
-      logger.atWarning().withCause(e).log("Failed to close GcsAnalyticsCoreOutputStreamWrapper");
-      throw e;
+    } finally {
+      closed = true;
     }
   }
 }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
@@ -41,7 +41,6 @@ class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
 
   @Override
   public synchronized void write(int b) throws IOException {
-    logger.atFiner().log("write(int)");
     delegate.write(b);
   }
 
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -414,7 +414,7 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
       }
       // TODO(user): Initialize analyticsCoreGcsFs lazily when GCS_LAZY_INITIALIZATION_ENABLE is
       // true, to avoid eager initialization.
-      if (isAnalyticsCoreEnabled() || isAnalyticsWriteEnabled()) {
+      if (isAnalyticsCoreEnabled() || isAnalyticsCoreWriteEnabled()) {
         analyticsCoreGcsFs = createAnalyticsGcsFs(config);
       }
     }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -44,7 +44,10 @@ import com.google.cloud.gcs.analyticscore.client.GcsFileInfo;
 import com.google.cloud.gcs.analyticscore.client.GcsFileSystem;
 import com.google.cloud.gcs.analyticscore.client.GcsFileSystemImpl;
 import com.google.cloud.gcs.analyticscore.client.GcsFileSystemOptions;
+import com.google.cloud.gcs.analyticscore.client.GcsItemId;
+import com.google.cloud.gcs.analyticscore.client.GcsWriteOptions;
 import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageInputStream;
+import com.google.cloud.gcs.analyticscore.core.GoogleCloudStorageOutputStream;
 import com.google.cloud.hadoop.fs.gcs.auth.GcsDelegationTokens;
 import com.google.cloud.hadoop.gcsio.CreateFileOptions;
 import com.google.cloud.hadoop.gcsio.FeatureHeaderGenerator;
@@ -409,7 +412,7 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
       } else {
         initializeGcsFs(createGcsFs(config));
       }
-      if (isAnalyticsCoreEnabled()) {
+      if (isAnalyticsCoreEnabled() || isAnalyticsWriteEnabled()) {
         analyticsCoreGcsFs = createAnalyticsGcsFs(config);
       }
     }
@@ -1802,6 +1805,11 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
   GoogleCloudStorageInputStream createAnalyticsCoreInputStream(GcsFileInfo gcsFileInfo)
       throws IOException {
     return GoogleCloudStorageInputStream.create(analyticsCoreGcsFs, gcsFileInfo);
+  }
+
+  GoogleCloudStorageOutputStream createAnalyticsCoreOutputStream(
+      GcsItemId gcsItemId, GcsWriteOptions writeOptions) throws IOException {
+    return GoogleCloudStorageOutputStream.create(analyticsCoreGcsFs, gcsItemId, writeOptions);
   }
 
   /** Checks if Analytics Core is enabled. */
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -412,6 +412,8 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
       } else {
         initializeGcsFs(createGcsFs(config));
       }
+      // TODO(user): Initialize analyticsCoreGcsFs lazily (e.g. using a Supplier) when
+      // GCS_LAZY_INITIALIZATION_ENABLE is true, to avoid eager credential loading.
       if (isAnalyticsCoreEnabled() || isAnalyticsWriteEnabled()) {
         analyticsCoreGcsFs = createAnalyticsGcsFs(config);
       }
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
@@ -412,8 +412,8 @@ public class GoogleHadoopFileSystem extends FileSystem implements IOStatisticsSo
       } else {
         initializeGcsFs(createGcsFs(config));
       }
-      // TODO(user): Initialize analyticsCoreGcsFs lazily (e.g. using a Supplier) when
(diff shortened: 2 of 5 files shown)


gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapperTest.java
```diff
@@ -0,0 +1,76 @@
...
+    verify(mockOutputStream).close();
+  }
+
+  @Test
+  public void methodsAreSynchronized() throws NoSuchMethodException {
+    verifyMethodIsSynchronized(GcsAnalyticsCoreOutputStreamWrapper.class, "write", int.class);
+    verifyMethodIsSynchronized(
+        GcsAnalyticsCoreOutputStreamWrapper.class, "write", byte[].class, int.class, int.class);
+    verifyMethodIsSynchronized(GcsAnalyticsCoreOutputStreamWrapper.class, "close");
+  }
```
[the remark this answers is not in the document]
dheerajsngh: As commented above, we will not be adding the flush method. Hence ignoring this.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -409,7 +412,7 @@ private synchronized void initializeGcsFs(Configuration config) throws IOExcepti
       } else {
         initializeGcsFs(createGcsFs(config));
       }
-      if (isAnalyticsCoreEnabled()) {
+      if (isAnalyticsCoreEnabled() || isAnalyticsWriteEnabled()) {
         analyticsCoreGcsFs = createAnalyticsGcsFs(config);
       }
```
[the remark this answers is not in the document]
dheerajsngh: Added a TODO. Will take it as a follow up later.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+
+  private final GoogleCloudStorageOutputStream delegate;
+
+  public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
+    this.delegate = delegate;
+    logger.atInfo().log("GcsAnalyticsCoreOutputStreamWrapper created");
+  }
```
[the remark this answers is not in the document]
dheerajsngh: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,29 +192,49 @@ public GoogleHadoopOutputStream(
...
-      GoogleCloudStorageEventBus.postOnException();
-      throw (FileAlreadyExistsException)
-          new FileAlreadyExistsException(String.format("'%s' already exists", gcsPath))
-              .initCause(e);
+      GoogleHadoopFileSystem ghfs, URI gcsPath, CreateFileOptions options) throws IOException {
+    if (ghfs.isAnalyticsWriteEnabled()) {
+      logger.atInfo().log("Using Analytics Core write path for %s", gcsPath);
+      GcsFileSystem analyticsGcsFs = ghfs.getAnalyticsCoreGcsFs();
+      GcsWriteOptions baseWriteOptions =
+          analyticsGcsFs.getFileSystemOptions().getGcsClientOptions().getGcsWriteOptions();
```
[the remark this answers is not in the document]
dheerajsngh: Fixed.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+  public synchronized void close() throws IOException {
+    logger.atInfo().log("Closing GcsAnalyticsCoreOutputStreamWrapper");
+    try {
+      delegate.close();
+      logger.atInfo().log("Closed GcsAnalyticsCoreOutputStreamWrapper successfully");
+    } catch (IOException e) {
+      logger.atWarning().withCause(e).log("Failed to close GcsAnalyticsCoreOutputStreamWrapper");
+      throw e;
+    }
+  }
```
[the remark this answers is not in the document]
dheerajsngh: Teh underneath SDK channel uses a buffer and auto-flushes the data if the data is more than the buffer size([link/](https://github.com/googleapis/google-cloud-java/blob/f84034bcd4ec9bff69fe2a01c8d40106232ab290/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedWritableByteChannel.java#L121))
Hence the above is not applicable for our case here.  Forcefully sending flush could lead to extra network calls.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * A synchronized wrapper around the non-thread-safe GoogleCloudStorageOutputStream.
+ *
+ * <p>TODO(user): Implement GoogleCloudStorageItemInfo.Provider once gcs-analytics-core exposes the
+ * finalized generation ID upon close.
+ */
+class GcsAnalyticsCoreOutputStreamWrapper extends OutputStream {
```
shrutisinghania, line 32: This wrapper omits delegating `flush()`. Is there any particular reason for that?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+  private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();
+
+  private final GoogleCloudStorageOutputStream delegate;
+
+  public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
+    this.delegate = checkNotNull(delegate, "delegate cannot be null");
+  }
+
+  @Override
+  public synchronized void write(int b) throws IOException {
```
shrutisinghania, line 43: All the methods in this wrapper is `synchronized` which is redundant because Hadoop's `FSDataOutputStream` strictly guarantees writes are not thread-safe, and the rest of the connector does'nt synchronize these paths anyway.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+    logger.atFine().log("write(byte[], off=%d, len=%d)", off, len);
+    delegate.write(b, off, len);
+  }
+
+  @Override
+  public synchronized void close() throws IOException {
+    try {
+      delegate.close();
+    } catch (IOException e) {
+      logger.atWarning().withCause(e).log("Failed to close GcsAnalyticsCoreOutputStreamWrapper");
```
shrutisinghania: Logging and rethrowing in `close()` creates duplicate stack traces in the logs (the caller or Telemetry will already record this failure). We should just let the exception handle it.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,28 +192,64 @@ public GoogleHadoopOutputStream(
...
+      GoogleCloudStorageOutputStream rawStream =
+          ghfs.createAnalyticsCoreOutputStream(itemId, writeOptions);
+      OutputStream outputStream = new GcsAnalyticsCoreOutputStreamWrapper(rawStream);
+      return maybeWrapInBufferedOutputStream(ghfs, outputStream);
+    } else {
+      GoogleCloudStorageFileSystem gcsfs = ghfs.getGcsFs();
+      WritableByteChannel channel;
+      try {
+        channel = gcsfs.create(gcsPath, options);
+      } catch (java.nio.file.FileAlreadyExistsException e) {
```
shrutisinghania: A full import can be done instead

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,28 +192,64 @@ public GoogleHadoopOutputStream(
...
-    WritableByteChannel channel;
-    try {
-      channel = gcsfs.create(gcsPath, options);
-    } catch (java.nio.file.FileAlreadyExistsException e) {
-      GoogleCloudStorageEventBus.postOnException();
-      throw (FileAlreadyExistsException)
-          new FileAlreadyExistsException(String.format("'%s' already exists", gcsPath))
-              .initCause(e);
+      GoogleHadoopFileSystem ghfs, URI gcsPath, CreateFileOptions options) throws IOException {
+    if (ghfs.isAnalyticsCoreWriteEnabled()) {
```
shrutisinghania: We are missing the exception translation and Event Bus posting that exists in the legacy path. If a file already exists and overwrite is disabled, Analytics Core might throw a standard exception. We must explicitly catch, post to `GoogleCloudStorageEventBus.postOnException()`, and translate this to `FileAlreadyExistsException` so that Hadoop committer logic handles it gracefully.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -349,6 +389,8 @@ private void commitTempFile() throws IOException {
     // TODO(user): return early when 0 bytes have been written in the temp files
     tmpOut.close();
```
shrutisinghania, line 425: Right now, calling `close()` will result in calling `tmpOut.close()` twice (once inside `commitTempFile()`, and then immediately after on line 466). While channel closure is generally idempotent, in Analytics Core's Telemetry metrics double closing will cause the `WRITE_CLOSE` event and its timing to be measured and recorded twice, corrupting the metrics.

shrutisinghania (changes requested):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
dheerajsngh: Yes, it is intentional to skip `flush`. We dont want to call flush on the SDK write channel. It might force the SDK channel to flush the data. The Sdk channel maintains a buffers and internally buffers the data and flushes it when its full([link/](https://github.com/googleapis/google-cloud-java/blob/f84034bcd4ec9bff69fe2a01c8d40106232ab290/java-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedWritableByteChannel.java#L121)). Since we want to it to be no-op we did not override it.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
dheerajsngh: >> Hadoop's FSDataOutputStream strictly guarantees writes are not thread-safe

We need to synchronize the write methods precisely because its not synchronized in [FSDataOutputStream](https://github.com/apache/hadoop/blob/388394cfd450ec1fd19aae1a4461d0accadc24ef/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FSDataOutputStream.java#L52). 

>> the rest of the connector does'nt synchronize these paths

We do synchronize write for the JSON path([link/](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/81b5605debd457a6ce0aa4c62c27b307d5132aca/util/src/main/java/com/google/cloud/hadoop/util/AbstractGoogleAsyncWriteChannel.java#L110)).

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
dheerajsngh: Makes sense. Updated

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: We need inline import is required here because there is already a  `org.apache.hadoop.fs.FileAlreadyExistsException`. To avoid the conflict, inline import is present.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: Updated. 
Thanks for pointing it out.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: Updated `GcsAnalyticsCoreOutputStreamWrapper`, to  keep a flag for checking if the stream was already closed, to stop double counting of the metrics. 

Thanks for pointing it out.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,28 +194,86 @@ public GoogleHadoopOutputStream(
...
+      boolean overwrite = options.getWriteMode() == CreateFileOptions.WriteMode.OVERWRITE;
+      StorageResourceId resourceId =
+          StorageResourceId.fromUriPath(gcsPath, /* allowEmptyObjectName= */ false);
+      GcsItemId.Builder itemIdBuilder =
+          GcsItemId.builder()
+              .setBucketName(resourceId.getBucketName())
+              .setObjectName(resourceId.getObjectName());
+      long generation = StorageResourceId.UNKNOWN_GENERATION_ID;
+      try {
+        GcsFileInfo fileInfo = analyticsGcsFs.getFileInfo(itemIdBuilder.build());
```
shrutisinghania: Calling `analyticsGcsFs.getFileInfo()` before every stream creation introduces an extra RPC on every write, doubling the operation count for file creation. Additionally, this condition will already be handled.

We should remove the `getFileInfo` pre-check entirely and catch `java.nio.file.FileAlreadyExistsException` during stream creation, matching the legacy path

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,28 +194,86 @@ public GoogleHadoopOutputStream(
...
+        // If we reach here, the file exists.
+        if (!overwrite) {
+          GoogleCloudStorageEventBus.postOnException();
+          throw new FileAlreadyExistsException(String.format("'%s' already exists", gcsPath));
+        }
+        generation = ((Optional<Long>) fileInfo.getItemInfo().getContentGeneration()).orElse(0L);
+      } catch (IOException e) {
+        if (e.getMessage() != null && e.getMessage().startsWith("Object not found:")) {
```
shrutisinghania: IMO, matching Error message is not a good idea

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,28 +194,86 @@ public GoogleHadoopOutputStream(
...
         createOutputStream(
-            ghfs.getGcsFs(),
-            tmpGcsPath,
-            tmpIndex == 0 ? createFileOptions : TMP_FILE_CREATE_OPTIONS);
+            ghfs, tmpGcsPath, tmpIndex == 0 ? createFileOptions : TMP_FILE_CREATE_OPTIONS);
     this.dstGenerationId = StorageResourceId.UNKNOWN_GENERATION_ID;
     this.traceFactory = ghfs.getTraceFactory();
   }
 
   private static OutputStream createOutputStream(
```
shrutisinghania, line 201: nit: Extract the Analytics Core and Legacy stream creation logic into dedicated private helper methods. This keeps `createOutputStream` high-level and easy to read

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapperTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GcsAnalyticsCoreOutputStreamWrapperTest {
+
+  @Mock private GoogleCloudStorageOutputStream mockOutputStream;
+
+  private GcsAnalyticsCoreOutputStreamWrapper adapter;
```
shrutisinghania: nit: `adapter` should be renamed to `wrapper` or `outputStreamWrapper` for consistency.

shrutisinghania (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: The above check had to be applied due to the [implementation](https://github.com/suni72/gcs-analytics-core/blob/3542db3dd25357a23ac3e0a5201365061c6d59ff/client/src/main/java/com/google/cloud/gcs/analyticscore/client/GcsClientImpl.java#L220) in analytics-core , which sends a generic IOException rather than `FileAlreadyExistsException`. Added a TODO to fix this after updating it in analytics-core.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: Done.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapperTest.java
dheerajsngh: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: Yes, ideally that should be the ideal way. But currently server does not throw the Exception during the stream creation, due to which the above had to be added. Currently hadoop-connector implementation also does a [get](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/c247104e43ca451774cf75f3c2f1bae30d97b067/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientImpl.java#L1711) call.
Added a TODO to fix this.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+  private final GoogleCloudStorageOutputStream delegate;
+  private boolean closed = false;
+
+  public GcsAnalyticsCoreOutputStreamWrapper(GoogleCloudStorageOutputStream delegate) {
+    this.delegate = checkNotNull(delegate, "delegate cannot be null");
+  }
+
+  @Override
+  public synchronized void write(int b) throws IOException {
+    logger.atFiner().log("write(int)");
```
animesh-g: This logging is not practical. If we log on every byte, it can be very noisy and wastes CPU cycles.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -188,16 +194,78 @@ public GoogleHadoopOutputStream(
...
+    // TODO(user): Check if FileAlreadyExistsException can be thrown during stream creation in
+    // Analytics Core rather than doing an explicit getFileInfo check here.
+    try {
+      GcsFileInfo fileInfo = analyticsGcsFs.getFileInfo(itemIdBuilder.build());
+      // If we reach here, the file exists.
+      if (!overwrite) {
+        GoogleCloudStorageEventBus.postOnException();
+        throw new FileAlreadyExistsException(String.format("'%s' already exists", gcsPath));
+      }
+      generation = ((Optional<Long>) fileInfo.getItemInfo().getContentGeneration()).orElse(0L);
```
animesh-g: nit: casting seems unnecessary here.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapperTest.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2026 Google Inc.
```
animesh-g: nit: Needs to be LLC now like GcsAnalyticsCoreOutputStreamWrapper.java‎ .

animesh-g (approved):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapperTest.java
dheerajsngh: Fixed.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
dheerajsngh: Updated.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GcsAnalyticsCoreOutputStreamWrapper.java
dheerajsngh: Remoed .

PR #1770, opened 2026-08-05
Map hadoop-connector write flags to gcs-analytics core flags

dheerajsngh:
* Map Hadoop-Connector write flags to Analytics-Core flags.

### DIFF
--- a/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
+++ b/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
@@ -34,6 +34,15 @@ final class AnalyticsCoreConfigMapper {
   static final String RANDOM_READ_MIN_REQ_SIZE_KEY = "analytics-core.random-read.min-request-size";
   static final String ADAPTIVE_READ_SEQ_THRESHOLD_KEY =
       "analytics-core.adaptive-read.sequential-read-threshold";
+  static final String UPLOAD_CHUNK_SIZE_KEY = "channel.write.chunk-size-bytes";
+  static final String UPLOAD_TYPE_KEY = "channel.write.upload-type";
+  static final String TEMPORARY_PATHS_KEY = "channel.write.temporary-paths";
+  static final String PCU_BUFFER_COUNT_KEY = "channel.write.pcu.buffer.count";
+  static final String PCU_BUFFER_CAPACITY_KEY = "channel.write.pcu.buffer.capacity-bytes";
+  static final String PCU_PART_FILE_CLEANUP_TYPE_KEY = "channel.write.pcu.part-file.cleanup-type";
+  static final String PCU_PART_FILE_NAME_PREFIX_KEY = "channel.write.pcu.part-file.name-prefix";
+  static final String ENCRYPTION_KEY_KEY = "encryption-key";
+  static final String CHECKSUM_VALIDATION_ENABLED_KEY = "channel.write.checksum-validation.enabled";
 
   private static final ImmutableMap<String, String> HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS =
       ImmutableMap.<String, String>builder()
@@ -62,6 +71,26 @@ final class AnalyticsCoreConfigMapper {
           .put(
               GoogleHadoopFileSystemConfiguration.GCS_INPUT_STREAM_FADVISE.getKey(),
               FILE_ACCESS_PATTERN_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_OUTPUT_STREAM_UPLOAD_CHUNK_SIZE.getKey(),
+              UPLOAD_CHUNK_SIZE_KEY)
+          .put(GoogleHadoopFileSystemConfiguration.GCS_CLIENT_UPLOAD_TYPE.getKey(), UPLOAD_TYPE_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_BUFFER_COUNT.getKey(),
+              PCU_BUFFER_COUNT_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_BUFFER_CAPACITY.getKey(),
+              PCU_BUFFER_CAPACITY_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_PART_FILE_CLEANUP_TYPE.getKey(),
+              PCU_PART_FILE_CLEANUP_TYPE_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_PART_FILE_NAME_PREFIX.getKey(),
+              PCU_PART_FILE_NAME_PREFIX_KEY)
+          .put(GoogleHadoopFileSystemConfiguration.GCS_ENCRYPTION_KEY.getKey(), ENCRYPTION_KEY_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_FILE_CHECKSUM_TYPE.getKey(),
+              CHECKSUM_VALIDATION_ENABLED_KEY)
           .build();
 
   private AnalyticsCoreConfigMapper() {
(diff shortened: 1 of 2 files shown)


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -84,10 +113,25 @@ static Map<String, String> mapConfigs(Configuration config, String prefix) {
         (hadoopKey, analyticsKey) ->
             mapAndRemoveSource(hadoopKey, mappedProperties, prefix + analyticsKey));
 
+    // Handle temporary paths: use fs.gs.write.temporary.dirs if set, otherwise fallback to
+    // hadoop.tmp.dir
+    String tempPaths =
+        mappedProperties.remove(
+            GoogleHadoopFileSystemConfiguration.GCS_WRITE_TEMPORARY_FILES_PATH.getKey());
```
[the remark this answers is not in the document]
dheerajsngh: We cannot define custom prefix in the hadoop connector. 
Hadoop-connector allows multiple prefixes(¨fs.gs" and ¨google.cloud¨) for properties in [HadoopCredentialConfiguration](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/util-hadoop/src/main/java/com/google/cloud/hadoop/util/HadoopCredentialsConfiguration.java). 

Since fs.gs.write.temporary.dirs is not present in the above file, hence hadoop-connector will ignore it if it we specify any other prefix apart from ¨fs.gs¨.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -62,6 +71,26 @@ final class AnalyticsCoreConfigMapper {
           .put(
               GoogleHadoopFileSystemConfiguration.GCS_INPUT_STREAM_FADVISE.getKey(),
               FILE_ACCESS_PATTERN_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_OUTPUT_STREAM_UPLOAD_CHUNK_SIZE.getKey(),
+              UPLOAD_CHUNK_SIZE_KEY)
+          .put(GoogleHadoopFileSystemConfiguration.GCS_CLIENT_UPLOAD_TYPE.getKey(), UPLOAD_TYPE_KEY)
```
shrutisinghania, line 77: `GCS_CLIENT_UPLOAD_TYPE` is mapped directly to `channel.write.upload-type` without string normalization or validation. If users pass lower-case strings or legacy enum values, Analytics Core configuration parsing will fail. Recommend adding a conversion helper similar to `toFileAccessPattern()` for enum `UploadType`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -62,6 +71,26 @@ final class AnalyticsCoreConfigMapper {
...
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_BUFFER_COUNT.getKey(),
+              PCU_BUFFER_COUNT_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_BUFFER_CAPACITY.getKey(),
+              PCU_BUFFER_CAPACITY_KEY)
+          .put(
+              GoogleHadoopFileSystemConfiguration.GCS_PCU_PART_FILE_CLEANUP_TYPE.getKey(),
```
shrutisinghania, line 85: Similar to previous comment, `GCS_PCU_PART_FILE_CLEANUP_TYPE` expects enum `PartFileCleanupType` so an explicit mapper method is required.

shrutisinghania (changes requested):
These flags are currently not documented in the [gcs-analytics-core CONFIGURATION.md](https://github.com/GoogleCloudPlatform/gcs-analytics-core/blob/main/CONFIGURATION.md) 
Could we update the documentation there for these new write path flags, including their valid string/enum values and default settings?
This will help us in making sure that the default values are same across both libraries.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
dheerajsngh: Done.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
dheerajsngh: We already do [normalization](https://github.com/GoogleCloudPlatform/gcs-analytics-core/blob/c9448289657de213efbd6d3137c9b623593591dd/client/src/main/java/com/google/cloud/gcs/analyticscore/client/GcsClientOptions.java#L135) in analytics-core when parsing the values.  Updated the code to normalize the enum values for safety.

dheerajsngh:
> These flags are currently not documented in the [gcs-analytics-core CONFIGURATION.md](https://github.com/GoogleCloudPlatform/gcs-analytics-core/blob/main/CONFIGURATION.md) Could we update the documentation there for these new write path flags, including their valid string/enum values and default settings? This will help us in making sure that the default values are same across both libraries.

We would be adding the configs to [Configuration.md,](https://github.com/GoogleCloudPlatform/gcs-analytics-core/blob/main/CONFIGURATION.md) after completing the hadoop-connectors and Iceberg integrations and benchmarking for stability.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapperTest.java
```diff
@@ -189,6 +241,133 @@ private Configuration createTestConfiguration() {
...
+        GoogleHadoopFileSystemConfiguration.GCS_CLIENT_UPLOAD_TYPE.getKey(),
+        "parallel-composite-upload");
+    assertThat(
+            AnalyticsCoreConfigMapper.mapConfigs(config, "fs.gs.")
+                .get("fs.gs." + AnalyticsCoreConfigMapper.UPLOAD_TYPE_KEY))
+        .isEqualTo("PARALLEL_COMPOSITE_UPLOAD");
+  }
+
+  @Test
+  public void mapConfigs_mapsRequesterPaysProjectIdConditionally() {
```
shrutisinghania, line 328: Could we also add a test case verifying that `user-project` falls back to `fs.gs.project.id` when `fs.gs.requester.pays.mode` is `AUTO`/`ENABLED` and `fs.gs.requester.pays.project.id` is not set?

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
shrutisinghania: What I meant regarding validation is that `normalizeEnum()` only performs string transformation (`replace('-', '_').toUpperCase()`), but does not validate whether the resulting string is actually a valid `UploadType` (or `PartFileCleanupType`).
For instance, with `fadvise`, `toFileAccessPattern()` explicitly checks against known enum values and provides a safe default fallback (`AUTO_SEQUENTIAL`) if an unrecognized value is passed.

shrutisinghania (changes requested):

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapperTest.java
dheerajsngh: Fixed.  Thanks.

shrutisinghania (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
```diff
@@ -34,13 +36,20 @@ final class AnalyticsCoreConfigMapper {
...
+  static final String ENCRYPTION_KEY_KEY = "encryption-key";
+  static final String CHECKSUM_VALIDATION_ENABLED_KEY = "channel.write.checksum-validation.enabled";
 
   private static final ImmutableMap<String, String> HADOOP_TO_ANALYTICS_CORE_KEY_MAPPINGS =
       ImmutableMap.<String, String>builder()
           .put(GoogleHadoopFileSystemConfiguration.GCS_PROJECT_ID.getKey(), PROJECT_ID_KEY)
-          .put(
-              GoogleHadoopFileSystemConfiguration.GCS_REQUESTER_PAYS_PROJECT_ID.getKey(),
-              USER_PROJECT_KEY)
+          .put(GoogleHadoopFileSystemConfiguration.GCS_ROOT_URL.getKey(), SERVICE_HOST_KEY)
```
animesh-g, line 52: nit: can you add fs.gs.universe.domain as well here?

animesh-g (approved):

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/AnalyticsCoreConfigMapper.java
dheerajsngh: Actually currently in analytics-core we do not set universe domain . We set it in [hadoop-connector](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/be4784e9a07aa0a3b85da193922a94d2bd3d6b6f/gcsio/src/main/java/com/google/cloud/hadoop/gcsio/StorageClientProvider.java#L153)
Will add a follow-up item for the it in analytics-core.

PR #1777, opened 2026-08-19
Create CODEOWNERS and add multi-approvers

IshaAg:
Create CODEOWNERS and add @GoogleCloudPlatform/gcs-connector


.github/workflows/members.json
```diff
@@ -0,0 +1,15 @@
+[
+  {"login": "rajeevpodar"},
+  {"login": "ShreyasSinha"},
+  {"login": "ajayky-os"},
+  {"login": "chandra-siri"},
+  {"login": "prudhvimaharishi"},
+  {"login": "animesh-g"},
+  {"login": "Dhriti07"},
+  {"login": "suni72"},
```
dheerajsngh: We can remove sun72. She has transitioned to a different team.

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @GoogleCloudPlatform/gcs-connector @singhravidutt
```
dheerajsngh: This might not work since the repo is owned by `GoogleCloudDataproc`. Can you check this error?

.github/workflows/members.json
```diff
@@ -0,0 +1,15 @@
+[
```
shrutisinghania: If we already have the CODEOWNERS defined, why is this file needed again?

CODEOWNERS
IshaAg: replacing with ldaps, I cannot view the available teams in GoogleCloudDataproc

.github/workflows/members.json
IshaAg: removing this file

.github/workflows/members.json
IshaAg: it was required for multi-approvers, but not needed now

shrutisinghania (changes requested):
Instead of hardcoding individual usernames in the `CODEOWNERS` and workflow files, could we use a GitHub Team (e.g., @GoogleCloudDataproc/gcs-hyd-connectors)? This will make it much easier to maintain, as we won't need to open a new PR every time someone joins or leaves the team

IshaAg:
> Instead of hardcoding individual usernames in the `CODEOWNERS` and workflow files, could we use a GitHub Team (e.g., @GoogleCloudDataproc/gcs-hyd-connectors)? This will make it much easier to maintain, as we won't need to open a new PR every time someone joins or leaves the team

this can be done separately, as it requires creating a teams on Github for which i don't have permission.

dheerajsngh (approved):

IshaAg:
> > Instead of hardcoding individual usernames in the `CODEOWNERS` and workflow files, could we use a GitHub Team (e.g., @GoogleCloudDataproc/gcs-hyd-connectors)? This will make it much easier to maintain, as we won't need to open a new PR every time someone joins or leaves the team
> 
> this can be done separately, as it requires creating a teams on Github for which i don't have permission.

team is created and added in the PR

dheerajsngh (approved):

shrutisinghania (approved):

IshaAg:
Looks like the multi-approvers workflow will fail until the PR is merged
7f26igbgabtoo

PR #1784, opened 2026-09-09
feat: Skip soft-delete of temporary objects during compose

animesh-g:
Implement go/skip-soft-delete-compose-connector to bypass soft-delete retention for temporary tail objects during compose operations in Hadoop writes.

- Add configuration property `fs.gs.operation.compose.delete-source.enable` (default: false).
- Plumb `isComposeDeleteSourceEnabled` through `GoogleCloudStorageOptions` and `CreateObjectOptions.isDeleteSourceObjects`.
- In `GoogleCloudStorageImpl`, set `deleteSourceObjects(true)` on `ComposeRequest` when enabled.
- In `GoogleHadoopOutputStream`, set `deleteSourceObjects = true` on `composeObjectOptions` and skip enqueuing asynchronous tail deletion tasks.
- Keep `GoogleHadoopFileSystem.concat` untouched (`deleteSourceObjects = false`).
- Support non-destination source object deletion in `InMemoryGoogleCloudStorage`.
- Add unit tests for configuration, `GoogleCloudStorageImpl`, and `GoogleHadoopOutputStream`.


gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStream.java
```diff
@@ -166,13 +169,21 @@ public GoogleHadoopOutputStream(
         minSyncInterval.isNegative() || minSyncInterval.isZero()
             ? null
             : RateLimiter.create(/* permitsPerSecond= */ 1_000.0 / minSyncInterval.toMillis());
+    this.composeDeleteSourceEnabled =
+        ghfs.getGcsFs() != null
+            && ghfs.getGcsFs().getOptions() != null
+            && ghfs.getGcsFs().getOptions().getCloudStorageOptions() != null
+            && ghfs.getGcsFs().getOptions().getCloudStorageOptions().isComposeDeleteSourceEnabled();
```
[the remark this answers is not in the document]
animesh-g: older comment. Implemented in both JSON and gRPC

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageTest.java
```diff
@@ -3667,6 +3671,118 @@ public void testComposeObjectsWithGenerationId() throws Exception {
...
+                new MockLowLevelHttpRequest() {
+                  @Override
+                  public LowLevelHttpResponse execute() throws IOException {
+                    return jsonDataResponse(destinationObject);
+                  }
+                };
+            requests.add(request);
+            return request;
+          }
+        };
```
dheerajsngh: nit: we can move the method into [MockHttpTransportHelper](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/util/src/main/java/com/google/cloud/hadoop/util/testing/MockHttpTransportHelper.java#L50) or use one of the existing method.

gcsio/src/test/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageClientTest.java
```diff
@@ -623,6 +623,65 @@ public void compose_succeeds() throws Exception {
...
+    try (FakeServer fakeServer = FakeServer.of(mockStorage)) {
+      GoogleCloudStorage gcs =
+          mockedGcsClientImpl(transport, fakeServer.getGrpcStorageOptions().getService());
+
+      gcs.composeObjects(
+          sources,
+          destination,
+          CreateObjectOptions.builder().setDeleteSourceObjects(false).build());
+    }
+
```
dheerajsngh, line 679: nit: add a comment as to why 2 requests are required similar to GoogleCloudStorageTest

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopOutputStreamTest.java
```diff
@@ -327,15 +337,146 @@ public void hsync_statistics() throws IOException {
...
+        fout.write(data2, 0, data2.length);
+        fout.hsync();
+      }
+
+      // Verify that compose request did not request deleteSourceObjects
+      assertThat(capturedOptions.get()).isNotNull();
+      assertThat(capturedOptions.get().isDeleteSourceObjects()).isFalse();
+
+      // Verify that the temporary tail file was deleted via the background cleanup thread pool
+      assertThat(backgroundDeletedObjects).isNotEmpty();
```
dheerajsngh: nit:fout is in a try-with-resources block, there would be an additional compose+delete call with close. Can we verify that as well?

dheerajsngh (approved):
LGTM, added some nit comments for tests.

PR #1795, opened 2026-09-17
Fixed GCS API Time metric to include both TTFB and transfer duration

mudit-97:


util/src/main/java/com/google/cloud/hadoop/util/GcsReadDurationTrackerStream.java
```diff
@@ -39,7 +41,7 @@ public GcsReadDurationTrackerStream(
       URI streamPath,
       HttpHeaders responseHeaders,
       long latencyLoggingThresholdMs) {
-    super(delegate);
+    super(checkNotNull(delegate, "delegate cannot be null"));
     this.streamPath = streamPath;
     this.responseHeaders = responseHeaders;
     this.latencyLoggingThresholdMs = latencyLoggingThresholdMs;
```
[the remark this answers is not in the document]
mudit-97: we always use captureAndResetThreadLocalStats which resets the metrics and shield from double counting. Also on the multi - close statement, that code path was not touched in this PR, it was already handled in master. Both sites close exactly once — openStream's tracker via the channelIsOpen-guarded closeContentChannel(), and cacheFooter's via its try-with-resources.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/VectoredIOImpl.java
```diff
@@ -350,6 +350,11 @@ private void readCombinedRange(
         // complete exception all the underlying ranges which have not already
         // finished.
         completeExceptionally(combinedFileRange, e);
+      } finally {
+        // try-with-resources closes the read channel before this block runs. Metrics that are
```
shameersss1, line 354: In readSingleRange(), range.getData().complete(dst) is currently called inside the try block while channel is still open. Because try-with-resources closes the channel after completing the future, the
data transfer metric is emitted and captured in finally after the caller thread has already unblocked from range.getData().get().

This creates a race condition where the caller may inspect the metrics before the worker thread reaches finally

shameersss1 (changes requested):

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
animesh-g: I think we should implement these changes in gRPC channels as well otherwise we will end up having inconsistencies between HTTP and gRPC metrics.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemIntegrationTest.java
```diff
@@ -2901,6 +2901,19 @@ private void verifyMetrics(
...
+          // GCS_API_TIME includes both HTTP request TTFB (summed above via expectedSum) and
+          // stream data transfer duration posted on stream close. Because
+          // STREAM_READ_DATA_TRANSFER_DURATION is recorded via updateStats() rather than
+          // updateMetric(), its base opsCount is 0, so we use STREAM_READ_CLOSE_OPERATIONS as
+          // the sample count.
+          long streamReadCloseCount =
+              stats.getLong(GhfsStatistic.STREAM_READ_CLOSE_OPERATIONS.getSymbol());
+          expectedValue +=
+              stats.getLong(GhfsStatistic.STREAM_READ_DATA_TRANSFER_DURATION.getSymbol() + "_mean")
+                  * streamReadCloseCount;
```
animesh-g: Optional:  Multiplying `streamReadCloseCount` by the mean `STREAM_READ_DATA_TRANSFER_DURATION` is brittle. In cases involving footer prefetching or larger read ranges, multiple data transfers may occur, meaning `streamReadCloseCount` will not accurately reflect the actual `dataTransferCount`.

gcsio/src/main/java/com/google/cloud/hadoop/gcsio/GoogleCloudStorageReadChannel.java
animesh-g: I think we should implement these changes in gRPC channels as well otherwise we will end up having inconsistencies between HTTP and gRPC metrics.

animesh-g (changes requested):

PR #1800, opened 2026-09-21
fix: correct credential translation on the Analytics Core path

prudhvimaharishi:
Reject DOWNSCOPED access tokens rather than silently falling back to the provider's full-scope token, and also substitute NoCredentials for the null that UNAUTHENTICATED resolves to.


shrutisinghania (changes requested):
We are using the feature branch `analytics-core-migration` for this migration. Can you raise this PR there instead of `master`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -468,7 +468,43 @@ private GcsFileSystem createAnalyticsGcsFs(Configuration config) throws IOExcept
...
+   * equivalent here. Downscoping is rejected rather than ignored: falling back to the provider's
+   * broad token would widen the access boundary the operator configured, on reads that look
+   * identical to every other read.
+   *
+   * @param credentials The credentials resolved for this filesystem.
+   * @throws IOException If the configured {@link AccessTokenProvider} issues downscoped tokens,
+   *     which Analytics Core cannot honor.
+   */
+  private static void rejectDownscopedTokens(GoogleCredentials credentials) throws IOException {
+    if (!(credentials instanceof AccessTokenProviderCredentials)) {
```
shrutisinghania: Is it possible for workloads to supply native DownscopedCredentials (from the Google Auth Library) directly, bypassing the Hadoop-specific AccessTokenProvider wrapper? If so, this `instanceof` check will evaluate to false and allow them through, which would silently broaden their scope and bypass this fix. 

Should we also check if `credentials instanceof DownscopedCredentials` here?

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
```diff
@@ -641,4 +642,89 @@ public void close_canBeCalledMultipleTimes() throws Exception {
...
+      return AccessTokenType.DOWNSCOPED;
+    }
+
+    @Override
+    public AccessToken getAccessToken() {
+      return new AccessToken("downscoped-access-token", Instant.now().plusSeconds(600));
+    }
+
+    @Override
+    public void refresh() {}
```
shrutisinghania: nit: To resolve the static analysis failure, consider throwing an `UnsupportedOperationException` or adding a simple `// no-op` comment inside the method body.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -468,7 +468,43 @@ private GcsFileSystem createAnalyticsGcsFs(Configuration config) throws IOExcept
...
+      return;
+    }
+    AccessTokenProvider accessTokenProvider =
+        ((AccessTokenProviderCredentials) credentials).getAccessTokenProvider();
+    if (accessTokenProvider.getAccessTokenType() != AccessTokenType.DOWNSCOPED) {
+      return;
+    }
+    throw new IOException(
```
shrutisinghania: `initializeGcsFs(createGcsFs(config))` runs before `createAnalyticsGcsFs(config)` so if `rejectDownscopedTokens(credentials)` throws an `IOException`, `fs.initialize(...)` aborts after `gcsFs` has already been created (and after `getCredentials(config)` has already been invoked once), leaving `gcsFs` unclosed.
Consider either validating/creating `analyticsCoreGcsFs` before `createGcsFs(config)` or closing `gcsFs` if `createAnalyticsGcsFs(config)` fails.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
```diff
@@ -468,7 +468,43 @@ private GcsFileSystem createAnalyticsGcsFs(Configuration config) throws IOExcept
...
+        String.format(
+            "%s does not support %s access tokens, which would be silently broadened to the"
+                + " provider's full scope. Either set %s to false, or configure an %s that issues"
+                + " %s tokens.",
+            GCS_ANALYTICS_CORE_ENABLE.getKey(),
```
shrutisinghania: Misleading error message when only `fs.gs.analytics.core.write.enable` is set.
We should reference both `GCS_ANALYTICS_CORE_ENABLE.getKey()` and `GCS_ANALYTICS_CORE_WRITE_ENABLE.getKey()` or just change the error message to say that currently analytics core does not support it.

shrutisinghania (changes requested):
The PR description mentions substituting `NoCredentials` for `null` when `UNAUTHENTICATED` is configured. Are we doing that?

prudhvimaharishi:
> The PR description mentions substituting `NoCredentials` for `null` when `UNAUTHENTICATED` is configured. Are we doing that?

No its stale. Updated

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
prudhvimaharishi: Fixed in https://github.com/GoogleCloudDataproc/hadoop-connectors/commit/db52e99545ab93793dc4c9a7b1ffa9cef1ced43c — updated the error message to say `Analytics Core (fs.gs.analytics.core.enable / fs.gs.analytics.core.write.enable) does not support DOWNSCOPED access tokens...` and added a unit test for `fs.gs.analytics.core.write.enable=true`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
prudhvimaharishi: Fixed in https://github.com/GoogleCloudDataproc/hadoop-connectors/commit/db52e99545ab93793dc4c9a7b1ffa9cef1ced43c — moved `createAnalyticsGcsFs(config)` before `createGcsFs(config)` in `initializeGcsFs` so validation/creation fails fast before `gcsFs` is allocated.

gcs/src/test/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemTest.java
prudhvimaharishi: Fixed in https://github.com/GoogleCloudDataproc/hadoop-connectors/commit/db52e99545ab93793dc4c9a7b1ffa9cef1ced43c — added `// no-op` inside `refresh()`.

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
prudhvimaharishi: Good question! Two reasons we only check `AccessTokenProviderCredentials` with `AccessTokenType.DOWNSCOPED`:\n1. `GoogleHadoopFileSystem.getCredentials(config)` resolves credentials exclusively from Hadoop `Configuration` via `HadoopCredentialsConfiguration`, which cannot construct `com.google.auth.oauth2.DownscopedCredentials` (and `GoogleHadoopFileSystem` has no constructor accepting a caller-built `Credentials` instance).\n2. Even if a static `DownscopedCredentials` were passed to Analytics Core, its `getRequestMetadata()` already returns the downscoped token directly—it would not be broadened. Privilege broadening only happens with Hadoop's `AccessTokenProviderCredentials` when `getAccessTokenType() == DOWNSCOPED`, because its 0-arg `getAccessToken()` returns the provider's full-scope token unless the caller explicitly invokes `getAccessToken(List<AccessBoundary>)` per request (as `createGcsFs` does).

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
prudhvimaharishi: Fixed

gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystem.java
prudhvimaharishi: This is not possible.

-  builds credentials exclusively from Hadoop Configuration via HadoopCredentialsConfiguration, which never produces com.google.auth.oauth2.DownscopedCredentials (and GoogleHadoopFileSystem has no constructor or setter accepting a pre-built Credentials instance).

- Even if a static DownscopedCredentials were passed to Analytics Core, its getRequestMetadata() already returns the downscoped token—it is not broadened. Privilege broadening only happens with Hadoop's AccessTokenProviderCredentials + AccessTokenType.DOWNSCOPED, because its 0-arg getAccessToken() returns the provider's full-scope token while createGcsFs calls the per-request getAccessToken(List<AccessBoundary>) overload.

